PR #6, opened 2019-10-18
Migrate rollup-plugin-replace

shellscape:
- Rollup Plugin Name: 

This PR contains:
- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?
- [ ] no

Breaking Changes?
- [x] no

### Description

This PR migrates `rollup-plugin-replace` to this monorepo.


packages/replace/rollup.config.js
```diff
@@ -0,0 +1,12 @@
+import buble from 'rollup-plugin-buble';
```
wesleygrimes: Just a thought here, doesn't affect this PR, but with it being a mono repo, one of the huge positives is that plugins are all tested with each other for compatibility and breaking changes. Do we want to consider actually importing the plugin from the monorepo once they are migrated over?

PR #11, opened 2019-10-28
chore: migrate rollup-plugin-wasm

shellscape:
- Rollup Plugin Name: 

This PR contains:
- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?
- [ ] no

Breaking Changes?
- [ ] no

This bumps the minimum version of Rollup to ^1.20.0. That's going to break users of `rollup-plugin-wasm` on older versions from the original repo.

### Description

Migrates rollup-plugin-wasm


packages/wasm/rollup.config.js
```diff
@@ -0,0 +1,15 @@
+const pkg = require('./package.json');
+
+const dependencies = Object.keys(pkg.dependencies || {});
+
+export default [
+  {
+    input: 'src/index.js',
+    output: { exports: 'named', file: 'dist/index.js', format: 'cjs' },
+    external: ['path'].concat(dependencies)
```
jamen: This is missing `'fs'` and produces a warning:

```
src/index.js → dist/index.js...
(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
fs (imported by src\index.js)
created dist/index.js in 35ms
```

Maybe could do all builtin modules?

```js
const pkg = require('./package.json')
const { builtinModules } = require('module')
const dependencies = Object.keys(pkg.dependencies || {})

export default {
    external: [ ...builtinModules, ...dependencies ]
}
```

packages/wasm/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "emscripten"
+  ],
+  "peerDependencies": {
+    "rollup": "^1.20.0"
+  },
+  "devDependencies": {
+    "del-cli": "^3.0.0",
+    "rollup": "^1.20.0",
+    "source-map": "^0.7.3"
+  },
```
jamen: I think `ava` is missing from dev dependencies, unless its supposed to be installed globally. Running `npm test`:
```
bash: ava: command not found
```

jamen: Ignore this: I looked at the rest of the repo, and saw dependency management for the monorepo is a bit more involved.

All good!

packages/wasm/rollup.config.js
shellscape: @jamen please feel free to commit to the branch. as a member of Plugin Maintainers, you have permissions.

jamen: Thank you, done.

packages/wasm/rollup.config.js
```diff
@@ -0,0 +1,16 @@
...
+
+const dependencies = Object.keys(pkg.dependencies || {});
+
+export default [
+  {
+    input: 'src/index.js',
+    output: { exports: 'named', file: 'dist/index.js', format: 'cjs' },
+    external: [ ...builtinModules, ...dependencies ]
+  }
+  // {
```
wesleygrimes: Are we intending to leave dead code here?

shellscape: no we can kill it

wesleygrimes:
> I'd like to hold off on typings for plugins that don't exist as it's added time onto the migration. We've agreed that we should convert all plugins to TS, and I'd like that to be a second-pass over the plugins once we're going strong with the monorepo.

OK 👍

jamen:
I've never wrote typings, but I can give it a shot on second-pass, unless others are doing it.

shellscape:
@jamen do you have the ability to deprecate `rollup-plugin-wasm`? I don't have publish permission. The command should be:

```console
npm deprecate rollup-plugin-wasm "This module has moved and is now available at @rollup/plugin-wasm. Please update your dependencies. This version is no longer maintained."
```

PR #30, opened 2019-11-07
chore: migrate rollup-plugin-json

shellscape:
- Rollup Plugin Name: 

This PR contains:
- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?
- [ ] no

Breaking Changes?
- [x] no

### Description

This PR migrates the rollup-plugin-json plugin.


packages/json/package.json
```diff
@@ -0,0 +1,64 @@
...
+  },
+  "ava": {
+    "files": [
+      "!**/fixtures/**",
+      "!**/helpers/**",
+      "!**/recipes/**",
+      "!**/types.ts"
+    ]
+  },
+  "jsnext:main": "dist/rollup-plugin-json.es.js",
```
TrySound: Why deprecated jsnext:main is still here after major release?

shellscape: it's faux-major. technically a new module, so there was no "breaking change release" when first published. The major bump was there as a signifier that this wouldn't work as a straight replacement for the old module name due to dependency changes. AFAIK rollup isn't displaying any deprecation messaging about that property, so it wasn't caught. Feel free to submit a PR to correct that. As it is, it doesn't hurt anything.

PR #36, opened 2019-11-15
chore: migrate rollup-plugin-yaml

shellscape:
- Rollup Plugin Name: 

This PR contains:
- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?
- [ ] no

Breaking Changes?
- [x] no

### Description

This PR migrates the rollup-plugin-yaml plugin. It also fixes up some noticed metadata.


packages/yaml/CHANGELOG.md
```diff
@@ -0,0 +1,15 @@
+# @rollup/plugin-yaml Change Log
+
+## 2.0.0
+*2019-10-18*
+
+* Add `transform` option [#6](https://github.com/rollup/rollup/pull/6) (by @CharlesHolbrow)
```
dumbmatter: URLs in this file are wrong

shellscape: good catch. this is a holdover from the rollup-plugin-alias repo, so we'll fix this here but the original repo will still have the errors.

packages/yaml/package.json
```diff
@@ -0,0 +1,69 @@
...
+  "keywords": [
+    "rollup",
+    "plugin",
+    "yaml"
+  ],
+  "peerDependencies": {
+    "rollup": "^1.20.0"
+  },
+  "dependencies": {
+    "@rollup/plugin-json": "^4.0.0",
```
pnevares: Am I being dense or is `@rollup/plugin-json` not used by this package?

packages/yaml/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "ci:lint:commits": "commitlint --from=${CIRCLE_BRANCH} --to=${CIRCLE_SHA1}",
+    "ci:test": "pnpm run test -- --verbose",
+    "lint": "pnpm run lint:js && pnpm run lint:docs && pnpm run lint:package",
+    "lint:docs": "prettier --single-quote --write README.md",
+    "lint:js": "eslint --fix --cache src test",
+    "lint:package": "prettier --write package.json --plugin=prettier-plugin-package",
+    "prebuild": "del-cli dist",
+    "prepare": "npm run build",
+    "prepublishOnly": "npm run lint",
+    "pretest": "npm run build",
```
pnevares: Looks like other plugins already in the repo also use `npm` for these scripts instead of `pnpm`, is that intentional? They seemed to work fine locally when I tested them with `pnpm`.

shellscape: good catch. this will need fixing in other plugins likely as well

packages/yaml/package.json
shellscape: this is interesting. I'll look into it

PR #46, opened 2019-11-20
chore: migrate rollup-plugin-url

pnevares:
## Rollup Plugin Name: `@rollup/plugin-url`

This PR contains:
- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?
- [ ] no

Breaking Changes?
- [ ] no

### Description

This PR migrates the rollup-plugin-url plugin.


packages/url/package.json
```diff
@@ -0,0 +1,73 @@
...
+  },
+  "description": "Import files as data-URIs or to ES6 modules.",
+  "license": "MIT",
+  "repository": "rollup/plugins",
+  "author": "Arpad Borsos <arpad.borsos@googlemail.com>",
+  "homepage": "https://github.com/rollup/plugins/packages/url/#readme",
+  "bugs": "https://github.com/rollup/plugins/issues",
+  "main": "dist/index.js",
+  "engines": {
+    "node": ">=10.0.0"
```
shellscape: Rollup is still supporting `>= 8.0.0`

packages/url/package.json
```diff
@@ -0,0 +1,73 @@
+{
+  "name": "@rollup/plugin-url",
+  "version": "3.0.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "description": "Import files as data-URIs or to ES6 modules.",
```
shellscape:
```suggestion
  "description": "Import files as data-URIs or as ES6 modules",
```

packages/url/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "README.md",
+    "LICENSE"
+  ],
+  "keywords": [
+    "rollup",
+    "plugin",
+    "url"
+  ],
+  "peerDependencies": {
+    "rollup": ">=0.60.0"
```
shellscape:
```suggestion
    "rollup": ">=1.20.0"
```

packages/url/package.json
```diff
@@ -0,0 +1,73 @@
+{
+  "name": "@rollup/plugin-url",
+  "version": "3.0.0",
```
shellscape: We'll have to bump this to `4.0.0` and note in the PR that this contains a breaking change in the form of a new minimum version of Rollup.

packages/url/package.json
```diff
@@ -0,0 +1,73 @@
...
+  },
+  "dependencies": {
+    "mime": "^2.4.4",
+    "rollup-pluginutils": "^2.8.2"
+  },
+  "devDependencies": {
+    "@babel/core": "^7.7.2",
+    "@babel/preset-env": "^7.7.1",
+    "@babel/register": "^7.7.0",
+    "mocha": "^6.2.2",
```
shellscape: swap for `ava`

packages/url/package.json
```diff
@@ -0,0 +1,73 @@
...
+    ],
+    "babel": {
+      "extensions": [
+        "png",
+        "svg"
+      ]
+    }
+  },
+  "jsnext:main": "src/index.js",
+  "maintainers": [
```
shellscape: these sections won't be pertinent moving forward as we'll have the `Plugin Maintainers` team on the Rollup org

packages/url/rollup.config.js
```diff
@@ -0,0 +1,28 @@
...
+export default {
+  input: 'src/index.js',
+  plugins: [
+    babel({
+      presets: [
+        [
+          '@babel/preset-env',
+          {
+            targets: {
+              node: 10
```
shellscape:
```suggestion
              node: 8
```

packages/url/test/test.js
```diff
@@ -0,0 +1,162 @@
...
+  );
+}
+
+const DEFAULT_OPTIONS = {
+  publicPath: '',
+  emitFiles: true
+};
+
+function run(input, options, experimentalCodeSplitting = false) {
+  const writeOptions = {
```
shellscape: Just a heads up: We've now got a utilities that'll test a bundle as well as get the code for a bundle, so this function shouldn't be needed anymore. You can leverage `t.snapshot` in `ava` and you won't have to write the file to disk, which will make the `assertExists` function no longer needed as well.

packages/url/package.json
Ashot-KR: @pnevares I will take care of this. can you give me permission to push to repo?

pnevares: @Ashot-KR I've invited you as a collaborator on my fork 👍

packages/url/package.json
pnevares: Thanks! Which section(s) other than `maintainers` are you referring to?

pnevares:
> If you need a hand with the `ava` conversion let me know.

Thanks, I hope haven't missed an obvious mistake: I spent a while trying to figure out why `ava` wasn't working at all for even the most basic tests I was trying to convert, and then I just blew away the file and replaced it with:

```js
// packages/url/test/test.js
const test = require('ava');

test('foo', (t) => {
  t.pass();
});
```

And I still get:
```bash
$ pnpm test
Scope: current workspace package
> @rollup/plugin-url@3.0.0 test /Users/pablonevares/Projects/rollup-plugins/packages/url
> ava


  ✖ Couldn't find any files to test

Scope: current workspace package
 ERROR  Test failed. See above for more details.```

File: packages/url/package.json (hunk 1 above)
Comment (Ashot-KR) [reply packages/url/package.json]: Fixed. Plugin will work with node >=8

**Comment by Ashot-KR** [comment]
@pnevares It's because of
```json
"babel": {
  "extensions": [
    "png",
    "svg"
  ]
}
```
in package.json in `ava` section.

AVA docs:
> extensions of test files that will be precompiled using AVA's Babel presets. Setting this overrides the default "js" value, so make sure to include that extension in the list

You override default "js" value, so ava trying to find tests in `.png` and `.svg` files. I think we don't need add `extentions` here at all – images can't be compiled by babel.

packages/url/package.json
shellscape: Probably `authors` and `contributors` too, but I haven't seen those in a plugin yet.

shellscape:
@Ashot-KR is correct. In package.json, you can re-add the `.js` extension in the `ava` config section. Interesting conflict on this one :)

pnevares:
Thanks for the help with `ava` folks, that was definitely the issue. 🙊

I updated several of the tests to use `ava` and `testBundle()`, but I haven't been able to understand why I get the following error when `emitFiles` is true, which was the old default option for these tests:

```  Rejected promise returned by test. Reason:

  TypeError (NodeError) {
    code: 'PLUGIN_ERROR',
    hook: 'generateBundle',
    plugin: 'at position 1',
    pluginCode: 'ERR_INVALID_ARG_TYPE',
    message: 'The "path" argument must be of type string. Received type undefined',
  }

  Object.write (src/index.js:73:65)
  node_modules/rollup/dist/rollup.js:12849:25```

**Comment by shellscape** [comment]
@pnevares if you can give me write access to your fork, I can give this a nudge for you.

**Comment by pnevares** [comment]
Invite sent, appreciated @shellscape 🙌

**Comment by shellscape** [comment]
@pnevares I've pushed my changes to the test. Basically reverted back to what they were in the plugin's repo and updated the `run` method, while simplifying it quite a bit. Using `testBundle` adds more complexity to the tests than it was worth. The last two tests with multiple inputs really weren't necessary, and the tests have 100% coverage, which is great.

File: packages/url/src/index.js [hunk 9]
```diff
@@ -0,0 +1,127 @@
...
+          const outputDirectory = path.join(base, path.dirname(output));
+          await makeDir(outputDirectory);
+          return copy(name, path.join(base, output));
+        })
+      );
+    }
+  };
+}
+
+function promise(fn, ...args) {
```
Andarist: that's rather surprising name - I would go with `promisify`, in fact there is `require('util').promisify` which already does this, but I'm not sure if it's available in all currently supported node versios, so that would have to be checked

packages/url/src/index.js
```diff
@@ -0,0 +1,127 @@
...
+    },
+    generateBundle: async function write(outputOptions) {
+      // Allow skipping saving files for server side builds.
+      if (!emitFiles) return;
+
+      const base = options.destDir || outputOptions.dir || path.dirname(outputOptions.file);
+
+      await makeDir(base);
+
+      Promise.all(
```
Andarist: if you are calling Promise.all here I assume you should await this 😉

pnevares: Nice catch, updated!

packages/url/src/index.js
pnevares: I had thought the same early on in this work but forgot to go back and check. Thanks for the reminder, util.promisify is from node 8 so it should be safe to use in this project, I've added it 👍

pnevares:
Thanks for the hard work @shellscape, everything is complete as far as I can tell.

Ashot-KR:
I think all ok, but we should check [this PR](https://github.com/rollup/rollup-plugin-url/pull/27) before merge.

packages/url/src/index.js
shellscape: @Andarist it's important to remember when reviewing migration PRs, that we're just trying to get the plugin moved in it's present form. We'll be opening issues to analyze each plugin after the entire migration is done.

packages/url/package.json
```diff
@@ -61,6 +61,7 @@
   "ava": {
     "files": [
       "!**/fixtures/**",
+      "!**/output/**",
```
shellscape: why was this necessary? I've not seen an issue with `output` in any of my tests

README.md
```diff
@@ -30,7 +30,7 @@ This repository houses plugins that Rollup considers critical to every day use o
 | [virtual](packages/virtual)           | Load virtual modules from memory                                                          |
 | [wasm](packages/wasm)                 | Import WebAssembly code with Rollup                                                       |
 | [yaml](packages/yaml)                 | Convert YAML files to ES6 modules                                                         |
-|                                       |                                                                                           |
+| [url](packages/url)                   | Import files as data-URIs or ES Modules                                                   |
```
shellscape: this should be alphabetized. I'll handle it.

packages/url/package.json
pnevares: Each test run left the final `/output` behind from the last unit that ran, and ava detected it the next time I ran all tests:

>   ✖ No tests found in test/output/bundle.js, make sure to import "ava" at the top of your test file

Could also solve by changing beforeEach to afterEach, but I guessed you had good reasons for using beforeEach.

shellscape: Weird. I never had that issue last night.

Ashot-KR:
@shellscape great!
can we publish `@rollup/plugin-url` on npm now?
I'll take care on deprecation `rollup-plugin-url` package

shellscape:
@Ashot-KR thanks, just did.

Please use this for the deprecation message, which follows our wording on other plugins:

```
"This module has been deprecated and is no longer maintained. Please use @rollup/plugin-url."
```

Ashot-KR:
`rollup-plugin-url` deprecated
@pnevares thanks for your work!

pnevares:
Thanks for being a welcoming group and helping when I got stuck! And to @shellscape for the tweet that brought me in 🎊

shellscape:
Glad to have you here! I hope you stick around and continue to be involved!

PR #55, opened 2019-11-25
Allow Replacement plugin to operate as an output plugin

darthtrevino:

This PR contains:
- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?
- [ ] no

Breaking Changes?
- [x] no

### Description
This PR allows for the replacement plugin to be used as an output plugin. As an example, one may wish to output a minimized production build and development build using the `rollup.output` configuration and use the replace plugin in both of these contexts.

For example, I would like to use this configuration with [react-dnd](github.com/react-dnd/react-dnd):

```js
import path from 'path'
import babel from 'rollup-plugin-babel'
import resolve from 'rollup-plugin-node-resolve'
import commonjs from 'rollup-plugin-commonjs'
import { terser } from 'rollup-plugin-terser'
import replace from '@rollup/plugin-replace'

export default {
	input: path.resolve(__dirname, 'lib/index.js'),
	external: ['react', 'react-dom'],
	globals: {
		react: 'React',
		'react-dom': 'ReactDOM',
	},
	output: [
		{
			name: 'ReactDnD',
			file: path.resolve(__dirname, 'dist/umd/ReactDnD.js'),
			format: 'umd',
			plugins: [
				replace({
					values: {'process.env.NODE_ENV': JSON.stringify('development')},
					delimiters: ['', ''],
				}),
			],
		},
		{
			name: 'ReactDnD',
			file: path.resolve(__dirname, 'dist/umd/ReactDnD.min.js'),
			format: 'umd',
			plugins: [terser(), replace({
				values: {'process.env.NODE_ENV': JSON.stringify('production')},
				delimiters: ['', ''],
			}),],
		},
	],
	plugins: [
		resolve(),
		commonjs(),
		babel({ /* babel-config */ }),
	],
}
```

Note:
The sourceMapsEnabled predicate looked weird to me, so I wrote some tests specifically for sourcemap emission with different configuration settings. It turned out that the predicate was fine, but I'm adding the tests that I wrote to this PR.


packages/replace/src/index.js
```diff
@@ -47,39 +47,49 @@ export default function replace(options = {}) {
...
-      if (!filter(id)) return null;
+    let hasReplacements = false;
+    let match;
+    let start;
+    let end;
+    let replacement;
 
-      const magicString = new MagicString(code);
+    // eslint-disable-next-line no-cond-assign
+    while ((match = pattern.exec(code))) {
```
wuweiweiwu: seems like it'll be cleaner if we pull this assignment to L54

darthtrevino: I can change this up if you'd prefer that to be restructured. My goal was to be as minimally intrusive as possible, so this is just the contents of the original `transform` function.

darthtrevino: changes made!

packages/replace/test/sourcemaps.js
```diff
@@ -0,0 +1,127 @@
+/* eslint-disable consistent-return, global-require, import/no-dynamic-require */
```
pnevares: It doesn't look like `global-require` and `import/no-dynamic-require` are disabling any warnings or errors in this file.

darthtrevino: You're right - I fixed this in misc.js as well

shellscape:
thanks!

PR #57, opened 2019-11-27
chore: add @rollup/plugin-html

shellscape:

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

This PR proposes adding a _**new**_ plugin to the repo. It's based on https://github.com/styleguidist/mini-html-webpack-plugin, with permission, and emits HTML files that can be used for serving a bundle. Coverage is at 100%.


packages/html/lib/index.js
```diff
@@ -0,0 +1,87 @@
+const { extname } = require('path');
+
+const getFiles = (bundle) => {
```
lukastaegert: This function should ONLY get entry point files (isEntry: true) and ignore the rest. Reason:
- For IIFE and UMD formats, there will only be a single entry point anyway
- All other formats import their dependencies themselves. If you add all files as script tags, you will at least destroy dynamic lazy loading of chunks (if it works at all).

shellscape: Disagree here. Various assets can be added to a bundle's files via plugins, and we should pick those up. CSS is the major example here.

packages/html/lib/index.js
```diff
@@ -0,0 +1,87 @@
...
+    name: 'html',
+
+    async generateBundle(output, bundle) {
+      const files = getFiles(bundle);
+      const source = await template({ attributes, files, publicPath, title });
+
+      const htmlFile = {
+        type: 'asset',
+        source,
+        name: 'Rollup HTML Asset',
```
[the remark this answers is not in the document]
shellscape: Noted. We need to update the Rollup docs as this is a copy/paste/modify from them. I'll see about doing that this week.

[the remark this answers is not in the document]
lukastaegert: It is written in the first paragraph after the code block here: https://rollupjs.org/guide/en/#thisemitfileemittedfile-emittedchunk--emittedasset--string

packages/html/lib/index.js
```diff
@@ -0,0 +1,87 @@
+const { extname } = require('path');
```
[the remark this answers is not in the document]
shellscape: Just faster on the draw with CJS. We're going to convert the entire repo to TypeScript here at some point. So there's no functional difference at the moment.

shellscape:
@lukastaegert I will look into and implement the various format warnings that we discussed via chat, prior to merge of this.

packages/html/lib/index.js
lukastaegert: This function puts all files .js files into an array and throws away all meta information. It is virtually impossible to distiguish  the auto-generated chunks from the entry points. This makes the plugin rather useless for code-splitting IMO. Either you provide some meta information, or you ignore the non-entry chunks. I would be very confused if a CSS plugin was broken if you ignored the auto-generated chunks, as they are still imported by the main JavaScript. And maybe this even breaks CSS plugins in so far as maybe their chunks are meant to be loaded dynamically together with their content while you make them load up-front instead.

packages/html/lib/index.js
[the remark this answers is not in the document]
shellscape: I did read that, but with the code sample above that having both and both being optional, I had assumed it would be prudent to set `name` in the event that something was inspecting it down the line. Some extra verbiage there could be useful to point out there's no need to set the other property.

packages/html/lib/index.js
shellscape: That makes the case for providing the bundle metadata to the `template` option function for complex builds that the default template doesn't suit.

packages/html/README.md
```diff
@@ -0,0 +1,136 @@
...
+### `publicPath`
+
+Type: `String`<br>
+Default: `''`
+
+Specifies a path to prepend to all bundle assets (files) in the HTML output.
+
+### `template`
+
+Type: `Fumction`<br>
```
bebraw: Fumction -> Function

packages/html/README.md
```diff
@@ -0,0 +1,136 @@
...
+```js
+const { makeHtmlAttributes } = require('@rollup/plugin-html');
+
+makeHtmlAttributes({ lang: 'en', 'data-batcave': 'secret' });
+// -> 'lang="en" data-batcave="secret"'
+```
+
+## Attribution
+
+This plugin was inspired by and is based upon [mini-html-webpack-plugin](https://github.com/styleguidist/mini-html-webpack-plugin) by Juho Vepsalainen, with permission.
```
bebraw: Vepsalainen -> Vepsäläinen. I would also mention Artem Sapegin as he is a co-author of the plugin.

packages/html/lib/index.js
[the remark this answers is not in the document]
lukastaegert: True. It does not hurt adding the name, it is just ignored.

packages/html/lib/index.js
lukastaegert: That would be sufficient

lukastaegert: With regard to my initial comment, maybe I mis-stated what I mean: This function should of course return all assets, but with regard to chunks, it should only return entry chunks, as all non-entry chunks (read: .js files) will be loaded automatically by the entry chunks, no additional work necessary. But additionally having the bundle available to the template still would not hurt.

packages/html/README.md
shellscape:
```suggestion
Type: `Function `<br>
```

packages/html/README.md
shellscape:
```suggestion
This plugin was inspired by and is based upon [mini-html-webpack-plugin](https://github.com/styleguidist/mini-html-webpack-plugin) by Juho Vepsäläinen and Artem Sapegin, with permission.
```

shellscape:
@lukastaegert great, thanks. we're going to do an improvement round after migration of the existing plugins and I'll slate some time to working on templates we can export for the unsupported formats.

PR #58, opened 2019-11-28
fix: Argument order in testBundle

NotWoods:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

Fixes the order of the arguments passed into `func` inside `testBundle`. 

`t` is the last argument in `params`, so it should also be the last value passed into `func`.

Also added some JSDoc comments to add intellisense.

### DIFF
--- a/util/test.js
+++ b/util/test.js
@@ -1,3 +1,7 @@
+/**
+ * Gets the code from a rollup bundle.
+ * @param {import('rollup').RollupBuild} bundle
+ */
 const getCode = async (bundle) => {
   const { output } = await bundle.generate({ format: 'cjs' });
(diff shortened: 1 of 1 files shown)


shellscape:
Got this fixed in https://github.com/rollup/plugins/pull/32/files#diff-8ac5641d44f25ade01cd49aee477d812

NotWoods:
I'll close and wait for that PR to merge

shellscape:
:+1: thanks for spotting this none the less

PR #73, opened 2019-12-03
docs(typescript): include jsx `preserve` instructions

zanona:

This PR contains:

- [x] documentation

Are tests included?

- [x] no

Breaking Changes?

- [x] no

List any relevant issue numbers: #72 

### Description
Adds instructions for preserving the `jsx: 'preserve'` option which will fail by default if the pluging `acorn-jsx` is not also used.


shellscape:
Looks solid. Thanks :beer:

PR #77, opened 2019-12-04
chore: remove placeholder security script from each plugin

### LINKED ISSUE #75: repo task: enable pnpm audit
- Rollup Plugin Name: all plugins
- Rollup Plugin Version: current

n/a

### Actual Behavior / Situation

All plugins currently have a placeholder for the `"security"` script. 

### Modification Proposal

We need to enable use of `pnpm audit`. I'm not sure yet how this will work. pnpm may work on the main lockfile only, which would negate the need for the script in each plugin. It may operate on individual package.json files, but that's unlikely.

pnevares:
## Rollup Plugin Name: `all`

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

Resolves #75 

### Description

- Removes the placeholder `security` script from each plugin's `package.json`
- I verified that `pnpm audit` only uses `pnpm-lock.yaml` to run the audit.
  - I installed `lodash@4.17.11` (contains a vulnerability) in the `run` package and `pnpm audit` reported the vulnerability when ran from the repo's root.
  - Only adding the vulnerable dependency to `packages/run/package.json` had no effect on `pnpm audit`
  - Only adding the vulnerable dependency to `pnpm-lock.yaml` caused the vulnerability to show up during `pnpm audit`
- _Unrelated change: the typescript package had two un-linted files, I've included them here_

### DIFF
--- a/packages/alias/package.json
+++ b/packages/alias/package.json
@@ -14,7 +14,7 @@
   "scripts": {
     "build": "rollup -c",
     "ci:coverage": "nyc pnpm run test && nyc report --reporter=text-lcov > coverage.lcov",
-    "ci:lint": "pnpm run build && pnpm run lint && pnpm run security",
+    "ci:lint": "pnpm run build && pnpm run lint",
     "ci:lint:commits": "commitlint --from=${CIRCLE_BRANCH} --to=${CIRCLE_SHA1}",
     "ci:test": "pnpm run test -- --verbose",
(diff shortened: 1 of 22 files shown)


pnevares:
It looks like we have 15 packages running `security` as part of their `lint` script and the other 5 as part of their `ci:lint` script. I can remove all but should I update `.circleci/config.yml` or let someone else handle?

PR #81, opened 2019-12-06
fix(json): cannot be imported by rollup 

foray1010:
## Rollup Plugin Name: `json`

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [ ] no

### Description

it's current jsnext:main field is pointing to a wrong path, which makes rollup failed to import the plugin

### DIFF
--- a/packages/json/package.json
+++ b/packages/json/package.json
@@ -58,6 +58,5 @@
       "!**/types.ts"
     ]
   },
-  "jsnext:main": "dist/rollup-plugin-json.es.js",
   "module": "dist/index.es.js"
 }


foray1010:
@shellscape the `jsnext:main` file of package.json in `@rollup/plugin-json` is pointing to a file that is not exist. This PR is meant to point it to correct path, which should be the same as `module`

I am not sure if I really need to test the changes, since it is not code related, just some misconfig

shellscape:
@foray1010 that would be an _other_ PR then, as that's metadata. Please make sure you fill out the entire template next time. The information in your reply should have been in the original post for the PR.

foray1010:
difficult to say as user cannot use this package if the path is wrong, which is definitely a bug to me

shellscape:
After @TrySound mentioned that `jsnext:main` is deprecated, I did some digging and asking around. Rollup was really the only bundler that embraced it fully. As mentioned by Rich Harris [here](https://twitter.com/Rich_Harris/status/1203299929285971968), `jsnext:main` has been superseded by `module`, and there are numerous sources that consider it deprecated as a result ([for example](https://github.com/stereobooster/package.json#jsnextmain)).

Basically it never made it out of discussion, and Rollup supported it out of necessity at the time it was implemented.

So all that goes to say is that we have sufficient evidence to remove support from `@rollup/node-resolve` and to remove that field from any plugins in the repo. As such, I'd like to ask that your PR remove the `jsnext:main` property instead of updating the value. We'll then release it as a major.

foray1010:
@shellscape I have created https://github.com/rollup/plugins/pull/85 which remove all `jsnext:main` field in this repo
also updated this PR

PR #88, opened 2019-12-10
chore: migrate rollup-plugin-multi-entry

shellscape:

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] no

### Description

Migrates `rollup-plugin-multi-entry`.

#### Breaking Changes

- Minimum Rollup version ^1.20.0
- Changes entry tag

### DIFF
--- a/packages/multi-entry/README.md
+++ b/packages/multi-entry/README.md
@@ -1,0 +1,125 @@
+[npm]: https://img.shields.io/npm/v/@rollup/plugin-multi-entry
+[npm-url]: https://www.npmjs.com/package/@rollup/plugin-multi-entry
+[size]: https://packagephobia.now.sh/badge?p=@rollup/plugin-multi-entry
+[size-url]: https://packagephobia.now.sh/result?p=@rollup/plugin-multi-entry
+
+[![npm][npm]][npm-url]
+[![size][size]][size-url]
+[![libera manifesto](https://img.shields.io/badge/libera-manifesto-lightgrey.svg)](https://liberamanifesto.com)
+
+# @rollup/plugin-multi-entry
+
+🍣 A Rollup plugin which allows use of multiple entry points for a bundle.
+
+As an added bonus, the _named exports_ from all entry points will be combined. This is particularly useful for tests, but can also be used to package a library.
+
+_Note: `default` exports cannot be combined and exported by this plugin. Only named exports will be exported._
+
+## Requirements
+
+This plugin requires an [LTS](https://github.com/nodejs/Release) Node version (v8.0.0+) and Rollup v1.20.0+.
+
+## Install
+
+Using npm:
+
+```console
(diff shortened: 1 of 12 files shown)


packages/multi-entry/test/test.js
```diff
@@ -0,0 +1,74 @@
+/* eslint-disable no-bitwise */
+
+import test from 'ava';
```
NotWoods: Most of our tests have been using CommonJS, should we use it here too?

packages/multi-entry/test/test.js
```diff
@@ -0,0 +1,74 @@
...
+import { rollup } from 'rollup';
+
+import { getCode } from '../../../util/test';
+
+import multiEntry from '../';
+
+test('takes a single file as input', async (t) => {
+  const bundle = await rollup({ input: 'test/fixtures/0.js', plugins: [multiEntry()] });
+  const code = await getCode(bundle);
+  t.truthy(~code.indexOf('exports.zero = zero;'));
```
NotWoods: These bitwise operators are hard to read, I think `includes` would be better.
```suggestion
  t.true(code.includes('exports.zero = zero;'));
```

pnevares:
Would adding this module to the monorepo be a good time to introduce a fresh `CHANGELOG.md`? I noticed one or two other projects don't have one, so I'm not sure what the standard is.

packages/multi-entry/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+```js
+import multi from '@rollup/plugin-multi-entry';
+
+export default {
+  input: ['batman.js', 'robin.js', 'joker.js'],
+  output: {
+    dir: 'output'
+  },
+  plugins: [multi()]
```
Andarist: in here the plugin is referred to as `multi` but in an example a little bit later it's referred to as `multiEntry` - I think it would be best to unify those (preferably to `multiEntry`)

shellscape: agreed. unifying is best. typically we'll choose the terser name. ala node-resolve and auto-install.

packages/multi-entry/test/test.js
shellscape: ava comes with baked in support for es6 via babel. eventually I'd like to see esm in tests and we have a stated goal of getting all source on typescript and esm. this test is just ahead of the curve.

packages/multi-entry/test/test.js
shellscape: you kids and your fancy new es6 methods

old man

shellscape:
@pnevares yes good observation. I'll be adding one for this module when I publish. I've been working on automated changefile updates and versioning but it's not quite ready. publishing still has too many manual tasks.

Issue #86, opened 2019-12-09
dedupe leads to errors in a monorepo with hoisted dependencies


jpaquim:
- Rollup Plugin Name: @rollup/plugin-node-resolve
- Rollup Plugin Version: 6.0.0
- Rollup Version: 1.27.9
- Operating System (or Browser): macOS
- Node Version: 13.2.0

### How Do We Reproduce?

Here is a repo reproducing the issue: https://github.com/jpaquim/rollup-monorepo-issue

Steps:
- Start with a monorepo configuration using package hoisting (the default for yarn workspaces, or using lerna with hoisting enabled)
- Setup rollup and @rollup/plugin-node-resolve in one of the monorepo's packages (let us assume it's in a directory under packages/).
- Configure @rollup/plugin-node-resolve to dedupe a certain package (in my case, I was using sveltejs/template, with `dedupe: importee => importee === 'svelte' || importee.startsWith('svelte/')` or more simply `dedupe: ['svelte', 'svelte/internal']`.
- Import the given package in one of the source files
- Bundle using rollup, using as working directory the package directory, and not the monorepo root.

### Expected Behavior
The imported package should be correctly bundled with the rest of the output.

### Actual Behavior
The imported package is not resolved correctly, rollup complains: `(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
svelte/internal (imported by src/App.svelte)`

Consequently, the import statement remains in the output. In my case: `import { SvelteComponent, init, safe_not_equal, element, text, space, attr, insert, append, set_data, noop, detach } from 'svelte/internal';`

### Comments
The issue seems to be related to [this line](https://github.com/rollup/plugins/blob/master/packages/node-resolve/src/index.js#L301), where the importee is overwritten with `join(process.cwd(), 'node_modules', importee)`.

In a monorepo context, `process.cwd()` will typically refer to the package where the rollup command is started from, but with hoisting, the relevant node_modules are actually at the root of the monorepo, thus leading to the conflict.

shellscape:
Thanks for the issue :beer: 

This is going to be a difficult one to remedy in a monorepo. I'm afraid the setup is going to be a bit complex in order to get multiple folks working on triage. Can you provide a repo that has a reproducible setup? One with the setup you're using would be most helpful.

jpaquim:
@shellscape sure thing, here is a repo reproducing the issue: https://github.com/jpaquim/rollup-monorepo-issue

[...]

PR #98, opened 2019-12-20
feat(node-resolve): add rootDir option

LarsDenBakker:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: fixes https://github.com/rollup/plugins/issues/86

### Description

Introduces a `rootDir` option, which is used to resolve modules which need to imported from root. Right now these are entrypoint modules and deduped modules. The default value is `process.cwd()`, which is also what it used right now making this a non-breaking change.


shellscape:
Thanks for creating this. It looks like the ESLint TS plugin is on the fritz again. I'll try to fix that in the next day or so.

packages/node-resolve/README.md
```diff
@@ -141,6 +141,17 @@ customResolveOptions: {
...
 
+### `rootDir`
+
+Type: `String`<br>
+Default: `process.cwd()`
+
+Root directory to resolve modules from. Used when resolving entrypoint imports, and when resolving deduplicated modules. Useful when executing rollup in a package of a monorepository.
+
+```
+rootDir: path.join(process.cwd(), '..', '..')
```
NotWoods: I think it would be helpful to have a comment indicating this refers to a parent directory.
```suggestion
// Set the root directory to be the parent folder
rootDir: path.join(process.cwd(), '..')
```

packages/node-resolve/test/root-dir.js
```diff
@@ -0,0 +1,25 @@
...
+const test = require('ava');
+const { rollup } = require('rollup');
+
+const { testBundle } = require('../../../util/test');
+
+const nodeResolve = require('..');
+
+process.chdir(join(__dirname, 'fixtures', 'monorepo-dedupe', 'packages', 'package-a'));
+
+test.only('deduplicates modules from the given root directory', async (t) => {
```
NotWoods: Using `only` will make it harder to add other tests to this file later. Does either `test` or `test.serial` work?

LarsDenBakker: Indeed, it should be just `test`. Updated it!

shellscape:
Anyone have a clue as to why Github Actions wouldn't be running for this PR and @LarsDenBakker other PRs?

shellscape:
thanks!

Issue #82, opened 2019-12-08
deep imports not deduped


LarsDenBakker:
- Rollup Plugin Name: `@rollup/plugin-node-resolve`
- Rollup Plugin Version: 6.0.0
- Rollup Version: 1.27.13
- Operating System (or Browser): macOS Sierra
- Node Version: 10.16.0

### How Do We Reproduce?

- Create a package called `foo` with two files:

`shallow.js`:
```js
console.log('shallow.js');
```
`deep/deep.js`:
```js
console.log('deep/deep.js');
```

- Create a package called `bar`, with a dependency on `foo` 1.0.0
- Create a package called `buz`, with a dependency on `foo` 2.0.0
- Create a main export for both `bar` and `buz` with this content:

```js
import * from 'foo/root.js';
import * from 'foo/deep/deep.js';
```

- Create a project which depends on `bar` and `buz`
- Create an `app.js` file:

```js
import 'bar';
import 'buz';
```

- Create a rollup config:

```js
import resolve from '@rollup/plugin-node-resolve';

export default {
  input: 'app.js',
  output: {
    dir: 'output',
  },
  plugins: [resolve({
    dedupe: ['foo']
  })]
};
```

### Expected Behavior
Both `shallow.js` and `foo/deep/deep.js` are deduplicated.

### Actual Behavior
Only `shallow.js` is deduplicated. This is because the `dedupe` option of `rollup-plugin-node-resolve` uses a literal match to find out whether an import should be deduplicated (imported from root): https://github.com/rollup/plugins/blob/master/packages/node-resolve/src/index.js#L168

I don't think this intentional behavior, people would expect deduplication to happen on a package level. 

A fix can be to look at the package name, which is guaranteed to be either the first part of an import path, or the first two is the first part is a scope starting with `@`. I can work on a PR if that's the intended solution.

shellscape:
@LarsDenBakker please fill out the appropriate issue template. This is not optional and something we ask of everyone.

LarsDenBakker:
Although I think the original writeup explained the issue more easily, I've updated the issue with the template now.

PR #99, opened 2019-12-20
feat(node-resolve): dedupe by package name

LarsDenBakker:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: fixes https://github.com/rollup/plugins/issues/82

### Description

This makes the `dedupe` option also dedupe by package name. 

For example if you have your `dedupe` option set as: `['my-package']` and in your code you have imports like this:

```js
import MyPackage from 'my-package';
import foo from 'my-package/foo.js';
```

Then you would expect both to be deduplicated. Right now imports are only matched literally, so you would need to list all possible imports in the dedupe array. This gets unmaintainable quickly, especially if you take into account imports with and without file extensions.

With change `node-resolve` first checks if the import is matched literally, and then checks if the import's package name is matched. [npm package name rules](https://docs.npmjs.com/files/package.json) ensure we can safely extract the first part of the import as name, or first two if the first part is a package scope.

This isn't a breaking change since the existing deduplication behavior is maintained, people can still list both package names and full import paths. Unless people are depending on the current behavior that deep imports are not deduplicated, which I think is unlikely.


packages/node-resolve/src/index.js
```diff
@@ -136,6 +136,14 @@ function resolveImportSpecifiers(importSpecifierList, resolveOptions) {
   return p;
 }
 
+function getPackageName(id) {
+  const split = id.split('/');
+  if (split[0][0] === '@') {
+    return `${split[0]}/${split[1]}`;
```
wuweiweiwu: should this just be `id`?

LarsDenBakker: No, this will turn an import of `@my-scope/my-package/foo.js`  into `@my-scope/my-package`.

shellscape: @LarsDenBakker that's excellent info. could you add a code comment to that effect there?

LarsDenBakker: Yeah makes sense, I've added comments to the function.

LarsDenBakker:
@shellscape Thanks, I've updated the README let me know if you think it needs more changes.

shellscape:
@LarsDenBakker I'm going to give this a few more days to let others weigh in if they'd like to. Really appreciate your patience on this one. It's a significant change to be sure.

PR #107, opened 2019-12-23
docs: <br> line breaks

NotWoods:
## Rollup Plugin Name: <all plugins with options>

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

Consistently use <br> in markdown for line breaks in `## Options`.

### DIFF
--- a/packages/buble/README.md
+++ b/packages/buble/README.md
@@ -46,21 +46,21 @@ Then call `rollup` either via the [CLI](https://www.rollupjs.org/guide/en/#comma
 
 ### `transforms`
 
-Type: `Object`
+Type: `Object`<br>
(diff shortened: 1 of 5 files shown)


shellscape:
I'd actually like to persist `<br>` here. The backslash appears to be only Github Flavored Markdown specific as an alternative to two spaces, and most folks are going to be used to seeing that character used to escape something. I ran one of the changes through a few markdown previewers and all of them that did not support GFM failed to render correctly. Semantically, `<br>` leaves little room to question what it's doing and why it's there.

PR #108, opened 2019-12-23
chore: migrate rollup-plugin-babel

Andarist:

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] no

### Description

This is still WIP and there is some work left to do regarding migration, but it's ready for a preliminary review. Other than that - I plan to include minor changes to the plugin before it gets released under `@rollup` scope, so **please do not publish this without my green light on that**.

I've managed to keep git history from the old repo, which I think is nice here - but it should be merged with squash here.


packages/babel/LICENSE.md
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2015 [these people](https://github.com/rollup/rollup-plugin-babel/graphs/contributors)
```
Andarist: what should be put here now? 🤔

packages/babel/package.json
```diff
@@ -0,0 +1,68 @@
...
+  "description": "Seamless integration between Rollup and Babel.",
+  "license": "MIT",
+  "repository": "rollup/rollup-plugin-babel",
+  "author": "Rich Harris",
+  "homepage": "https://github.com/rollup/rollup-plugin-babel",
+  "bugs": {
+    "url": "https://github.com/rollup/rollup-plugin-babel/issues"
+  },
+  "main": "./dist/rollup-plugin-babel.cjs.js",
+  "engines": {
```
Andarist: I'd like to avoid source transpilation (apart of producing CJS from ESM source) which shouldn't be needed for node 10 which will become the oldest LTS in a few days. WDYT about this?

packages/babel/LICENSE.md
shellscape: Since pnpm uses the license from the root of the repo when publishing a package, we can safely remove this file. But you raise a good question. I would add a section to the `## Meta` portion of the `babel` plugin README (perhaps `### Prior Contributors`?) that stated that those contributors were responsible for the previous version in the old repo.

packages/babel/package.json
shellscape: The source is ESM, so rollup should produce both CJS and ESM. bear in mind that the plugins need to follow Rollup's lead on which engines are supported. As that's a breaking change, I doubt @lukastaegert will want to rush to that, even if a version has fallen out of LTS. It's probably going to take some time, and that build step doesn't really cost us anything.

packages/babel/test/as-input-plugin.js
```diff
@@ -0,0 +1,435 @@
...
+            visitor: {
+              Program(path, state) {
+                state.file.addHelper('__nonexistentHelper');
+              }
+            }
+          };
+        }
+      ]
+    });
+    t.fail();
```
NotWoods: `t.throwsAsync`

packages/babel/test/as-input-plugin.js
```diff
@@ -0,0 +1,435 @@
...
+
+              counter += 1;
+            }
+          }
+        };
+      }
+    ]
+  });
+
+  t.pass();
```
NotWoods: `t.notThrowsAsync`

packages/babel/test/as-input-plugin.js
```diff
@@ -0,0 +1,435 @@
...
+          code: `${result.code}\n// Generated by some custom loader`
+        };
+      }
+    };
+  });
+  const bundle = await rollup({
+    input: 'fixtures/basic/main.js',
+    plugins: [customBabelPlugin({ babelHelpers: 'bundled' })]
+  });
+  const {
```
NotWoods: `getCode`

packages/babel/test/as-input-plugin.js
```diff
@@ -0,0 +1,435 @@
...
```
NotWoods: `getCode`

packages/babel/test/as-input-plugin.js
```diff
@@ -0,0 +1,435 @@
...
+test('can be used as an input plugin while transforming the output', async (t) => {
+  const bundle = await rollup({
+    input: 'fixtures/basic/main.js',
+    plugins: [
+      babelPlugin.generated({
+        presets: ['@babel/env']
+      })
+    ]
+  });
+  const {
```
NotWoods: `getCode`

packages/babel/test/as-output-plugin.js
```diff
@@ -0,0 +1,345 @@
...
+        const { opts } = state;
+        if (path.node.property.name === 'log') {
+          path.get('property').replaceWith(t.identifier(opts.replace));
+        }
+      }
+    }
+  };
+}
+
+async function generate(input, babelOptions = {}, generateOptions = {}, rollupOptions = {}) {
```
NotWoods: In cases where only `code` is used from `generated`, `getCode` could be used instead.

packages/babel/LICENSE.md
Andarist: Hm, thanks to migrated git history and https://github.blog/changelog/2019-12-19-improved-attribution-when-squashing-commits/ we might not need to do that. But let's see what happens once we merge this in.

packages/babel/package.json
Andarist: What are currently supported engines then? Hard to find that info quickly in Rollup's repo.

Andarist: What are currently supported engines then? Hard to find that info quickly in Rollup's repo.

NotWoods: Node version (v8.0.0+) and Rollup v1.20.0+. You can check the README of individual plugins and its listed under "Requirements". It might be worthwhile to add it to the root README somewhere.

shellscape: @Andarist agreed. when Node 8 exits LTS my plan is to introduce `engines` and a requirements section to rollup core. that will require a new major, so it's something that I/we need to get lukas on board with. For the time being, all plugins (and Rollup, even though it isn't documented) are Node 8+.

packages/babel/test/as-input-plugin.js
```diff
@@ -0,0 +1,435 @@
...
+async function generate(input, babelOptions = {}, generateOptions = {}, rollupOptions = {}) {
+  const bundle = await rollup({
+    input,
+    plugins: [babelPlugin({ babelHelpers: 'bundled', ...babelOptions })],
+    ...rollupOptions
+  });
+
+  const {
+    output: [generated]
+  } = await bundle.generate(Object.assign({ format: 'cjs' }, generateOptions));
```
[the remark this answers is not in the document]
Andarist: That might fail in the future when we start to use TS here. Would prefer to keep things explicit either way.

packages/babel/test/as-input-plugin.js
```diff
@@ -0,0 +1,435 @@
...
+  t.true(code.includes('class Jsx '), 'should not transpile .jsx');
+  t.true(code.includes('class Mjs '), 'should not transpile .mjs');
+  t.false(code.includes('class Other '), 'should transpile .other');
+});
+
+test('throws when trying to add babel helper unavailable in used @babel/core version', async (t) => {
+  try {
+    await generate('fixtures/basic/main.js', {
+      plugins: [
+        function testPlugin() {
```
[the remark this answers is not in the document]
Andarist: This is a babel plugin, not the rollup one - written correctly as far as I can tell.

packages/babel/src/index.js
```diff
@@ -0,0 +1,221 @@
...
+        } = unpackOutputPluginOptions(pluginOptionsWithOverrides, outputOptions);
+        /* eslint-enable no-unused-vars */
+
+        return transformCode(code, babelOptions, overrides, customOptions, this);
+      }
+    };
+  };
+}
+
+const babelPluginFactory = createBabelInputPluginFactory();
```
Andarist: I would gladly restructure this - I believe those should be 4 separate exports, rather than a single export with static properties. This PR will land v5 of this plugin so this is IMHO OK to do.

WDYT about those:
- default export (result of `createBabelInputPluginFactory()`)
- `createBabelInputPluginFactory`
- `babelOutputPluginFactory`
- `createBabelOutputPluginFactory`

Those named exports would be quite mouthful 🤷‍♂ but correct. I'm open for other names - but cant figure out anything better myself.

Andarist:
@NotWoods I've resolved your comments, could you take another look at this?

packages/babel/package.json
Andarist: Ok, I've added transpilation using itself.

packages/babel/src/preflightCheck.js
```diff
@@ -0,0 +1,76 @@
...
+import { addBabelPlugin } from './utils';
+
+const MODULE_ERROR =
+  'Rollup requires that your Babel configuration keeps ES6 module syntax intact. ' +
+  'Unfortunately it looks like your configuration specifies a module transformer ' +
+  'to replace ES6 modules with another module format. To continue you have to disable it.' +
+  '\n\n' +
+  "Most commonly it's a CommonJS transform added by @babel/preset-env - " +
+  'in such case you should disable it by adding `modules: false` option to that preset ' +
+  '(described in more detail here - https://github.com/rollup/rollup-plugin-babel#modules ).';
```
NotWoods: This should link to the README in the monorepo instead.

packages/babel/src/preflightCheck.js
```diff
@@ -0,0 +1,76 @@
...
+
+const UNEXPECTED_ERROR =
+  'An unexpected situation arose. Please raise an issue at ' +
+  'https://github.com/rollup/rollup-plugin-babel/issues. Thanks!';
```
NotWoods: This link too

packages/babel/test/as-input-plugin.js
```diff
@@ -0,0 +1,430 @@
...
+
+  t.deepEqual(original, {
+    source: 'fixtures/class/main.js'.split(nodePath.sep).join('/'),
+    line: 3,
+    column: 12,
+    name: target
+  });
+});
+
+test('works with proposal-decorators (#18)', async (t) => {
```
NotWoods: I wonder if we should adjust the issue numbers. Maybe `rollup/rollup-plugin-babel#18`? We can also just leave it if you like.

packages/babel/test/as-input-plugin.js
```diff
@@ -0,0 +1,430 @@
...
+};
+
+export default Foo;
+`
+  );
+});
+
+test('allows transform-runtime to be used instead of bundled helpers, but throws when CommonJS is used', async (t) => {
+  await t.throwsAsync(
+    () => generate('fixtures/runtime-helpers-commonjs/main.js', { babelHelpers: 'runtime' }),
```
NotWoods: I think `throwsAsync` takes a Promise instead of a function. `throws` takes a function.

Andarist: This test wouldnt pass if expected parameters would be different, but you are right - it can receive a promise, but can also receive an async function (which this kinda is):
https://github.com/avajs/ava/blob/master/docs/03-assertions.md#throwsasyncthrower-expected-message

I can change this if you'd like, but personally I kinda like passing functions to every `throws*`-related assertion - less thinking about what "expected" parameter is.

NotWoods: I wasn't aware that you could also pass an async function. In that case you can leave it as it currently is :)

Andarist:
@NotWoods resolved your comments, please take a look especially at the documentation changes.

I still have some work to do, but in the meantime, I'd like to get "partial" approvals on what is already included here.

My TODO list:
- docs about `babelHelpers` option (compile info from Helpers section into it as well) + link to that from the source (there is currently a verbose console.warn there about it)
- establish export names - https://github.com/rollup/plugins/pull/108#discussion_r361335584
- document every change in CHANGELOG

packages/babel/README.md
```diff
@@ -0,0 +1,368 @@
...
+[size]: https://packagephobia.now.sh/badge?p=@rollup/plugin-babel
+[size-url]: https://packagephobia.now.sh/result?p=@rollup/plugin-babel
+
+[![npm][npm]][npm-url]
+[![size][size]][size-url]
+[![libera manifesto](https://img.shields.io/badge/libera-manifesto-lightgrey.svg)](https://liberamanifesto.com)
+
+# @rollup/plugin-babel
+
+🍣 Seamless integration between Rollup and Babel.
```
NotWoods:
```suggestion
🍣 A Rollup plugin for seamless integration between Rollup and Babel.
```

packages/babel/README.md
```diff
@@ -0,0 +1,368 @@
...
+
+## Why?
```
NotWoods: @shellscape Should this section be moved down?

packages/babel/README.md
```diff
@@ -0,0 +1,368 @@
...
+
+Type: `String | RegExp | Array[...String|RegExp]`<br>
+
+A [minimatch pattern](https://github.com/isaacs/minimatch), or array of patterns, which specifies the files in the build the plugin should _ignore_. When relaying on Babel configuration files you can only exclude additional files with this option, you cannot override what you have configured for Babel itself.
+
+### `include`
+
+Type: `String | RegExp | Array[...String|RegExp]`<br>
+
+A [minimatch pattern](https://github.com/isaacs/minimatch), or array of patterns, which specifies the files in the build the plugin should _ignore_. When relaying on Babel configuration files you cannot include files already excluded there.
```
NotWoods:
```suggestion
A [minimatch pattern](https://github.com/isaacs/minimatch), or array of patterns, which specifies the files in the build the plugin should operate on. When relying on Babel configuration files you cannot include files already excluded there.
```

packages/babel/README.md
```diff
@@ -0,0 +1,368 @@
...
+
+You can also run Babel on the generated chunks instead of the input files. Even though this is slower, it is the only way to transpile Rollup's auto-generated wrapper code to lower compatibility targets than ES5 (or most likely ES6 when Rollup 2 will be released), see [Running Babel on the generated code](#running-babel-on-the-generated-code) for details.
+
+All options are as per the [Babel documentation](https://babeljs.io/docs/en/options), plus the following:
+
+### `exclude`
+
+Type: `String | RegExp | Array[...String|RegExp]`<br>
+
+A [minimatch pattern](https://github.com/isaacs/minimatch), or array of patterns, which specifies the files in the build the plugin should _ignore_. When relaying on Babel configuration files you can only exclude additional files with this option, you cannot override what you have configured for Babel itself.
```
NotWoods: relying*

packages/babel/README.md
```diff
@@ -0,0 +1,368 @@
...
+
+const config = {
+  input: 'src/index.js',
+  output: [
+    {
+      file: pkg.module,
+      format: 'esm'
+    }
+  ],
+  plugins: [babel({ babelHelpers: 'bundled' })]
```
NotWoods:
```suggestion
  plugins: [babel()]
```
Since this is the default its fine to omit and be consistent with other usage examples.

packages/babel/README.md
```diff
@@ -0,0 +1,368 @@
...
+
+```js
+import babel from '@rollup/plugin-babel';
+import pkg from './package.json';
+
+const config = {
+  input: 'src/index.js',
+  output: [
+    {
+      file: pkg.module,
```
NotWoods: Let's make this example simpler and just hardcode `dir: 'output'` like the other READMEs.

packages/babel/package.json
```diff
@@ -0,0 +1,69 @@
...
+  "license": "MIT",
+  "repository": "rollup/plugins",
+  "author": "Rich Harris",
+  "homepage": "https://github.com/rollup/plugins/tree/master/packages/babel",
+  "bugs": "https://github.com/rollup/plugins/issues",
+  "main": "./dist/rollup-plugin-babel.cjs.js",
+  "engines": {
+    "node": ">= 8.0.0"
+  },
+  "scripts": {
```
NotWoods: `scripts` should be about the same as the other plugins

packages/babel/package.json
```diff
@@ -0,0 +1,69 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
+  "description": "Seamless integration between Rollup and Babel.",
+  "license": "MIT",
+  "repository": "rollup/plugins",
+  "author": "Rich Harris",
+  "homepage": "https://github.com/rollup/plugins/tree/master/packages/babel",
+  "bugs": "https://github.com/rollup/plugins/issues",
+  "main": "./dist/rollup-plugin-babel.cjs.js",
```
NotWoods:
```suggestion
  "main": "./dist/index.js",
```

packages/babel/package.json
```diff
@@ -0,0 +1,69 @@
...
+      "!**/helpers/**",
+      "!**/recipes/**",
+      "!**/types.ts"
+    ]
+  },
+  "contributors": [
+    "Bogdan Chadkin <trysound@yandex.ru>",
+    "Mateusz Burzyński <mateuszburzynski@gmail.com> (https://github.com/Andarist)"
+  ],
+  "module": "./dist/rollup-plugin-babel.esm.js"
```
NotWoods:
```suggestion
  "module": "./dist/index.es.js"
```

packages/babel/package.json
Andarist: do we have to do this? IMHO this makes stack traces less readable and using file search in dev tools a little bit less helpful, would really prefer keeping a specific file name here and would like to advocate to change this in all plugins

packages/babel/README.md
Andarist: This is default but relying on it results currently causes a console warn. The idea is that everyone should decide explicitly which one they want to use to improve the discoverability of this option.

packages/babel/README.md
Andarist: I was thinking where this should be put when I was tweaking the README and I've decided to keep it here, at the top, as it made the most sense for me. Can, of course, change this but this acts kinda as an introduction to the whole thing, so it made sense to keep it at the beginning

packages/babel/README.md
Andarist: I'm not sure if people expect to use `dir` here - isn't it targeting multi-entry scenarios (so a more complex one)? I often find people wanting to have control over the single output filename.

packages/babel/package.json
Andarist: what does it mean? what's missing here?

packages/babel/README.md
shellscape: @Andarist the real point of this section is an easy, frictionless intro into using the plugin. more complex, nuanced examples are typically located further down in either options sections or in "meta" sections after options that explain various concepts in more detail. this is a core tenant we've been following.

packages/babel/README.md
shellscape: I would say in 99% of cases, this is "meta" and should go after the really important bits (reqs, install, use, options). but the babel plugin does require some "do you really need this" knowledge up front. I'm good leaving this here as an exception.

packages/babel/package.json
shellscape: this is an easy one - take a look at literally any of the other plugins. copy and paste, adjust for build where needed.

packages/babel/package.json
```diff
@@ -0,0 +1,69 @@
...
+  },
+  "scripts": {
+    "build": "rollup -c",
+    "prebuild": "rimraf dist",
+    "prepare": "pnpm run build",
+    "pretest": "pnpm run build",
+    "preversion": "pnpm run test",
+    "test": "ava"
+  },
+  "files": [
```
shellscape: please match `files`

packages/babel/package.json
```diff
@@ -0,0 +1,69 @@
...
+  },
+  "devDependencies": {
+    "@babel/core": "^7.7.4",
+    "@babel/plugin-external-helpers": "^7.7.4",
+    "@babel/plugin-proposal-decorators": "^7.7.4",
+    "@babel/plugin-syntax-dynamic-import": "^7.7.4",
+    "@babel/plugin-transform-runtime": "^7.7.4",
+    "@babel/preset-env": "^7.7.4",
+    "@rollup/plugin-json": "^4.0.0",
+    "ava": "^2.4.0",
```
shellscape: `ava` and `rimraf` should be removed. we don't use `rimraf` in this repo, we have `del-cli` as a root dep, and ava is hoisted as well so no need to have it in devDeps. comparing against any of the other plugins' package.json is a very easy, fast way to get this right. any diff tool will work.

packages/babel/package.json
shellscape: yes, please.

IMHO it makes a stack trace _less_ readable. In a stack trace there is no need for redundancy.

```
.../node_modules/@rollup/plugin-babel/dist/index.js
```

is far cleaner than

```
.../node_modules/@rollup/plugin-babel/dist/rollup-plugin-babel.cjs.js
```

Captain Redundancy Captain does not work here :smile:

shellscape:
@Andarist there seems to be a lot of pushback against conventions that we've set in the repo for migrating this plugin. the conventions we've chosen were to suit the various styles being brought in from repos that were very fragmented. so far, they've worked quite well to bring everything in line and have made maintenance much easier.

If you'd like, I'm happy to go through and update the files that need updating to established conventions so we don't have to engage in the frequent back and forth. I'd like to get you on board, but perhaps it's better to have these discussions after the babel plugin is finished so we can complete the migration? the babel plugin is the last and only outlier, so the likelihood that we'll change the 23 other other plugins we've already completed migration for, before we complete the babel migration is pretty slim.

packages/babel/package.json
Andarist: I can ofc tweak this to whatever you like - but personally I would still vote for a meaningful file name there. Take a look at an example screenshot from my current debugging session:
Screenshot 2020-01-08 at 22 41 19
All of those `index.js` are totally unhelpful for me until I click through every one of them.

shellscape: I'll agree, in that context that's a compelling argument. I'm also good with changing the convention across the repo as long as it's consistent. With that said, the one thing I absolutely abhor is incredibly long variable names and incredibly long filenames. 

I propose `dist/@rollup.babel.js` and  `dist/@rollup.babel.es.js`.

The `@` is a valid file character in Windows according to this: https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions. It's a shame we can't use path separator characters in a Windows filename, but it is what it is. That should make it glaringly obvious what the scope of the file is, and what the context of the file is.

Andarist: @shellscape do we want to change this on master? Should I prepare a PR for this? Or should I just adjust it here to the current convention?

LarsDenBakker:
What's the status on this? `rollup-plugin-babel` 5.x is still on an alpha release. This is a vital plugin in the rollup ecosystem.

Andarist:
@LarsDenBakker I'm going to pick up working on this next week. Would appreciate ideas on https://github.com/rollup/plugins/pull/108#discussion_r361335584

packages/babel/src/index.js
lukastaegert: IMO the `createXY` names are ok as they will not be used often and only by people who know what they are doing. However I think functions should start with a verb, and the return value of `createBabelOutputPluginFactory` should be called `getOutputPlugin`, which also sounds less intimidating. For symmetry, I would furthermore export the default export additionally as  `getInputPlugin`. This will make @TrySound happy 😃

lukastaegert: On second thought `getBabelOutputPlugin` and `getBabelInputPlugin` so that it does not conflict with other plugins

lukastaegert: Also note that using named and default export together will switch the plugin to named-exports-mode when imported from CJS. This means that `const babel = require(‘rollup-plugin-babel’)` will no longer work. And I am not sure if the new CLI `rollup -c —plugin babel` would work. The latter could be fixed by providing a named export `babel` instead of a default. In any case, this change would be breaking.

Andarist: I've changed exports to pretty much to what you have suggested, you can check out this commit: https://github.com/rollup/plugins/pull/108/commits/4c5e1a8bf885d701af5d87edc4b595458870dcf9

Andarist:
@rollup/plugin-maintainers after quite a long time, way longer than I've anticipated 😢, I've got back to it, resolved all comments, updated docs and changelog. This is no longer a draft from my perspective - I would appreciate a review 🙏 so we can move this forward and finally publish it as `@rollup/plugin-babel`

shellscape:
Thank you, and no worries on the time frame! I'll get to looking at this today!

shellscape:
@Andarist I think we're good on this. I'll fix the pnpm conflict later today/tonight and then we'll merge. 🍻

shellscape:
@Andarist no matter what I try locally, I cannot get this to build:

```
./src/index.js → dist/index.js, dist/index.es.js...
[!] (plugin babel) Error: [BABEL] /Users/powella/code/rollup/plugins/packages/babel/src/index.js: Could not find plugin "proposal-numeric-separator". Ensure there is an entry in ./available-plugins.js for it. (While processing: "/Users/powella/code/rollup/plugins/packages/babel/node_modules/@babel/preset-env/lib/index.js")
```

Please take a look.

Andarist:
@shellscape fixed

shellscape:
@Andarist I noticed that 5.0.0-alpha.2 was the latest version on rollup-plugin-babel, but the plugin here is only 5.0.0-alpha.1. I also didn't see that you were in an alpha stage, so I'm not sure how to release this because we should release a stable version to archive rollup-plugin-babel.

I guess this is my fault for not looking at the old repo first. Please advise.

Andarist:
Oh, sorry for not mentioning this recently - my intention was this to be a major bump, so just release this as 5.0.0

shellscape:
No worries at all. I'll do that tonight.

enjikaka:
Is this broken again? https://github.com/rollup/rollup-plugin-babel/issues/360

I just tried making two outputs in Rollup, both with a plugin field with babel in them. Getting the same errors as in the issue above;
```
(!) The "load" hook used by the output plugin babel is a build time hook and will not be run for that plugin. Either this plugin cannot be used as an output plugin, or it should have an option to configure it as an output plugin.
(!) The "options" hook used by the output plugin babel is a build time hook and will not be run for that plugin. Either this plugin cannot be used as an output plugin, or it should have an option to configure it as an output plugin.
(!) The "resolveId" hook used by the output plugin babel is a build time hook and will not be run for that plugin. Either this plugin cannot be used as an output plugin, or it should have an option to configure it as an output plugin.
```

Andarist:
> Is this broken again?

We don't know as you haven't really provided a repro case - there is a non-zero chance that you made some mistake. You could also try to actually debug the problem - the code is all there, on your machine, ready to be investigated.

stevage:
I have made a repro case that shows the same thing: https://replit.com/@stevage/rollup-plugin-repro

Andarist:
@stevage you are using the input plugin in the output plugin's position. Please refer to this doc if you want to use the output plugin: https://github.com/rollup/plugins/blob/master/packages/babel/README.md#running-babel-on-the-generated-code

PR #115, opened 2019-12-25
Add typings to `run`

cprecioso:
## Rollup Plugin Name: `run`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description
Adds TypeScript types for the `@rollup/plugin-run` plugin, as well as links to it from the README.

### DIFF
--- a/packages/run/index.d.ts
+++ b/packages/run/index.d.ts
@@ -1,13 +1,8 @@
+import { ForkOptions } from 'child_process'
 import { Plugin } from 'rollup'
-
-interface RollupRunOptions {
-  /**
-   * Pass options through to `child_process.fork(...)`
-   */
-  execArgv?: string[]
-}
 
 /**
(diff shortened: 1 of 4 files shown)


shellscape:
@cprecioso Thanks for this!

@NotWoods do we want to accept this or wait on it until we convert to TypeScript? Seems no harm in adding this now. The only caveat is that if the types change with the conversion, we'll need to bump major.

packages/run/index.d.ts
```diff
@@ -0,0 +1,13 @@
+import { Plugin } from 'rollup'
+
+interface RollupRunOptions {
```
NotWoods: The options are all equivalent to `import('child_process').ForkOptions`. It might be better to just use that type.

PR #116, opened 2019-12-25
fix(pluginutils): fix makeLegalIdentifier using potentially unsafe input for computed blacklisted identifier

Andarist:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

https://github.com/rollup/rollup-plugin-babel/issues/343

### Description

fix makeLegalIdentifier using potentially unsafe input for computed blacklisted identifier


packages/pluginutils/test/makeLegalIdentifier.ts
```diff
@@ -17,3 +17,7 @@ test('blacklists arguments (https://github.com/rollup/rollup/issues/871)', (t) =
 test('empty', (t) => {
   t.is(makeLegalIdentifier(''), '_');
 });
+
+test('handles input evaluated to blacklisted identifier', (t) => {
```
Andarist: this kinda mentions the internal implementation, but not sure how to avoid it as the bug was caused by the way this was implemented

PR #121, opened 2019-12-26
fix(commonjs): module.exports object spread

matthewp:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

The follow scenario was throwing previously:

```js
const obj = { a: 'b' }

module.exports = {
  ...obj
};
```

This is because the transformer looks for named exports on the
`module.exports` object, however it was not accounting for a spread.
This fixes that scenario.

### Error

This is the error that is displayed without this fix:

```
'Cannot read property \'type\' of undefined'
```

With the stack of:

```
 forEach (src/transform.js:291:43)
  Object.enter (src/transform.js:290:33)
  visit (/Users/matthew/dev/rollup-plugins/node_modules/.pnpm/registry.npmjs.org/estree-walker/0.6.1/node_modules/estree-walker/src/estree-walker.js:22:9)
  visit (/Users/matthew/dev/rollup-plugins/node_modules/.pnpm/registry.npmjs.org/estree-walker/0.6.1/node_modules/estree-walker/src/estree-walker.js:44:4)
  visit (/Users/matthew/dev/rollup-plugins/node_modules/.pnpm/registry.npmjs.org/estree-walker/0.6.1/node_modules/estree-walker/src/estree-walker.js:39:34)
  Object.walk (/Users/matthew/dev/rollup-plugins/node_modules/.pnpm/registry.npmjs.org/estree-walker/0.6.1/node_modules/estree-walker/src/estree-walker.js:2:2)
  transformCommonjs (src/transform.js:173:3)
  Object.call (src/index.js:92:25)
  Object.transform (src/index.js:170:48)
```

### Output

Prior to this fix there is no output because rollup throws due to the above error message.


packages/commonjs/test/test.js
```diff
@@ -737,3 +737,29 @@ test('normalizes paths used in the named export map', async (t) => {
...
+
+test('can spread an object into module.exports', async (t) => {
+  const bundle = await rollup({
+    input: 'fixtures/samples/module-exports-spread/main.js',
+    plugins: [
+      commonjs()
+    ]
+  });
+  const code = await getCodeFromBundle(bundle);
+  t.is(
```
shellscape: please use `t.snapshot()` here.

matthewp: This is done.

matthewp:
@shellscape I updated the description per your request and also changed to use t.snapshot. Thanks for the review!

packages/commonjs/src/transform.js
```diff
@@ -288,7 +288,7 @@ export function transformCommonjs(
 
         if (flattened.keypath === 'module.exports' && node.right.type === 'ObjectExpression') {
           node.right.properties.forEach((prop) => {
-            if (prop.computed || prop.key.type !== 'Identifier') return;
+            if (prop.computed || !('key' in prop) || prop.key.type !== 'Identifier') return;
```
wuweiweiwu: nit:

I think this is more readable if we do something like `'key' in prop && `

matthewp: That wouldn't do what we want, if `key` is not in `prop` it would not satisfy this condition and would proceed to the next line. We want this condition to be true for `SpreadElement`.

shellscape:
thanks!

matthewp:
Thank you for taking the time to review this PR, especially around holiday time. A lot of respect to you @shellscape and @wuweiweiwu

matthewp:
@shellscape Is it my responsibility to submit a release for this? It's unclear to me from reading the docs..

shellscape:
We're still working on smoothing out releases. I'll manually publishing later today.

matthewp:
Hey @shellscape, any chance we can get a commonjs release today?

PR #122, opened 2019-12-26
Use typescript in typescript plugin

NotWoods:

This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

Refactoring the TypeScript plugin to use TypeScript itself. This is to prepare for larger refactors to add new features down the line.

This also fixes the types for options.tslib to reflect what is actually supported.


packages/typescript/rollup.config.js
```diff
@@ -1,21 +1,15 @@
-import buble from '@rollup/plugin-buble';
+import typescript from '@rollup/plugin-typescript';
```
pnevares: Shouldn't this be a relative import from local code?

packages/typescript/package.json
```diff
@@ -49,17 +49,25 @@
   },
   "dependencies": {
     "@rollup/pluginutils": "^3.0.0",
-    "resolve": "^1.12.2"
+    "resolve": "^1.13.1"
   },
   "devDependencies": {
     "@rollup/plugin-buble": "^0.20.0",
+    "@rollup/plugin-typescript": "^2.0.1",
```
pnevares: This definitely feels unusual 🤔

packages/typescript/rollup.config.js
NotWoods: Making it a relative import depends on the code already being compiled in order to compile it. We kind of follow this pattern already as we don't depend on relative packages in the monorepo, instead depending on the npm version of pluginutils, babel, etc.

pnevares: Thanks, I must have never noticed! 👍

shellscape:
Looks like an eslint comment is needed for a fixture to pass linting

PR #127, opened 2019-12-31
Fix default include option regex

yannayl:
## Rollup Plugin Name: strip

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

#128

### Description
Default include did not match relative module ids. Module ids are not always resolved when invoking the transformation step (e.g. when using the alias plugin). Leaving the default include undefined leads to a consistent behavior.

### DIFF
--- a/packages/strip/src/index.js
+++ b/packages/strip/src/index.js
@@ -32,8 +32,7 @@ function flatten(node) {
 }
 
 export default function strip(options = {}) {
-  const include = options.include || '**/*.js';
-  const { exclude } = options;
+  const { include, exclude } = options;
   const filter = createFilter(include, exclude);
   const sourceMap = options.sourceMap !== false;
 


shellscape:
Thanks for the PR. I don't believe this is the right fix, and though the tests aren't failing, this may cause issues with default behavior. See https://github.com/rollup/rollup-plugin-strip/pull/7 for history on why the code was added, which this PR proposes to remove.

yannayl:
I  wasn't sure it's the right patch either but thought it's a first step in the right direction. There is something off with the minimatch pattern, in my project I resort to using a regex `/.*\.js/`

About the tests - I haven't studied the project enough so I don't know yet how to add them. When I think I have the correct solution I might add them, but honestly, using the regex solves the problem so I'm not sure I'll keep investigating.

shellscape:
I'm going to close this one _for now_ only for housekeeping. Let's continue to work on the cause in #128. Please don't delete your fork in the mean time.

Issue #69, opened 2019-12-02
auto-install yarn tests


shellscape:
- Rollup Plugin Name: auto-install
- Rollup Plugin Version: latest

### Expected Behavior / Situation

I'd really enjoy it if the tests passed consistently. I frequently have to rerun the Circle CI workflows because of failed `yarn` tests. 

### Actual Behavior / Situation

☝️ same same

### Modification Proposal

I'd like to nuke the yarn test that is failing intermittently. If someone can spot a way to get it working 100% of the time, I'm all for it. 

Thoughts?

/cc @rollup/plugin-contributors @rollup/plugin-maintainers

Andarist:
While mocking is often not the right answer it seems that it might be here. We don't actually care how yarn (or any other package manager) installs stuff (what artifacts does it leave behind) - we could only check if it has been invoked correctly and trust it to do its job fine when invoked for real.

shellscape:
That's very apt, and sounds like the right way to go.

PR #135, opened 2020-01-02
chore: change workspace packages pattern

Andarist:
### Description

I've noticed that the currently used pattern was overly greedy and has classified such things as potential packages:
```
packages/auto-install/test/fixtures/npm
packages/auto-install/test/fixtures/yarn
packages/auto-install/test/fixtures/yarn-bare
packages/json/test/fixtures/named
packages/node-resolve/test/fixtures
```

Things were working, but including those as workspace packages was not intentional so better to fix that.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,8 +1,8 @@
 importers:
   .:
     devDependencies:
-      '@typescript-eslint/eslint-plugin': 2.12.0_66854ba8971b2256e3dae1b103509efe
-      '@typescript-eslint/parser': 2.12.0_typescript@3.7.3
+      '@typescript-eslint/eslint-plugin': 2.14.0_c75538670b52ef0bc80e769266babcdc
+      '@typescript-eslint/parser': 2.14.0_typescript@3.7.4
       ava: 2.4.0
       chalk: 2.4.2
       codecov-lite: 0.3.1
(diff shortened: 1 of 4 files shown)


pnpm-lock.yaml
```diff
@@ -68,12 +68,6 @@ importers:
       node-noop: ^1.0.0
       rollup: ^1.20.0
       rollup-plugin-node-resolve: ^5.2.0
-  packages/auto-install/test/fixtures/npm:
```
Andarist: you can actually see how those were incorrectly included here

pnpm-lock.yaml
```diff
@@ -1,8 +1,8 @@
 importers:
```
Andarist: I have no idea why there are so many lockfile changes. I've just executed `pnpm i` after fixing that workspace option - which had to be done, because it affects lockfile. Not sure why any versions here were changed though.

Andarist:
A little strange situation - some snapshots have failed due to end of line difference (classic crlf vs lf). It works on master though and I haven't touched those files as part of this PR. Any ideas? My guess is that some dependency has changed but I don't know why they have changed in the first place.

shellscape:
When you changed the scope and removed the (incorrectly) included packages, pnpm recognized that and cleaned up the lockfile. I don't know why your changes are now causing issues in snapshots, because that doesn't happen on other branches.

Andarist:
I've "fixed" the problem by normalizing yarn-written files on our end (maybe yarn version has changed on CI in the meantime? although can't find any changelog entry in yarn about a change that could affect this).

shellscape:
This goes back to https://github.com/rollup/plugins/issues/69 and your comment about mocking yarn out. I'm OK if we want to put that off to another PR, but let's put a comment above the changed test lines to explain why that's there. Otherwise it's going to look superfluous to someone not aware of this PR.

PR #136, opened 2020-01-03
fix(image): don't encode svg as base64

### LINKED ISSUE #49: base64 svg may end up in bigger size
In this example of 2 data strings the base64 is bigger.

http://codepen.io/anon/pen/ZONxYR

Would it may make sense to not base64 svg?

shellscape:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

Fixes #49 

### Description

This change renders svg as utf-8 rather than base64, which is larger in size when encoding an svg.


packages/image/rollup.config.js
```diff
@@ -4,7 +4,7 @@ import pkg from './package.json';
 
 export default {
   input: 'src/index.js',
-  external: Object.keys(pkg.dependencies),
+  external: [...Object.keys(pkg.dependencies), ...['fs', 'path']],
```
NotWoods: For [other plugins](https://github.com/rollup/plugins/blob/master/packages/inject/rollup.config.js#L5) we've followed the format of having `external` be a separate variable.

```js
const external = Object.keys(pkg.dependencies).concat('fs', 'path');
```

packages/image/src/index.js
```diff
@@ -36,8 +36,9 @@ export default function image(opts = {}) {
         return null;
       }
 
-      const source = readFileSync(id, 'base64');
-      const data = `'data:${mime};base64,${source}'`;
+      const format = mime.includes('svg') ? 'utf-8' : 'base64';
```
NotWoods: I think it'd be better to be precise with the mime type.
```suggestion
      const format = mime === 'image/svg+xml' ? 'utf-8' : 'base64';
```
or
```suggestion
      const format = mime === mimeTypes['.svg'] ? 'utf-8' : 'base64';
```

packages/image/rollup.config.js
shellscape: there's functionally no difference there, and really no readability difference either. I don't think this is a spot we need to enforce a pattern.

packages/image/src/index.js
shellscape: Good call here.

packages/image/rollup.config.js
NotWoods: In that case let's remove the unnessecary spread:
```suggestion
  external: [...Object.keys(pkg.dependencies), 'fs', 'path'],
```

shellscape: That's a neat trick. And this whole time I thought I had to spread anything behind a spread.

neoel:
This breaks setups where xml is being escaped. 
Like [here](https://github.com/sveltejs/svelte/issues/4302) in svelte

naggie:
This also forces the usage of an embedded SVG instead of an `img` tag resulting in problems where you can't constrain the size of the SVG necessarily, as the behavior  of `img` tags and `svg` tags is different on the DOM, and also dependent on what width/height viewbox is set. 

Perhaps ideally this should be optional, or not done at all. The ~30% encoding overhead is mitigated by gzip often, and it it really matters (for instance you need caching) then using an external image is more appropriate anyway.

PR #137, opened 2020-01-03
feat(yaml): add documentMode, safe options

### LINKED ISSUE #39: Provide a `safeLoadAll` option
Hi, I have a stream which has multiple documents inside e.g:
```yaml
---
kind: x
apiVersion: v1
spec:
  selector: example1
---
kind: x
apiVersion: v1
spec:
  selector: example2
---
kind: x
apiVersion: v1
spec:
  selector: example3
```

in the plugin you have:

```
let data = YAML.load(yaml);
```

but this causes this error when loading my file `(plugin yaml) YAMLException: expected a single document in the stream, but found more`.

For my workflow I need an option to have it loaded as 

```
let data = YAML.safeLoadAll(yaml);
```
This will load both single and multiple document streams

thanks in advance 🙌🏻

shellscape:
## Rollup Plugin Name: `yaml`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

Fixes #39. This PR adds support for multiple document streams in a single YAML file, as well as the ability to load YAML using `safeLoad` and `safeLoadAll`, rather than the default `load`.

/cc @lufego


shellscape:
Whew finally got this one sorted. It seems that the containers that Github Actions uses has changed somewhat in the last week, and directory paths are now windows standard paths and had to be escaped.

shellscape:
24 hours before I merge this badboy. Please review!

packages/yaml/src/index.js
```diff
@@ -1,11 +1,29 @@
...
 
+const defaults = {
+  documentMode: 'single',
+  safe: true,
+  transform: null
+};
 const ext = /\.ya?ml$/;
 
-export default function yamll(options = {}) {
+export default function yamll(opts = {}) {
```
NotWoods: Since `opts` is just used with `Object.assign` we can remove the default value.
```suggestion
export default function yamll(opts) {
```

shellscape: This is a stylistic choice. We don't lose anything by defining the default value there.

PR #84, opened 2019-12-09
fix(pluginutils): Escape glob characters in folder

NotWoods:
## Rollup Plugin Name: `pluginutils`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] yes (_bugfixes and features will not be merged without tests_)
- [ ] no

Breaking Changes?

- [ ] yes (_breaking changes will not be merged unless absolutely necessary_)
- [x] no

List any relevant issue numbers:

### Description

Fixes #80

Escapes some special characters for globs that can appear in paths.

PR #141, opened 2020-01-04
fix(pluginutils): createFilter Windows regression

shellscape:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

Something recently changed with the Windows containers on Github Actions and a few tests that were passing were suddenly not. _(I suspect a bugfix on their end, perhaps something to do with WSL)_ **But** that revealed a regression in the code that was added in #84 for a fix for paths that included parens.

The fix for this was a refactor of that code. Additionally, several more characters that could be used in globs are now escaped.


shellscape:
Pushing this one out, as it's a pretty big regression.

PR #146, opened 2020-01-07
fix(url): windows compatible, tests

shellscape:
## Rollup Plugin Name: `url`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

As with a few other plugin tests, the recent change in Github Actions' Windows image/pipeline was causing `@rollup/plugin-url` to fail tests on Windows. After looking into it, the plugin would have been fairly unusable on Windows. Previously, the tests were passing the Windows Github Actions - somehow. This change addresses Windows file paths and normalizes paths that need to be.

Because this is a critical flaw, this will be merged with or without review after 12 hours.

### DIFF
--- a/packages/url/src/index.js
+++ b/packages/url/src/index.js
@@ -9,7 +9,7 @@ import { createFilter } from 'rollup-pluginutils';
 
 const fsStatPromise = util.promisify(fs.stat);
 const fsReadFilePromise = util.promisify(fs.readFile);
-const { sep } = path;
+const { posix, sep } = path;
 const defaultInclude = ['**/*.svg', '**/*.png', '**/*.jpg', '**/*.gif'];
 
 export default function url(options = {}) {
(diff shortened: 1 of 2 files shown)


packages/url/src/index.js
```diff
@@ -9,7 +9,7 @@ import { createFilter } from 'rollup-pluginutils';
 
 const fsStatPromise = util.promisify(fs.stat);
 const fsReadFilePromise = util.promisify(fs.readFile);
-
+const { sep } = path;
```
wuweiweiwu: nit: I prefer using `path.sep` so I know its from the node package. but no strong opinions here

packages/url/src/index.js
```diff
@@ -47,17 +47,19 @@ export default function url(options = {}) {
...
           const outputFileName = fileName
             .replace(/\[hash\]/g, hash)
             .replace(/\[extname\]/g, ext)
-            .replace(/\[dirname\]/g, `${relativeDir}/`)
+            // use `sep` for windows environments
+            .replace(/\[dirname\]/g, `${relativeDir}${sep}`)
             .replace(/\[name\]/g, name);
-          data = `${publicPath}${outputFileName}`;
+          // Windows fix - exports must be in unix format
+          data = `${publicPath}${outputFileName.split(sep).join('/')}`;
```
wuweiweiwu: should we pull this `/` into a constant?

NotWoods: Could also use `path.posix.sep`

shellscape: `path.posix.sep` is a solid replacement.

packages/url/src/index.js
shellscape: let's address this when we do the TS conversion next month

PR #150, opened 2020-01-08
chore(repo): new plugin data-uri

### LINKED ISSUE #125: plugin request: Handle data:mime,<data> urls
Request Checklist:

- [x] I have searched https://npmjs.com/ before opening this issue
- [x] I have searched https://github.com/rollup/awesome before opening this issue

### New Plugin Use Case

I would like to import code as a module from a `data:{mime}` declaration. 

```js
import 'data:text/javascript, console.log('Buhhh')'
import me from 'data:application/json, { "my": 'json''}
import('data:application/json, { "my": 'json''})
``` 

### New Plugin Proposal

Perhaps `@rollup/plugin-data-mime`. The plugin could look like:

```js
{
  resolveId(id) {
    if (isDataUrl(id)) {
      return id;
    }
    return null;
  },
  load(id) {
    if (isDataUrl(id)) {
      return getCodeFromDataUrl(id);
    }
    return null;
  }
}
```

shellscape:
## Rollup Plugin Name: `data-uri`

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: Fixes #125

### Description

This plugin allows importing from `data:<mime>, <content` urls. For example:

`import 'data:text/javascript, export const batman = true;'`


packages/data-uri/README.md
```diff
@@ -0,0 +1,57 @@
...
+```console
+npm install @rollup/plugin-data-uri --save-dev
+```
+
+## Usage
+
+Create a `rollup.config.js` [configuration file](https://www.rollupjs.org/guide/en/#configuration-files) and import the plugin:
+
+```js
+const dataUri = require('@rollup/plugin-data-uri');
```
NotWoods: Should use ES modules

packages/data-uri/src/index.ts
```diff
@@ -0,0 +1,81 @@
...
+import { dataToEsm } from '@rollup/pluginutils';
+
+const reDataUri = /^([^/]+\/[^;,]+)(;base64)?,([\s\S]*)$/;
+const mimeTypes = {
+  js: 'text/javascript',
+  json: 'application/json'
+};
+
+export default function dataUri(): Plugin {
+  const resolved: { [key: string]: any } = {};
```
NotWoods:
```suggestion
  const resolved: { [key: string]: { mime?: string, content?: string } } = {};
```

We might want to use a `Map` instead as well.

packages/data-uri/src/index.ts
```diff
@@ -0,0 +1,81 @@
...
+        return content;
+      } else if (mime === 'application/json') {
+        let json = '';
+        try {
+          json = JSON.parse(content);
+        } catch (e) {
+          const error: RollupError = {
+            message: e.toString(),
+            parserError: e,
+            plugin: '@rollup/plugin-data-uri',
```
NotWoods: I think this field gets [set by rollup automatically](https://github.com/rollup/rollup/blob/master/src/utils/pluginUtils.ts#L18)

packages/data-uri/README.md
```diff
@@ -0,0 +1,57 @@
...
+[size]: https://packagephobia.now.sh/badge?p=@rollup/plugin-data-uri
+[size-url]: https://packagephobia.now.sh/result?p=@rollup/plugin-data-uri
+
+[![npm][npm]][npm-url]
+[![size][size]][size-url]
+[![libera manifesto](https://img.shields.io/badge/libera-manifesto-lightgrey.svg)](https://liberamanifesto.com)
+
+# @rollup/plugin-data-uri
+
+🍣 A Rollup plugin which imports modules from Data URIs.
```
NotWoods: We should document which data URIs are supported.

packages/data-uri/types/index.d.ts
```diff
@@ -0,0 +1,3 @@
+import { Plugin } from 'rollup';
+
+export default function dataUri(): Plugin;
```
NotWoods: nit: It would be nice to have comments for IDEs/VS code
```suggestion
/**
 * A Rollup plugin which imports modules from Data URIs.
 */
export default function dataUri(): Plugin;
```

LarsDenBakker:
I think it looks good, just to double check: this supports base 64 encoded urls as well? Maybe it's good to add a test for it.

packages/data-uri/src/index.ts
shellscape: but why

not being snarky whatsoever - the only reason I can for see for using Map here is for the sake of using Map. this is never going to be large enough (someone prove me wrong, I dare you) to have perf implications (we're not running in a browser) and there's no immediate benefit of `Map.has()` over `!!object[key]`. New school might say it's more readable but then fundamentals crowd might say they're equal.

packages/data-uri/src/index.ts
shellscape: It's superfluous but not harmful.

shellscape:
@LarsDenBakker base64 support added

PR #151, opened 2020-01-09
fix(pluginutils): keep for(const..) out of scope

### LINKED ISSUE #113: attachScopes doesn't recognize for loop correctly
- Rollup Plugin Name: pluginutils
- Rollup Version:
- Operating System (or Browser): Windows 7
- Node Version: 13.5.0

### How Do We Reproduce?

```js
const {parse} = require("acorn");
const {attachScopes} = require("@rollup/pluginutils");

const code = `
for (const foo of []) {
  const bar = 1;
  var baz = 2;
}
for (var bak of []);
`;

const ast = parse(code);
let scope = attachScopes(ast, "scope");
console.log(scope.contains("foo"));
console.log(scope.contains("bar"));
console.log(scope.contains("baz"));
console.log(scope.contains("bak"));
```

https://repl.it/@eight041/attachScopes-and-for-scope?language=nodejs

### Expected Behavior

```
false
false
true
true
```

### Actual Behavior

```
true
false
true
true
```

I'm a little surprised that we didn't find this bug earlier. This breaks a lot of stuff IMHO.

shellscape:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: Fixes #113

### Description

As mentioned in #113, in `for(const foo of y)`, `foo` was leaking into program scope when it should not have been.


.github/workflows/node-windows.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       run: npm install pnpm -g
     - name: pnpm install
       run: pnpm install --ignore-scripts
-    - name: run linting
-      run: pnpm run ci:lint
+    # - name: run linting
```
wuweiweiwu: is this intentional?

PR #152, opened 2020-01-09
chore(commonjs): remove jsnext:main

### LINKED ISSUE #83: Remove jsnext:main support from repo
- Rollup Plugin Name: `node-resolve`, all others
- Rollup Plugin Version: n / a

n / a 

### Actual Behavior / Situation

After @TrySound mentioned that `jsnext:main` is deprecated, I did some digging and asking around. Rollup was really the only bundler that embraced it fully. As mentioned by Rich Harris [here](https://twitter.com/Rich_Harris/status/1203299929285971968), `jsnext:main` has been superseded by `module`, and there are numerous sources that consider it deprecated as a result ([for example](https://github.com/stereobooster/package.json#jsnextmain)).

Basically it never made it out of discussion, and Rollup supported it out of necessity at the time it was implemented.

### Modification Proposal

So all that goes to say is that we have sufficient evidence to remove support from `@rollup/node-resolve` and to remove that field from any plugins in the repo.

shellscape:

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

List any relevant issue numbers: Fixes #83

### Description

This is the last of the `jsnext:main` package.json entries that needs to be cleaned up. It's not technically breaking since the `module` field has existed beside it for several versions.

### DIFF
--- a/packages/commonjs/package.json
+++ b/packages/commonjs/package.json
@@ -85,7 +85,6 @@
       "!**/types.ts"
     ]
   },
-  "jsnext:main": "dist/index.es.js",
   "module": "dist/index.es.js",
   "types": "types/index.d.ts"
 }


shellscape:
@wuweiweiwu https://github.com/shellscape/prettier-plugin-package is the prettier plugin we're using to standardize package.json, so that would be the spot to raise this. right now, `module` isn't part of the package.json standard properties, and non-standard props are moved to the bottom and alphabetized.

PR #155, opened 2020-01-09
refactor(node resolve): clean codebase and fix external warnings

### LINKED ISSUE #74: Add only option to external
- Rollup Plugin Name:  node-resolve
- Rollup Plugin Version: 6.0.0

### Feature Use Case
Transfer from https://github.com/rollup/rollup-plugin-node-resolve/issues/245

### Feature Proposal
Transfer from https://github.com/rollup/rollup-plugin-node-resolve/issues/245

shellscape:

This PR contains:

- [x] bugfix
- [x] feature
- [x] refactor
- [x] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] no

List any relevant issue numbers: Fixes #74

### Description

This PR gives `node-resolve` a refactor and reorganization that should make it easier to work with moving forward. This also fixes the problem with issue #74.

#### Breaking Changes

- The `jsnext`, `main`, and `module` options have been removed. (deprecated in previous major)
- The `only` option is deprecated in favor of the `resolveOnly` option - they are identical in functionality but `resolveOnly` is far more descriptive.


shellscape:
So I've narrowed down the source of the failing test. It seems to be `@rollup/pluginutils` with anything higher than version`3.0.2`. I'm baffled as to why though, because the changes to 3.0.3 were to a function unrelated to the failures of the tests.

To verify, checkout `master`, run `pnpm install` from root, and `cd packages/node-resolve` then `pnpm run test`. The tests should fail. Then `pnpm add @rollup/pluginutils@3.0.2` and `pnpm run test` again - they should all succeed. 

My only thought is a transitive dependency. I need to get some more eyes on this because I'm having trouble spotting it. @NotWoods @pnevares @wuweiweiwu @Kocal

FredKSchott:
Hey yall, just a heads up that this introduced a breaking regression in Snowpack where we were passing `dedupe: undefined` when no dedupes were defined.  You can see this break here: https://github.com/rollup/plugins/blame/master/packages/node-resolve/src/index.js#L48

We'll fix this upstream to pass `[]` instead, but you may want to look at your defaults logic to not overwrite values with undefined.

shellscape:
@FredKSchott we're also going to be putting option validation in place in the near future to try and mitigate this kind of thing. Here's the culprit, and why your `undefined` value was replaced by a default: https://github.com/rollup/plugins/blame/master/packages/node-resolve/src/index.js#L30

bjornharrtell:
Seems I also have encountered a regression from 7.0.0 to 7.1.0 causing a incomplete build with `Missing global variable name` for a local module. Trying to find a better description/reproduction of this problem.

arantes555:
@bjornharrtell I have provided a minimal reproduction of the issue in my report for #209 : https://github.com/arantes555/node-resolve-local-files.git

PR #156, opened 2020-01-09
allow options.only regexp from other realms

jdalton:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

The use of the regexp check `o instanceof RegExp` cause issues with our config since it's coming from a different realm. By using a more robust regexp check regexes from other realms may be used for `options.only`.


shellscape:
Thanks for the PR. Can you go into more depth as to what the other "realms" are you're speaking of, and how the `instance of` check is insufficient? 

The timing of this one is just a tad unfortunate as we're working on a rather comprehensive refactor of the plugin in #155. We'll have to hold off on merging this until that's complete, hopefully soon.

jdalton:
> Thanks for the PR. Can you go into more depth as to what the other "realms" are you're speaking of, and how the instance of check is insufficient?

Code running in a new Node `vm` context will have a different `RegExp` constructor than the one of the primary context (aka realm). In this case our config is coming from another realm which means the regexp object created while still a regexp will fail the `o instanceof RegExp` check. The unit test create for this PR is a simplified repro.

shellscape:
Thanks for the extended explanation. PR looks good, we're just waiting on #155. I'll help update this PR once that lands, since there will be some conflicts.

shellscape:
I'm almost wrapped up with an onsite and will be able to move on this and other PRs this coming week.

shellscape:
Forgive the git shenanigans, should have this back open in another moment.

shellscape:
@jdalton I think I've run into a github bug of some kind. After resetting the branch even with upstream/master I'm no longer able to push to the branch. Here's a summary of where the changes to need to go:

```js
// index.js:6
import isRegex from 'is-regex';

// index.js:54-60
  const resolveOnly = options.resolveOnly.map((pattern) => {
    if (isRegex(pattern)) {
      return pattern;
    }
    const normalized = pattern.replace(/[\\^$*+?.()|[\]{}]/g, '\\$&');
    return new RegExp(`^${normalized}$`);
  });

// test/only.js: add at line 49
test('regex from another realm', async (t) => {
  const regex = runInNewContext('/^@scoped\\/.*$/');
  const warnings = [];
  const bundle = await rollup({
    input: 'only.js',
    onwarn: (warning) => warnings.push(warning),
    plugins: [
      nodeResolve({
        only: [regex]
      })
    ]
  });
  const imports = await getImports(bundle);

  t.is(warnings.length, 1);
  t.snapshot(warnings);
  t.deepEqual(imports, ['test']);
});
```

```console
$ cd plugins/packages/node-resolve
$ pnpm add is-regex
$ pnpm run test
```

Add the changed files and commit, and we should be good.

PR #157, opened 2020-01-10
fix(typescript) extended config file path

hasangenc0:
This PR contains:
- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?
- [x] yes (bugfixes and features will not be merged without tests)
- [ ] no

Breaking Changes?
- [ ] yes (breaking changes will not be merged unless absolutely necessary)
- [x] no

### Description
This plugin has `extends` option support  but plugin does not give the right extended config file path.

For example: If your tsconfig contains `"extends": "../generic-tsconfig.json"` typescript plugin looks parent directory of project root folder because of the dots.
I added proper path resolver to fix this bug.

UPDATE: Bug happens when tsconfig option given.

### DIFF
--- a/packages/typescript/src/options.ts
+++ b/packages/typescript/src/options.ts
@@ -24,8 +24,8 @@ export function readTsConfig(ts: typeof import('typescript'), tsconfigPath: stri
   if (!tsconfig.config || !tsconfig.config.compilerOptions) return { compilerOptions: {} };
 
   const extendedTsConfig: string = tsconfig.config.extends;
-  if (extendedTsConfig) {
-    tsconfig.config.extends = join(process.cwd(),existingTsConfig, '..', extendedTsConfig );
+  if (tsconfigPath && extendedTsConfig) {
+    tsconfig.config.extends = join(process.cwd(),existingTsConfig, '..', extendedTsConfig);
   }
 
   return tsconfig.config;
--- a/packages/typescript/src/options.ts
+++ b/packages/typescript/src/options.ts
@@ -1,5 +1,5 @@
 import { readFileSync } from 'fs';
-import { join } from 'path';
+import { resolve } from 'path';
 
 export function getDefaultOptions() {
   return {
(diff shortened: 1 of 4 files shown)


shellscape:
Thanks for the PR. In order to review this, please get CI checks passing. To approve this, we'll need you to add tests to show that your change has the intended consequences.

hasangenc0:
> Thanks for the PR. In order to review this, please get CI checks passing. To approve this, we'll need you to add tests to show that your change has the intended consequences.

I added tests.

packages/typescript/src/options.ts
```diff
@@ -17,9 +18,16 @@ export function readTsConfig(ts: typeof import('typescript'), tsconfigPath: stri
...
     return {};
   }
+
   const tsconfig = ts.readConfigFile(existingTsConfig, (path) => readFileSync(path, 'utf8'));
 
   if (!tsconfig.config || !tsconfig.config.compilerOptions) return { compilerOptions: {} };
+
+  const extendedTsConfig: string = tsconfig.config.extends;
+  if (tsconfigPath && extendedTsConfig) {
+    tsconfig.config.extends = join(process.cwd(),existingTsConfig, '..', extendedTsConfig);
```
NotWoods: Could you tweak `parseCompilerOptions` instead, where the extends path is resolved? 

nit: maybe use `resolve` instead of `join`

NotWoods:
Looks good, I just have a nit about moving the code closer to where extends get processed :)

packages/typescript/src/options.ts
hasangenc0: Actually this bug happens when `tsConfig` option exists at plugin's options so, `parseCompilerOptions` does not relative with plugin's options, it is relative with the config file that read at `readTsConfig`.

Also, you are right because `parseCompilerOptions` should make this. But like I said before, this function doesn't know about `tsConfig` path is given.

I could write a function to make this but it would cause performance issues.

But I did the `resolve` thing 😄.

Issue #167, opened 2020-01-18
"firstpass" regex causes labels not to be removed (@rollup/plugin-strip)


trbrc:
- Rollup Plugin Name: @rollup/plugin-strip
- Rollup Plugin Version: 1.3.1
- Rollup Version: 1.28.0
- Node Version: 12.13.1

### How Do We Reproduce?

REPL.it reproduction: https://repl.it/repls/KookyMicroGui

### Expected Behavior

All labels `y` should be removed, as specified in the `strip` config in `rollup.config.js`.

### Actual Behavior

The `y` label in `input/fails.js` is not removed.

---

The bug is this check, which causes an early return when processing `fails.js`: https://github.com/rollup/plugins/blob/d3db3bbdd0981bb5c5a830f23c7b7dd730a17a66/packages/strip/src/index.js#L55

Which happens because the `firstpass` regex only looks for the function names, not labels:
https://github.com/rollup/plugins/blob/d3db3bbdd0981bb5c5a830f23c7b7dd730a17a66/packages/strip/src/index.js#L47

In this case, the config is falling back on the default `{functions: ['console.*', 'assert.*']}`. Since `fail.js` doesn't contain `console` or `assert` or `debugger`, the plugin incorrectly skips processing the labels.

shellscape:
Thanks for reporting this and for the great reproduction and info. We'll get on it as soon as possible. If you have an immediate need and would like to PR a fix, we'd be happy to review one!

trbrc:
I did a PR with what I think is the right fix, but would be grateful if someone who is already familiar with the test setup would add a test.

In the mean time I found that it's easy to work around. Simply use the plugin twice, once for the labels, once for the functions, and make sure to specify `functions: []` to override the default for the instance that strips labels.

(As an aside, if I may: I think the `['console.*', 'assert.*']` default is a pretty bad idea, at least if the config specifies `labels`. It's very easy to overlook. It certainly made this particular issue a lot more difficult to troubleshoot, as it made it appear as if labels were being stripped or kept at random.)

shellscape:
Accompanying PR was abandoned.

PR #169, opened 2020-01-18
Add labels to `firstpass` regex (@rollup/plugin-strip)

trbrc:
## Rollup Plugin Name: `strip`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [ ] no
- [x] https://xkcd.com/1172/

List any relevant issue numbers: #167

### Description

This adds the labels to the `firstpass` regex. I also removed the `functions.length > 0` check which felt a little odd when the regex also looks for `debugger` and labels.

I'm marking this PR as a draft as it doesn't include tests. I tried this fix in my local repo and it has the right behavior with my config and code. Someone more familiar with the testing setup in the `rollup/plugins` repo will probably get the tests right faster.

### DIFF
--- a/packages/strip/src/index.js
+++ b/packages/strip/src/index.js
@@ -52,7 +52,7 @@ export default function strip(options = {}) {
 
     transform(code, id) {
       if (!filter(id)) return null;
-      if (functions.length > 0 && !firstpass.test(code)) return null;
+      if (!firstpass.test(code)) return null;
 
       let ast;
 
--- a/packages/strip/src/index.js
+++ b/packages/strip/src/index.js
@@ -44,7 +44,7 @@ export default function strip(options = {}) {
 
   const labels = options.labels || [];
 
-  const firstpass = new RegExp(`\\b(?:${functions.join('|')}|debugger)\\b`);
+  const firstpass = new RegExp(`\\b(?:${[...functions, ...labels].join('|')}|debugger)\\b`);
   const pattern = new RegExp(`^(?:${functions.join('|')})$`);
 
   return {


packages/strip/src/index.js
```diff
@@ -44,15 +44,15 @@ export default function strip(options = {}) {
 
   const labels = options.labels || [];
 
-  const firstpass = new RegExp(`\\b(?:${functions.join('|')}|debugger)\\b`);
+  const firstpass = new RegExp(`\\b(?:${[...functions, ...labels].join('|')}|debugger)\\b`);
```
shellscape: I'd combine functions and labels into something that could be checked on line 55, e.g.:

```suggestion
  const targets = [...functions, ...labels];
  const firstpass = new RegExp(`\\b(?:${targets.join('|')}|debugger)\\b`);
```

packages/strip/src/index.js
```diff
@@ -44,15 +44,15 @@ export default function strip(options = {}) {
...
   const pattern = new RegExp(`^(?:${functions.join('|')})$`);
 
   return {
     name: 'strip',
 
     transform(code, id) {
       if (!filter(id)) return null;
-      if (functions.length > 0 && !firstpass.test(code)) return null;
+      if (!firstpass.test(code)) return null;
```
shellscape: And then

```suggestion
      if (targets.length > 0 && !firstpass.test(code)) return null;
```

shellscape:
Closing as abandoned.

PR #181, opened 2020-01-25
core(typescript): Add note about old behaviour

NotWoods:

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

Sucrase has the same behaviour as the `ts.transpileModule` API we previously used, and is also 20x faster. Let's direct users to this plugin if they need the old behaviour.

Also adding myself to the contributors list :)

### DIFF
--- a/packages/typescript/package.json
+++ b/packages/typescript/package.json
@@ -8,6 +8,9 @@
   "license": "MIT",
   "repository": "rollup/plugins",
   "author": "Oskar Segersvärd",
+  "contributors": [
+    "Tiger Oakes <contact@tigeroakes.com> (https://tigeroakes.com)"
+  ],
   "homepage": "https://github.com/rollup/plugins/tree/master/packages/typescript/#readme",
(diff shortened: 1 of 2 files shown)


packages/typescript/package.json
```diff
@@ -8,6 +8,9 @@
   "license": "MIT",
   "repository": "rollup/plugins",
   "author": "Oskar Segersvärd",
+  "contributors": [
```
shellscape: `maintainers` would be better here. we track and credit contributors here: https://github.com/rollup/plugins/blob/07d0fbf99b7cedf106b89df7961c229510893a3c/LICENSE#L3

NotWoods: I'll just remove this for now and we can potentially add that field later. It seems that `maintainers` is set by NPM automatically so whatever we add could get overwritten.

Issue #183, opened 2020-01-27
Inlined SVG with Stencil in Chrom causes errors


rsmogura:
- Rollup Plugin Name: image
- Rollup Plugin Version: 2.0.1
- Rollup Version: ?
- Operating System (or Browser): Chrome
- Node Version: 12.13.1

### How Do We Reproduce?

Use instructions provided in Plugin `README.MD`. Inline SVG using import statements, and pass in TSX as image `src` (code samples below)

### Expected Behavior

SVG displayed in browser

### Actual Behavior

SVG not displayed in browser. Data URI generates error in Chrome only. Not visible in Edge, nor Safari.

Error doesn't appear if Data URI is encoded with Base 64 (see https://github.com/rollup/plugins/pull/136)

Problematic SVG Data URI:
```
data:image/svg+xml;utf-8,<?xml version="1.0" encoding="UTF-8"?><svg width="46px" height="43px" viewBox="0 0 46 43" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">    <!-- Generator: Sketch 62 (91390) - https://sketch.com -->    <title>Slice 1</title>    <desc>Created with Sketch.</desc>    <g id="Page-1" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">        <polyline id="Path" stroke="#979797" points="0.6875 0.78515625 28.890625 40.9804688 44.796875 0.78515625"></polyline>    </g></svg>
```

Error:
```
This page contains the following errors:
error on line 1 at column 426: AttValue: ' expected
Below is a rendering of the page up to the first error.
```

shellscape:
Thanks for reporting the issue. This is being tracked in #173.

PR #184, opened 2020-01-27
fix(image): On-the-fly wrap SVG into Base64 encoded data URI

rsmogura:

This PR contains:

- [x ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ x] yes (_bugfixes and features will not be merged without tests_)
- [ ] no

Breaking Changes?

- [ x] no

### Description

This change encodes the URI in Base64 on the fly
keeping original SVG as text; no size increase due to
Base64 & other conversions.

When option `preferBase64Uri` is used and image is SVG:
* the code checking if `btoa` exists is generated;
* original image is kept in helper var `src`;
* on-the-fly data conversion is generated.

Output code could be modeled as:
```
const src = "<svg></svg>";
const img = 'data:image/svg+xml;'+ (btoa ? 'base64,' + btoa(src) : 'utf-8,' + src);
```


shellscape:
Thank you for taking the time to open a PR. Unfortunately we can't accept it at this time. This PR introduces a new feature in a new option and we have not previously discussed that feature. As mentioned in #183, we are tracking the fix for this in #173.

PR #185, opened 2020-01-27
chore: Use typechecking with some plugins

NotWoods:
## Rollup Plugin Names: `data-uri, pluginutils, typescript`

This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

Now that the new Typescript plugin is released lets start using it for typechecking our plugins! 

I've worked to normalize the tsconfig settings we use across the different plugins. The es2017 target was chosen because its the highest that Node 8 supports. The `src` glob is used so that VSCode and other IDEs can tell that a tsconfig file affects the corresponding plugin.

estree-walker doesn't export the `Node` type so its imported from `@types/estree` instead. I use an inline import because ESLint complains about importing a module we don't depend on.

### DIFF
--- a/packages/data-uri/package.json
+++ b/packages/data-uri/package.json
@@ -48,7 +48,7 @@
     "rollup": "^1.20.0"
   },
   "devDependencies": {
-    "@rollup/plugin-typescript": "^2.1.0",
+    "@rollup/plugin-typescript": "^3.0.0",
     "@rollup/pluginutils": "^3.0.1",
     "rollup": "^1.27.14",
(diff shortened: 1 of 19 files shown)


shellscape:
So for these types of multi-plugin PRs, we need a separate commit for each plugin. Otherwise the publish script won't pick up on each individual change (because versions are based on commits). Can you break this out to one commit per plugin? And then we'll do a regular merge of this PR.

PR #191, opened 2020-02-01
Output compilerOption Diagnostics

samcooke98:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

Typescript Rollup didn't output compiler options diagnostics. The most obvious case where this fails is when the supplied tsconfig leaves `composite: true`. (as declaration is ignored by rollup plugin) 

Previously, the plugin would fail silently, and print no detail about why it failed.

```
./src/index.ts → ./dist/index.js...
[!] (typescript plugin) Error: Couldn't compile 
src/index.ts
Error: Couldn't compile src/index.ts
    at Object.transform (/node_modules/@rollup/plugin-typescript/dist/index.js:466:23)
    at /node_modules/rollup/dist/rollup.js:16611:25

```

By adding the `services.getCompilerOptionsDiagnostics()` call the plugin will emit reasons for the failure.


packages/typescript/test/fixtures/invalid-compiler-options/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "compilerOptions": {
+    "composite": true,
+    "declaration": false,
+  },
+  "include": ["./"]
+}
```
NotWoods: Missing newline at the end

packages/typescript/test/test.js
```diff
@@ -359,6 +359,29 @@ test('complies code that uses browser functions', async (t) => {
...
+        })
+      ],
+      onwarn({ toString, ...warning }) {
+        warnings.push(warning);
+      }
+    })
+  );
+
+  t.true(
+    err.message.includes('Composite projects may not disable declaration emit'),
```
NotWoods: Should we override this flag too?

packages/typescript/test/test.js
```diff
@@ -359,6 +359,29 @@ test('complies code that uses browser functions', async (t) => {
   t.true(code.includes('navigator.clipboard.readText()'), code);
 });
 
+test('reports compiler options errors', async (t) => {
+  process.chdir('fixtures/invalid-compiler-options');
+  const warnings = [];
```
NotWoods: Since the warnings aren't used they should be removed

packages/typescript/test/test.js
NotWoods: I think we should add `composite` to our list of overridden compiler options if it causes an error.

samcooke98: I agree - I think this is best done in a separate PR however. It seems ideal to output compilerOptions errors, regardless of the overridden compiler options - Future typescript versions may introduce new compilerOption errors. 

(And thinking about it, when I do this in a separate PR, it'll break this test)

shellscape:
@samcooke98 apologies that this one sat so long. Any chance you'd be willing to address the conflicts and failing CI?

samcooke98:
> @samcooke98 apologies that this one sat so long. Any chance you'd be willing to address the conflicts and failing CI?

No worries, it's my bad too, it slipped off my radar. I'll try to get to it this weekend (ie within the next 3 days)

shellscape:
@samcooke98 we had some CI shenanigans going that were fixed up today. Please do a merge from master and push and it should start passing.

samcooke98:
Just tested this with`@rollup/plugin-typescript@4.1.1` and couldn't reproduce - seems fixed? 

Will close now (Sorry I forgot about this PR)

PR #198, opened 2020-02-05
feat(multi-entry): add typescript typings

LukasBombach:
## Rollup Plugin Name: `multiEntry`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

The `multiEntry` plugin does not appear to support `TypeScript`. This PR should fix this. Please review and test me carefully, I'm not entirely sure I did everything right.

### DIFF
--- a/packages/multi-entry/types/index.d.ts
+++ b/packages/multi-entry/types/index.d.ts
@@ -1,0 +1,17 @@
+import { Plugin } from "rollup";
+
+export interface Options {
+  /**
+   * If `true`, instructs the plugin to export named exports to the bundle from all
+   * entries. If `false`, the plugin will not export any entry exports to the bundle.
+   * This can be useful when wanting to combine code from multiple entry files, but
+   * not necessarily to export each entry file's exports.
+   * @default true
+   */
+  exports?: boolean;
+}
+
+/**
+ * allow the use of multiple entry points for a bundle
+ */
+export default function multiEntry(options?: Options): Plugin;
--- a/packages/multi-entry/types/index.d.ts
+++ b/packages/multi-entry/types/index.d.ts
@@ -12,6 +12,6 @@ export interface Options {
 }
 
 /**
- * allow the use of multiple entry points for a bundle
+ * 🍣 A Rollup plugin which allows use of multiple entry points for a bundle.
  */
 export default function multiEntry(options?: Options): Plugin;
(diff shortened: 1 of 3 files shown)


packages/multi-entry/types/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+   * entries. If `false`, the plugin will not export any entry exports to the bundle.
+   * This can be useful when wanting to combine code from multiple entry files, but
+   * not necessarily to export each entry file's exports.
+   * @default true
+   */
+  exports?: boolean;
+}
+
+/**
+ * allow the use of multiple entry points for a bundle
```
NotWoods:
```suggestion
 * 🍣 A Rollup plugin which allows use of multiple entry points for a bundle.
```

packages/multi-entry/types/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export interface Options {
+  /**
+   * If `true`, instructs the plugin to export named exports to the bundle from all
+   * entries. If `false`, the plugin will not export any entry exports to the bundle.
+   * This can be useful when wanting to combine code from multiple entry files, but
+   * not necessarily to export each entry file's exports.
+   * @default true
+   */
+  exports?: boolean;
```
NotWoods: The options type can also be the `input` type: string, string array, or object with includes/excludes. See the [configure function](https://github.com/rollup/plugins/blob/master/packages/multi-entry/src/index.js#L12)

LukasBombach: Ah ok, that contradicts the readme, that should be updated too then

https://github.com/rollup/plugins/tree/master/packages/multi-entry#options

NotWoods: @shellscape Is this intended behaviour? If so I agree we should update the README

shellscape: The README isn't necessarily wrong, but it needs an update to be more in line with other plugins. Specifically around `include` and `exclude`

shellscape:
@LukasBombach ahoy there. since this is coming up on 60 days old, I wanted to check in to see if you were interested in continuing work on it or if we should close.

PR #212, opened 2020-02-15
fix(url): Don't append path separator on empty dirname

stevezhu:
## Rollup Plugin Name: `url`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] no

### Description

This is a fix for where the output of file paths with public path would be `publicPath//filename.ext` instead of `publicPath/filename.ext` for files in the root directory.

This is more consistent for concatenation `fileName`s with `publicPath` and relative paths in general (`/filename.ext` vs `filename.ext`).


shellscape:
Thanks for the PR. Because this is a breaking change, we're going to need to take some extra time to examine it. Please bear with us.

shellscape:
@pnevares @NotWoods @wuweiweiwu Please have a look at this

shellscape:
@NotWoods ping. this is one that needs a second set of eyes on it I think

Issue #211, opened 2020-02-14
'onwrite' deprecated?


farhanRIT:
Hi, I get this warning message when building my project:

(!) The "onwrite" hook used by plugin node-resolve is deprecated. The "generateBundle/writeBundle" hook should be used instead.

I am not really sure which node-resolve it uses.
But, based on my package: "rollup-plugin-node-resolve": "3.3.0"

- Rollup Plugin Name: 
- Rollup Plugin Version:
- Rollup Version:
- Operating System (or Browser):
- Node Version:

shellscape:
Thanks for opening an issue. Citing the issue template:

>   Issues without minimal reproductions will be closed! Please provide a link to one by:
>   1. Using the REPL at https://rollupjs.org/repl/, or
>   2. Using the REPL.it reproduction template at https://repl.it/@rollup/rollup-repro
>      (allows full use of all rollup options and plugins), or
>   3. Provide a minimal repository link (Read https://git.io/fNzHA for instructions).
>      These may take more time to triage than the other options.

Ignoring the entire issue template is not OK. You're also using a deprecated npm package. Follow the instructions in the deprecation message to get the current package.

farhanRIT:
> Thanks for opening an issue. Citing the issue template:
> 
> > Issues without minimal reproductions will be closed! Please provide a link to one by:
> > 
> > 1. Using the REPL at https://rollupjs.org/repl/, or
> > 2. Using the REPL.it reproduction template at https://repl.it/@rollup/rollup-repro
> >    (allows full use of all rollup options and plugins), or
> > 3. Provide a minimal repository link (Read https://git.io/fNzHA for instructions).
> >    These may take more time to triage than the other options.
> 
> Ignoring the entire issue template is not OK. You're also using a deprecated npm package. Follow the instructions in the deprecation message to get the current package.

What do I need to do? I have 0 idea. First time posting

shellscape:
https://www.npmjs.com/package/rollup-plugin-node-resolve

farhanRIT:
> https://www.npmjs.com/package/rollup-plugin-node-resolve

Awesome thanks. So that is where the original package came from. npm search gave me this and thought this were the location for the deprecated package. Updated to this new package, and warning message is gone and glad that I don't have to change anything on my end.

Issue #205, opened 2020-02-09
Optional Chaining Causes Build to Fail


bennypowers:
- Rollup Plugin Name: `@rollup/plugin-typescript`
- Rollup Plugin Version:  ^3.0.0
- Rollup Version: ^1.31.0
- Operating System (or Browser): Mac OS
- Node Version: v12.3.1

### How Do We Reproduce?

```
git clone git@github.com:bennypowers/rollup-plugin-typescript-optional-chaining-repro.git
cd rollup-plugin-typescript-optional-chaining-repro
npm i
npx rollup -c
```

### Expected Behavior
Compiles

### Actual Behavior
Fails with

```
index.ts → stdout...
[!] Error: Unexpected token (Note that you need plugins to import files that are not JavaScript)
index.ts (2:3)
1: const a = { b: { c: undefined } };
2: a?.b?.c;
      ^
Error: Unexpected token (Note that you need plugins to import files that are not JavaScript)
    at error (/Users/bennyp/Documents/rt-repro/node_modules/rollup/dist/shared/node-entry.js:5400:30)
    at Module.error (/Users/bennyp/Documents/rt-repro/node_modules/rollup/dist/shared/node-entry.js:9820:16)
    at tryParse (/Users/bennyp/Documents/rt-repro/node_modules/rollup/dist/shared/node-entry.js:9713:23)
    at Module.setSource (/Users/bennyp/Documents/rt-repro/node_modules/rollup/dist/shared/node-entry.js:10076:33)
    at /Users/bennyp/Documents/rt-repro/node_modules/rollup/dist/shared/node-entry.js:12362:20
    at async Promise.all (index 0)
    at async Promise.all (index 0)
    at async Promise.all (index 0)
```

I now present the relevant contents of said repro repo

## package.json
```json
{
  "devDependencies": {
    "@rollup/plugin-typescript": "^3.0.0",
    "rollup": "^1.31.0",
    "typescript": "^3.7.5"
  }
}
```
## rollup.config.js
```js
import typescript from '@rollup/plugin-typescript';

export default {
  input: 'index.ts',
  output: {
    format: 'es',
    outDir: '.'
  },
  plugins: [typescript({ typescript: require('typescript') })]
}
```
## index.ts
```ts
const a = { b: { c: undefined } };
a?.b?.c;
```

## tsconfig.json
```
{
  "include": ["index.ts"],
  "compilerOptions": {
    "experimentalDecorators": true,
    "target": "esnext",
[...]

Issue #204, opened 2020-02-09
Failing on compilation due to `TS2403: Subsequent variable declarations must have the same type.`


NickGlenn:
- Rollup Plugin Name: typescript
- Rollup Plugin Version: 3.0.0
- Rollup Version: 1.30.1, 1.27.14
- Operating System (or Browser): Windows, OSX
- Node Version: 8.10.0 (Win), 12.15.0 (Win), 12.15.0 (OSX)

### How Do We Reproduce?

[Repl.it Project Replicating the Issue](https://repl.it/repls/PeskyBigheartedForms)

I've seen this occur on a couple different libraries know, but the most notable one is `react-router-dom`. Set up a React project and install `react-router-dom` along with the types, `@types/react-router-dom`.

### Expected Behavior

Ideally it should compile unless there's actual user error. Using the exact same project but with the old [rollup-plugin-typescript](https://github.com/rollup/rollup-plugin-typescript) module works flawlessly.

### Actual Behavior

Build fails due to `TS2403: Subsequent variable declarations must have the same type.  Variable 'History' ...`.

```
src/index.tsx → dist/bundle.js...
[!] (plugin typescript) Error: @rollup/plugin-typescript TS2403: Subsequent variable declarations must have the same type.  Variable 'History' must be of type 'typeof import("/home/runner/PeskyBigheartedForms/node_modules/@types/history/index")', but here has type '{ new (): import("/home/runner/PeskyBigheartedForms/node_modules/@types/history/index"); prototype: import("/home/runner/PeskyBigheartedForms/node_modules/@types/history/index"); }'.
node_modules/typescript/lib/lib.dom.d.ts (9211:13)

9211 declare var History: {
                 ~~~~~~~

  node_modules/@types/history/index.d.ts:6:1
    6 export as namespace History;
      ~~~~~~
    'History' was also declared here.

Error: @rollup/plugin-typescript TS2403: Subsequent variable declarations must have thesame type.  Variable 'History' must be of type 'typeof import("/home/runner/PeskyBigheartedForms/node_modules/@types/history/index")', but here has type '{ new (): import("/home/runner/PeskyBigheartedForms/node_modules/@types/history/index"); prototype: import("/home/runner/PeskyBigheartedForms/node_modules/@types/history/index"); }'.
    at error (/home/runner/PeskyBigheartedForms/node_modules/rollup/dist/rollup.js:5365:30)
[...]

PR #217, opened 2020-02-20
feat(typescript): Move to BuilderProgram API

NotWoods:

This PR contains:

- [x] bugfix
- [x] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] no

List any relevant issue numbers: #216, #205, #204, #70, #61

### Description

This is another refactor to move from `LanguageService` (meant for IDEs) to `BuilderProgram` (meant for partial recomplation). This makes some big changes to the plugin to better align with what Typescript does: rather than Rollup giving us files to transform and the plugin transforming them on the fly, Typescript detects all source files and does a entire project transpilation at the very start. This is similar to Jake Archibald's simple-ts plugin, except the compilation is done in-memory rather than with a temp directory. `BuilderProgram` is also used by Webpack's ts-loader. Since the behavior is better aligned with Typescript, this fixes some odd inconsistencies in the typechecking like #204 and #211. 

We now support the `declaration` and `declarationMap` options with this change! When generating declaration files, the `outDir` must be set and be pointed to Rollup's output `dir` (or some subfolder). This allows us to use Rollup's asset emit API, and other plugins can then intercept these assets to potentially bundle the declaration files later. We currently emit each file separately, similar to rollup-plugin-typescript2.

This also adds experimental support for incremental builds, and paves the way for project references/build mode too.

### Breaking changes
All files discovered by Typescript will be compiled. The `include`/`exclude` filter can be used to filter out the files that Typescript finds. By default, all files in a directory with the tsconfig.json will be included, so most projects should not need additional changes. Our tests have been updated to have tsconfig.json files in each fixture directory, since we don't want all fixtures to be compiled for each test.

`tslib` can no longer be overriden by passing in source code directly. Now, it must be a path to a file. This better aligns with other plugins and is probably the expected behavior. Users can migrate by placing their custom tslib source code inside a file in their project, then including a path to it inside the plugin options.

The declaration, declarationMap, incremental, and tsBuildInfoFile options are no longer ignored. If these options are set, then a Rollput output `dir` must be used and Typescript's `outDir`/`declarationDir` must be located inside of it.Compilationn will fail if these directories are not set to be within `dir`.

We also fail compilation if the deprecated `out` option is used, or if the incompatible `outFile` is used.


paulrostorp:
Is this ready for review ?

NotWoods:
Yes

PR #222, opened 2020-02-21
feat(commonjs): export properties defined using Object.defineProperty(exports, ..)

edeustace:
## Rollup Plugin Name: `commonjs `

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

This pr adds exports that have been defined using the `Object.defineProperty(exports, ...)` technique. I was running into an issue with a 3rd party lib (`@material-ui/core@3.9.4`) where it was using this technique to define the exports. Rollup was falling over with the error `export 'foo' is not exported by 'path/to/module'`.

### DIFF
--- a/packages/commonjs/src/transform.js
+++ b/packages/commonjs/src/transform.js
@@ -67,6 +67,25 @@ export function checkEsModule(parse, code, id) {
   return { isEsModule, hasDefaultExport: false, ast };
 }
 
+function defineProperty(node, targetName) {
+  if (node.type !== 'CallExpression') return;
+
+  const {
+    callee: { object, property }
+  } = node;
+
+  if (!object || object.type !== 'Identifier' || object.name !== 'Object') return;
+
+  if (!property || property.type !== 'Identifier' || property.name !== 'defineProperty') return;
+
+  if (node.arguments.length !== 3) return;
+
+  const [target, val] = node.arguments;
+  if (target.type !== 'Identifier' || target.name !== targetName) return;
+  // eslint-disable-next-line consistent-return
+  return val.value;
+}
+
 export function transformCommonjs(
   parse,
   code,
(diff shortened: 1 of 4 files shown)


packages/commonjs/src/transform.js
```diff
@@ -299,6 +318,10 @@ export function transformCommonjs(
         return;
       }
 
+      // Is this a call to Object.defineProperty(exports, ...)?
+      const def = defineProperty(node, 'exports');
```
lukastaegert: I would rename the function to `isDefinePropertyCall`. Then the comment is also not needed. The return value could be `propertyName`, then I do not have to check the function definition what it means.

edeustace: done @lukastaegert :+1:

edeustace: actually on reflection I propose it be called `getDefinePropertyCallName`, the java programmer in me is complaining that an `isX` function should only return a boolean. See latest commit. That  said - happy to rollback if you prefer `isDefinePropertyCall`.

shellscape: Why not return a boolean?

edeustace: We could, but we’d also need to get the property name to add to namedExports?

edeustace: We could, but we’d also need to get the property name to add to namedExports?

lukastaegert: You are right, and you definitely got what I meant 😉

shellscape: Looks good. :shipit:

shellscape:
thanks!

PR #228, opened 2020-02-26
feat(alias): Move to Typescript

NotWoods:

This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

Rewrite the alias plugin in Typescript.

### DIFF
--- a/packages/alias/rollup.config.js
+++ b/packages/alias/rollup.config.js
@@ -1,10 +1,13 @@
+import typescript from '@rollup/plugin-typescript';
+
 import pkg from './package.json';
 
 export default {
-  input: 'src/index.js',
-  external: [...Object.keys(pkg.dependencies), 'path', 'fs', 'os'],
+  input: 'src/index.ts',
+  external: [...Object.keys(pkg.dependencies), 'os'],
   output: [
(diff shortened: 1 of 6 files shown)


packages/alias/rollup.config.js
```diff
@@ -1,10 +1,13 @@
+import typescript from '@rollup/plugin-typescript';
+
 import pkg from './package.json';
 
 export default {
-  input: 'src/index.js',
-  external: [...Object.keys(pkg.dependencies), 'path', 'fs', 'os'],
+  input: 'src/index.ts',
+  external: [...Object.keys(pkg.dependencies), 'os'],
```
TrySound: This allows to not modify externals at all
```
external: id => !id.startsWith('.') && !path.isAbsolute(id)
```

NotWoods: I'm happy to do this but we should be consistent, all other packages currently use a hardcoded list. We could also use [`builtinModules`](https://nodejs.org/api/modules.html#modules_module_builtinmodules).

PR #280, opened 2020-03-24
feat(typescript): Add CustomTransformers support

ghepesdoru:
## Rollup Plugin Name: `{typescript}`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description
This PR adds the ability to use TypeScript custom transformers in @rollup/plugin-typescript, similar to @rollup-plugin-typescript2.

The ability to use custom transformers will be a part of the TypeScript configuration file one day, but for the time being existing solutions:
- usage of ttypescript in order to allow for the compilerOptions.transformers configuration (seems not to work in my project)
- usage of @rollup-plugin-typescript2 (unable to resolve references when jumping from js->ts->js->ts in a PnPfied context)

haven't work wonders in my case.

In adition to my immediate needs, I also see the addition as being good for the plugin as it keeps rollup build environments free of unnecesarry dependencies (arguabely one should be able to achive most common use cases with the core plugins).

The PR has only additions and should not introduce any breaking changes. Users of ttypescript won't be affected by the native support of custom transformers as transformers are merged from all sources before program emit. The supported transformers are identical to TypeScript's expected format `CustomTransformers`.

Tests demonstrating a before, after and afterDeclarations transformer has been added.

Edit: I'we added 2 special transformer factories objects allowing transformers in need to request access to the current Program or TypeChecker instances. 

If you consider Program as being harmful as an export, I am more then happy to drop it from the PR as a supported transformer factory type; most 3rd party transformers out there are written against accessing the program instance.

In what TypeChecker is concerned, I consider it to be safe for export and has not means of touching the Program instance.

Documentation added illustrating the new types and configuration properties.


NotWoods:
I really appreciate all this work, but I don't think its a good time to merge it right now. I'm still experimenting with how the builder program is set up, and a future version of the plugin might not have a way to expose it. I'll keep this open for now and keep you updated.

ghepesdoru:
Thanks for the explanation @NotWoods. If it's ok with you I will try to maintain this PR in case it will ever be needed. For the time being I will still have to use it in my current project.

shellscape:
@NotWoods @ghepesdoru it's been a while since we took a look at this. has anything changed?

NotWoods:
I'm still planning to rewrite the typing engine again as we have various bugs around declaration support. I'd like to keep this open so there's a useful fork but not merge it yet.

shellscape:
@NotWoods I think we're approaching a point here where the value of adding support for this is outweighing the wait on the rewrite. what are your thoughts?

NotWoods (approved):
Good call :+1:

shellscape:
@ghepesdoru thank you for sticking with this!

TeoTN:
Well, I would've make use of that feature but I see it hasn't been released yet...

PR #281, opened 2020-03-24
Fix run plugin could fork before bundle generated

istarkov:
## Rollup Plugin Name: `run`

This PR contains:

- [x] bugfix

Are tests included?

Basically hard to create test on race condition, we have situation that 80% of our developers
has no issue at all as bundle has written to disk before fork, and other 20% sometimes can't start app at all

- [x] no 

Breaking Changes?

- [x] no

Description:

At the moment of fork inside run, there is possible that bundle file is not created at the moment. Or is in the process of writing. This cause fork to break without any errors.

PS:
also fixed editorconfig as its not in sync with eslint rules

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,6 +8,8 @@ indent_size = 2
 end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
+quote_type = single
+max_line_length = 100
 
 [*.md]
 insert_final_newline = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,8 +8,6 @@ indent_size = 2
 end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
-quote_type = single
-max_line_length = 100
 
 [*.md]
 insert_final_newline = true
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     }
   },
   "lint-staged": {
-    "*.js,*.ts": [
+    "*.{ts,js}": [
       "eslint --fix",
       "git add"
     ],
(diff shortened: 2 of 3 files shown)


.editorconfig
```diff
@@ -8,6 +8,8 @@ indent_size = 2
 end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
+quote_type = single
```
shellscape: please don't commit edits to this file.

TrySound:
We did not succeed in strong reproduction. Sometimes it works, sometimes it does not.

.editorconfig
TrySound: Looks like lint-staged is not configured properly. Committed code is not reformatted. This is the main problem why these changes are added.

TrySound:
@istarkov Change this
```
  "lint-staged": {
    "*.js,*.ts": [
```
to this in root package.json please
```
  "lint-staged": {
    "*.{ts,js}": [
```

shellscape:
thanks!

TrySound:
@shellscape Looks like this was a breaking change. writeBundle signature is changed in rollup 2.

shellscape:
tests passed using Rollup v1.x though.

TrySound:
How? Rollup 2 is everywhere
https://github.com/rollup/plugins/blob/master/packages/run/package.json#L41

shellscape:
I don't understand your question. If the `writeBundle` signature changed, then we need to update the code for v1 support of `writeBundle`, downgrade the peerDeps, publish a patch so that there isn't a breaking version, and then re-update for v2 and publish a major.

TrySound:
Didn't see "edit".
Here's old signature
https://github.com/rollup/rollup/blob/v1.32.1/src/rollup/types.d.ts#L376

shellscape:
OK got this fixed. 1.2.2 was published to revert the break for rollup v1 users. v2.0.0 was published for Rollup v2 users.

TrySound:
Thank you!

PR #292, opened 2020-03-31
fix(run): use correct hook for reading options, handle facade chunks

TrySound:
## Rollup Plugin Name: `run`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description
Ref https://github.com/rollup/plugins/pull/241#issuecomment-605697947

- take input from latest options when build is started
- check facade id instead of modules list

### DIFF
--- a/packages/run/src/index.ts
+++ b/packages/run/src/index.ts
@@ -16,7 +16,7 @@ export default function run(opts: RollupRunOptions = {}): Plugin {
   return {
     name: 'run',
 
-    options(options) {
+    buildStart(options) {
       let inputs = options.input!;
 
       if (typeof inputs === 'string') {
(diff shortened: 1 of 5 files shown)


shellscape:
@TrySound can you elaborate on what the PR does? It's not clear from the PR title. This is important for the changelog.

TrySound:
Here quotes of Lukas response from linked PR

The first case
> Also note that if a plugin just wants to read options, it should NOT use the options hook because other plugins can still mutate options at that point. If you just want to read, always use buildStart which receives the final options used by Rollup.

The second case
> The point is that when Rollup cannot match the export signature of an entry chunk because it needs to add additional exports, it will create a facade chunk that reexports exactly those exports that the original entry point exposed. A facade chunk will have its chunk.modules empty because it does not contain the code of any module, it just contains reexports.

PR #301, opened 2020-04-10
feat(node-resolve): Export defaults

### LINKED ISSUE #299: Export `defaults` from `node-resolve` plugin
- Rollup Plugin Name: `node-resolve`
- Rollup Plugin Version: 7.1.1

### Feature Use Case

I want to be able to add `.ts` and `.tsx` extensions when using TypeScript but I have to manually repeat default values which adds up unnecessary manual work. If the defaults were exported, I could reuse them easily.

This would function similarly to [`const { defaults } = require('jest-config')`](https://unpkg.com/browse/jest-config@25.3.0/build/Defaults.js).

I can create a PR if you want.

### Feature Proposal

Add `export` keyword to https://github.com/rollup/plugins/blob/cabe7108f5fa0f618eee07fc2c09392cb88e3de5/packages/node-resolve/src/index.js#L20

MichaelDeBoey:

This PR contains:
- [x] feature

Are tests included?

- [x] no

Breaking Changes?

- [ ] no

_Note from @shellscape: We're going to change this to a named export and will break. Needs a new major._

### Description

Closes #299

### DIFF
--- a/packages/node-resolve/src/cjs-wrapper.js
+++ b/packages/node-resolve/src/cjs-wrapper.js
@@ -1,0 +1,6 @@
+import nodeResolve, { DEFAULT_OPTIONS } from './index';
+
+const wrapper = nodeResolve;
+wrapper.DEFAULT_OPTIONS = DEFAULT_OPTIONS;
+
+export default wrapper;
--- a/packages/node-resolve/src/cjs-wrapper.js
+++ b/packages/node-resolve/src/cjs-wrapper.js
@@ -1,6 +1,0 @@
-import nodeResolve, { DEFAULT_OPTIONS } from './index';
-
-const wrapper = nodeResolve;
-wrapper.DEFAULT_OPTIONS = DEFAULT_OPTIONS;
-
-export default wrapper;
--- a/packages/node-resolve/src/index.js
+++ b/packages/node-resolve/src/index.js
@@ -17,7 +17,7 @@ import {
 const builtins = new Set(builtinList);
 const ES6_BROWSER_EMPTY = '\0node-resolve:empty.js';
 const nullFn = () => null;
-export const DEFAULT_OPTIONS = {
+export const defaults = {
   customResolveOptions: {},
   dedupe: [],
   // It's important that .mjs is listed before .js so that Rollup will interpret npm modules
@@ -26,8 +26,8 @@ export const DEFAULT_OPTIONS = {
   resolveOnly: []
 };
 
-export default function nodeResolve(opts = {}) {
-  const options = Object.assign({}, DEFAULT_OPTIONS, opts);
+export const nodeResolve = (opts = {}) => {
+  const options = Object.assign({}, defaults, opts);
   const { customResolveOptions, extensions, jail } = options;
   const warnings = [];
   const packageInfoCache = new Map();
@@ -250,4 +250,4 @@ export default function nodeResolve(opts = {}) {
       return idToPackageInfo.get(id);
     }
   };
-}
+};
--- a/packages/node-resolve/src/index.js
+++ b/packages/node-resolve/src/index.js
@@ -1,6 +1,8 @@
 /* eslint-disable no-param-reassign, no-shadow, no-undefined */
 import { dirname, normalize, resolve, sep } from 'path';
 
+import deepFreeze from 'deep-freeze';
+
 import builtinList from 'builtin-modules';
 import isModule from 'is-module';
 
@@ -17,14 +19,15 @@ import {
 const builtins = new Set(builtinList);
 const ES6_BROWSER_EMPTY = '\0node-resolve:empty.js';
 const nullFn = () => null;
-export const defaults = Object.freeze({
+const defaults = {
   customResolveOptions: {},
   dedupe: [],
   // It's important that .mjs is listed before .js so that Rollup will interpret npm modules
   // which deploy both ESM .mjs and CommonJS .js files as ESM.
   extensions: ['.mjs', '.js', '.json', '.node'],
   resolveOnly: []
-});
+};
+export const DEFAULTS = deepFreeze(defaults);
 
 export const nodeResolve = (opts = {}) => {
   const options = Object.assign({}, defaults, opts);
--- a/packages/node-resolve/src/index.js
+++ b/packages/node-resolve/src/index.js
@@ -17,7 +17,7 @@ import {
 const builtins = new Set(builtinList);
 const ES6_BROWSER_EMPTY = '\0node-resolve:empty.js';
 const nullFn = () => null;
-const defaults = {
+export const defaults = {
   customResolveOptions: {},
   dedupe: [],
   // It's important that .mjs is listed before .js so that Rollup will interpret npm modules
--- a/packages/node-resolve/src/index.js
+++ b/packages/node-resolve/src/index.js
@@ -1,9 +1,9 @@
 /* eslint-disable no-param-reassign, no-shadow, no-undefined */
 import { dirname, normalize, resolve, sep } from 'path';
 
-import deepFreeze from 'deep-freeze';
-
 import builtinList from 'builtin-modules';
+import deepFreeze from 'deep-freeze';
+import deepMerge from 'deepmerge';
 import isModule from 'is-module';
 
 import { isDirCached, isFileCached, readCachedFile } from './cache';
@@ -27,7 +27,7 @@ const defaults = {
   extensions: ['.mjs', '.js', '.json', '.node'],
   resolveOnly: []
 };
-export const DEFAULTS = deepFreeze(defaults);
+export const DEFAULTS = deepFreeze(deepMerge({}, defaults));
 
 export const nodeResolve = (opts = {}) => {
   const options = Object.assign({}, defaults, opts);
--- a/packages/node-resolve/src/index.js
+++ b/packages/node-resolve/src/index.js
@@ -17,7 +17,7 @@ import {
 const builtins = new Set(builtinList);
 const ES6_BROWSER_EMPTY = '\0node-resolve:empty.js';
 const nullFn = () => null;
-const DEFAULT_OPTIONS = {
+export const DEFAULT_OPTIONS = {
   customResolveOptions: {},
   dedupe: [],
   // It's important that .mjs is listed before .js so that Rollup will interpret npm modules
@@ -26,7 +26,7 @@ const DEFAULT_OPTIONS = {
   resolveOnly: []
 };
 
-const mod = function nodeResolve(opts = {}) {
+export default function nodeResolve(opts = {}) {
   const options = Object.assign({}, DEFAULT_OPTIONS, opts);
   const { customResolveOptions, extensions, jail } = options;
   const warnings = [];
@@ -250,6 +250,4 @@ const mod = function nodeResolve(opts = {}) {
       return idToPackageInfo.get(id);
     }
   };
-};
-mod.DEFAULT_OPTIONS = DEFAULT_OPTIONS;
-export default mod;
+}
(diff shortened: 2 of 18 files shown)


shellscape:
Hey thanks for the PR! Looks like that broke a few things. Give that a test run locally.

MichaelDeBoey:
Hmmm tests are failing because I'm doing an export default **and** a named export now.
But that shouldn't fail, right? 🤔

NotWoods:
We might need some wrapper for the common JS output.

Before:
```js
module.exports = nodeResolve(...)
```

After:
```js
exports.default = nodeResolve(...)
exports.defaults = defaults
```

Ideal:
```js
mod = nodeResolve(...)
mod.defaults = defaults
module.exports = mod
```

---

Also, since `export default` and `export defaults` are quite similar, I think `defaultOptions` or `DEFAULT_OPTIONS` would be better names.

MichaelDeBoey:
@NotWoods Wrapping the module still fails

shellscape:
@MichaelDeBoey please merge from master again. we fixed up some CI related things today.

MichaelDeBoey:
@shellscape Rebased, but CI is still failing tho 😕

shellscape:
Bummer. Next step is to run tests locally to triage from there: `cd packages/node-resolve && pnpm run test`

MichaelDeBoey:
@shellscape It's still the same error

```shell
 TypeError {
    message: 'nodeResolve is not a function',
  }
```

This is because the plugin now exports both a default _and_ a named export.
Wrapping it like @NotWoods suggested isn't working either.

So I don't think we can do this without a breaking change/major release. 🤔

NotWoods:
The wrapper should be extracted to a separate file, and used as the entry point for only the CJS build. The presence of an `export const` and `export default` statement is throwing off Rollup's CJS build.

MichaelDeBoey:
@NotWoods Could you point me in the right direction on how I could best fix this? 🤔

NotWoods:
This behavior might be better inside an output plugin, but here's something I whipped together:

```js
// cjs-wrapper.js
import nodeResolve, { DEFAULT_OPTIONS } from './index';

nodeResolve.DEFAULT_OPTIONS = DEFAULT_OPTIONS;

export default nodeResolve;
```

```
// rollup.config.js
import babel from 'rollup-plugin-babel';
import json from '@rollup/plugin-json';

const plugins = [
    json(),
    babel({
      presets: [
        [
          '@babel/preset-env',
          {
            targets: {
              node: 6
            }
          }
        ]
      ]
    })
];
const external = Object.keys(pkg.dependencies).concat(['fs', 'path', 'os', 'util']);

export default [
	{
		input: 'src/index.js',
		plugins,
		external,
		output: [
    		{ file: pkg.module, format: 'es' }
  		]
	},
	{
		input: 'src/cjs-wrapper.js',
		plugins,
		external,
		output: [
    		{ file: pkg.main, format: 'cjs' }
  		]
	},
];
```

MichaelDeBoey:
@shellscape @NotWoods All is green now 🙂

packages/node-resolve/src/index.js
```diff
@@ -17,7 +17,7 @@ import {
 const builtins = new Set(builtinList);
 const ES6_BROWSER_EMPTY = '\0node-resolve:empty.js';
 const nullFn = () => null;
-const defaults = {
+export const DEFAULT_OPTIONS = {
```
shellscape, line 20: Not down with this change. We might want to expose this as something else, but the name internally shouldn't change. Let's keep `defaults` here.

packages/node-resolve/src/cjs-wrapper.js
```diff
@@ -0,0 +1,6 @@
+import nodeResolve, { DEFAULT_OPTIONS } from './index';
```
shellscape, line 1:
```suggestion
import nodeResolve, { defaults } from './index';
```

packages/node-resolve/src/cjs-wrapper.js
```diff
@@ -0,0 +1,6 @@
...
+
+const wrapper = nodeResolve;
+wrapper.DEFAULT_OPTIONS = DEFAULT_OPTIONS;
```
shellscape, line 4:
```js
wrapper.DEFAULT_OPTIONS = deepFreeze(defaults);
```

This should export a deeply frozen copy of defaults. We don't want the risk of a consumer modifying the defaults.

packages/node-resolve/src/index.js
MichaelDeBoey: This will lead to having `defaults` in the es build and having `DEFAULT_OPTIONS` in cjs build, no? 🤔

shellscape: The wrapper should be the source of truth for both. Sorry, I thought that was the way this was going. I can check out your fork tomorrow and take a look at making this happen tomorrow.

MichaelDeBoey: That’s not what @NotWoods was saying.
He made me create a separate `cjs-wrapper`.
I can revert and create the wrapper back in `index.js` if you want

NotWoods: @shellscape We should pick a single direction for this so @MichaelDeBoey doesn't have to run in circles :)

Using the wrapper for ES modules breaks the usual import format of `import default, { others } from 'module';`. The wrapper is only meant for CommonJS because it mutates the default export instead of exporting two separate items. We can find some alternative solution if preferred, such as making a breaking change and only exporting named exports like the Terser plugin. (`import { nodeResolve, DEFAULT_OPTIONS } from 'module'`)

MichaelDeBoey: As suggested in #299, this was meant as a non-breaking change, but I can export both as named if you want

shellscape: Yeah I'm sorry for the pivot suggestion, but the wrapper started to feel wrong the more I looked at it. Let's do a new major, nothing wrong with that. And do two named exports. I really don't want to change the internal name of defaults, but I'm good to export it as something else - that much doesn't matter to me. I do think we should only export a deeply frozen copy of that though.

MichaelDeBoey: So I export both `nodeResolve` & `defaults` (freezed) as named exports? 🤔

MichaelDeBoey:
@shellscape @NotWoods All is green again now 🙂

MichaelDeBoey:
@shellscape @NotWoods All is green again now 🙂

packages/node-resolve/src/index.js
```diff
@@ -17,16 +17,16 @@ import {
 const builtins = new Set(builtinList);
 const ES6_BROWSER_EMPTY = '\0node-resolve:empty.js';
 const nullFn = () => null;
-const defaults = {
+export const defaults = Object.freeze({
```
shellscape, line 20: You're almost there. I mentioned that we needed to **deep freeze** the object exported. Here's why:

```js
> const o = Object.freeze({ batcave: [] })
undefined
> o.batcave.push('joker')
1
> o
{ batcave: [ 'joker' ] }
```

Just using Object.freeze isn't enough here. Users can still modify properties that are themselves objects.

We also don't want to export this directly here. `defaults` should stay internal and we should be able to mutate it as needed internally. The value _exported_ however, should be a deeply frozen copy of this. `Object.freeze` is an in-place method, meaning the return value is the same as the value passed. 

```js
> const a = {}
undefined
> const b = Object.freeze(a)
undefined
> a === b
true
```

So we can't `export` here and we cant `freeze` here either.

packages/node-resolve/src/index.js
```diff
@@ -250,4 +250,4 @@ export default function nodeResolve(opts = {}) {
       return idToPackageInfo.get(id);
     }
   };
-}
+};
```
shellscape, line 253: What you want to do is copy and freeze down here:

```suggestion
};

export const DEFAULTS = deepFreeze(deepMerge({}, defaults));
```

Using https://npm.runkit.com/deepmerge and https://www.npmjs.com/package/deep-freeze. We might also consider https://www.npmjs.com/package/fast-copy but it looks possibly unproven.

This may seem excessive, but believe me, you do not want users opening issues because an errant modification of defaults caused some obscure, hard to triage bug. These things in the bundler realm need to be locked down effectively.

packages/node-resolve/src/index.js
MichaelDeBoey: Can you elaborate why it's important to do a **deep freeze** please?

I can also just do `export const DEFAULTS = { ...defaults };` to become the same I think? 🤔

KubaJastrz: Object.freeze only freezes first-level keys.

This is allowed with shallow freeze, while it shouldn't be with deep freeze:

```
const a = Object.freeze({
  b: 42,
  c: {
    d: 'foo',
  },
});

a.c.d = 'bar';
```

packages/node-resolve/src/index.js
```diff
@@ -25,8 +27,9 @@ const defaults = {
   extensions: ['.mjs', '.js', '.json', '.node'],
   resolveOnly: []
 };
+export const DEFAULTS = deepFreeze(defaults);
```
shellscape, line 30: you missed something here. go back and read my comment again :)

packages/node-resolve/src/index.js
KubaJastrz: > We also don't want to export this directly here. defaults should stay internal and we should be able to mutate it as needed internally. The value exported however, should be a deeply frozen copy of this.

@shellscape, what is the point of mutating defaults internally if exported value will stay the same? That will just lead to more bugs.

packages/node-resolve/src/index.js
KubaJastrz: I don't see the point of deep-cloning here. Deep-freeze should be enough. Am I missing something?

KubaJastrz:
```js
const a = { x: { y: 1 } };
const b = deepFreeze(a);

b.x = 4;
console.log(a.x); // { y: 1 }
```

shellscape: https://github.com/rollup/plugins/pull/301#discussion_r409597038

> defaults should stay internal and we should be able to mutate it as needed internally

KubaJastrz: Deep-freezing exports doesn't disallow mutating defaults, does it?

shellscape: We want to export a clone of `defaults` that is deeply frozen. We don't want the internal `defaults` to be frozen.

packages/node-resolve/src/index.js
shellscape: > That will just lead to more bugs.

It won't. The exported defaults is merely a reference.

packages/node-resolve/src/index.js
MichaelDeBoey: Spreading the object will do what you want, right? 🤔

```suggestion
export const DEFAULTS = deepFreeze({ ...defaults });
```

shellscape: No.

```js
→ node
Welcome to Node.js v12.15.0.
> const a = { b: [0] }
undefined
> const b = { ...a }
undefined
> b
{ b: [ 0 ] }
> b.b.push(1)
2
> b.b
[ 0, 1 ]
> a.b
[ 0, 1 ]
```

You're not catching onto the differences between a shallow copy and a deep copy. In a shallow copy, any properties which contain objects are by-reference, meaning a change to them also affects the "parent." 

Please try these out in your local REPL. The REPL comes in handy to demo ideas before proposing them

shellscape:
@MichaelDeBoey I'll fix this conflict when I get the chance to circle back.

MichaelDeBoey:
@shellscape Rebase onto `upstream/master`

shellscape:
Thanks for all the hard work that went into this one.

MichaelDeBoey:
With pleasure! 🙂

PR #306, opened 2020-04-13
refactor(pluginutils): replace micromatch with more lightweight picomatch.

Patcher56:

This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

This just replaces `micromatch` dependency with the `picomatch` package. `picomatch` is part of `micromatch` and the `pluginutils`-plugin only uses the `matcher` function which is the same as using `picomatch` directly (See https://github.com/micromatch/micromatch/blob/master/index.js#L104).
This reduces dependency size.

Also added custom types until `picomatch` adds types on it's own. (See https://github.com/micromatch/picomatch/pull/53#issuecomment-612801229)


shellscape:
Thanks for suggesting this. The one big hangup for me right now is that I don't want plugins to be in the business of maintaining types for dependencies. That's ripe for technical debt. If the picomatch folks don't respond over the next few days, I would suggest adding those types to https://github.com/DefinitelyTyped/DefinitelyTyped/. If they get published there or in picomatch itself, then I'm all about merging this.

Patcher56:
Alright, I understand that. I'll add the types to DefinitelyTyped if they don't respond.

shellscape:
@Patcher56 We're still open to this change. Any word on the types situation?

Patcher56:
Sadly no comment on the picomatch pull request. I am very busy at work rn, but will add the types to DefinitelyTyped on weekend. After that we can we remove the type declarations from this PR.

Is this ok for you?

Patcher56:
I just replaced the picomatch.d.ts file with the newly created [@types/picomatch](https://www.npmjs.com/package/@types/picomatch) package.

PR #317, opened 2020-04-19
Update index.js : *.mjs files should also be included in the scripts list.

BitYoungjae:
Fix : *.mjs files to be included in script tag lists.

## Rollup Plugin Name: `html`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

not only *.js files but also *.mjs files should be included in the scripts list.

### DIFF
--- a/packages/html/lib/index.js
+++ b/packages/html/lib/index.js
@@ -7,7 +7,8 @@ const getFiles = (bundle) => {
   const result = {};
   for (const file of files) {
     const { fileName } = file;
-    const extension = extname(fileName).substring(1);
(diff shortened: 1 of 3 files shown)


shellscape:
Thanks for the PR. You'll need to add a test for this change before we can accept it.

BitYoungjae:
I have no idea how to use ava for the test..

BitYoungjae:
simple change

PR #325, opened 2020-04-22
Support for explicit module.require calls (Closes #310)

### LINKED ISSUE #310: `module.require` is not supported
- Rollup Plugin Version: up to 11.1.0
- Rollup Version: 1.27.14
- Operating System (or Browser): any
- Node Version: 12

### How Do We Reproduce?

Node has 2 ways inside a CommonJS module to require another CommonJS module : the classic `require()` that everyone uses, and the less-well-known `module.require()` (doc here : https://nodejs.org/api/modules.html#modules_module_require_id ).

Here is an REPL that reproduces the issue : https://repl.it/repls/SociableSimpleDictionaries

If you directly do `node cjs.js`, it works. But if you do `npx rollup -c && node output/bundle.js` it fails because of `module.require`.

(To be honest, I have no idea why anyone would want to use `module.require` in real life, but as it happens some well-known projects do use it, like lodash - https://github.com/lodash/lodash/blob/master/.internal/nodeTypes.js#L20 - and this is how I happened upon this bug, by trying to rollup lodash)

### Expected Behavior

Calls to `module.require`should be considered like calls to `require`.

### Actual Behavior

Calls to `module.require`are ignored by rollup-plugin-commonjs, so they fail because `module` is modified to an object that just has an `module.export` property.

danielgindi:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: #310 

### Description

Detect `module.require` calls as `require` calls, the rest continues as usual.


danielgindi:
Tests fail here due to latest commits on `master`, they pass 3 commits back. The snapshot seems to be different on different node versions (10, 12, 13).

shellscape:
Thanks for the PR. We'll have to figure out why the snapshots are different between versions before we can move forward. A merge from upstream/master is a good idea too, we just added Node 14 to the test matrix.

shellscape:
OK so I figured out why your tests were failing; basically, you're a bit behind from master and need up update the snapshots.

```console
$ cd packages/commonjs
$ pnpm run test -- --u
```

Should do the trick. If it doesn't, you'll have to add an upstream remote to your fork and merge upstream/master into the PR branch, and then update the snapshots. Tested that approach on #326 and it looks like the solution.

danielgindi:
I guess the issue is fixed on latest master (this was previously based on `ee5be6a`).
I rebased and pushed so it's clean now. Let's see if it passes.

danielgindi:
Well, still the same issue. It passes locally but not on the CI.

Diff:
```diff
  -   var x = /*#__PURE__*/Object.freeze(/*#__PURE__*/Object.assign(/*#__PURE__*/Object.create(null), answer$1, {␊
  +   var x = /*#__PURE__*/Object.freeze(/*#__PURE__*/Object.assign({␊
  +   	__proto__: null,␊
      	'default': answer$1,␊
      	__moduleExports: answer,␊
      	answer: answer_1␊
  -   }));␊
  +   }, answer$1));␊
```

This is the exact same issue as before. On one node version it generates the first snapshot, on another the second. But now it fails on all CI platforms so it's more "stable".

shellscape:
You're going to have to update snapshots. I ran `pnpm run test` on your branch and it failed locally. `pnpm run test -- --u` will update the snapshots (which is correct with your changes) to stop the failure.

danielgindi:
@shellscape What was missing was `pnpm i` after pulling the 2 latest commits from `master`.

packages/commonjs/src/transform.js
```diff
@@ -146,10 +146,28 @@ export function transformCommonjs(
...
     if (!node) return false;
     if (node.type !== 'CallExpression') return false;
-    if (node.callee.name !== 'require' || scope.contains('require')) return false;
-    // Weird case of require() without arguments
-    if (node.arguments.length === 0) return false;
-    return true;
+
+    const { callee } = node;
+
+    if (callee.type === 'Identifier' && callee.name === 'require') {
```
NotWoods, line 152: It would be nice to have comments here to indicate the `require()` form that is checked :smile: 

```js    
    if (callee.type === 'Identifier' && callee.name === 'require') {
        // module.require()
```

```js
    } else if (callee.type === 'MemberExpression') {
        // require()
```

shellscape: great suggestion. @danielgindi could you add those in real quick?

danielgindi: Done :-)

Issue #333, opened 2020-04-25
@rollup/plugin-wasm does not support Web Workers (`window` is undefined)


vedantroy:
- Rollup Plugin Name: `@rollup/plugin-wasm`
- Rollup Plugin Version: 3.0.0
- Rollup Version: 2.7.2
- Operating System (or Browser): Chrome
- Node Version: Not relevant, but here it is: 13.9.0

### How Do We Reproduce?

I cannot provide a minimal reproduction (not because I don't know how to reproduce the error), but because the issue requires running the code outputted by rollup in a Service Worker, which I don't know how to do in repl.it. However, I will point out the line of code where the error is occurring.

Fake minimal repro so this isn't automatically closed (sorry!): https://repl.it/@VedantRoy/rollup-plugin-repro

### Expected Behavior
No errors when running the code outputted by rollup in a Service Worker.

### Actual Behavior
Get "window" is not defined error.

The error occurs because `@rollup/plugin-wasm` places this function in the code outputted by rollup. However, this function uses `window.atob`, which breaks when the code is run in a service worker because service workers don't have a `window`.

A possible fix is to replace `window.atob` with just `atob`, or maybe `(window || self).atob`.

```javascript
function _loadWasmModule(sync, src, imports) {
  var buf = null;
  var isNode =
    typeof process !== "undefined" &&
    process.versions != null &&
    process.versions.node != null;
  if (isNode) {
    buf = Buffer.from(src, "base64");
  } else {
    // window is not defined in a service worker
    var raw = window.atob(src);
    var rawLength = raw.length;
    buf = new Uint8Array(new ArrayBuffer(rawLength));
    for (var i = 0; i < rawLength; i++) {
      buf[i] = raw.charCodeAt(i);
    }
  }

  if (imports && !sync) {
    return WebAssembly.instantiate(buf, imports);
  } else if (!imports && !sync) {
    return WebAssembly.compile(buf);
  } else {
    var mod = new WebAssembly.Module(buf);
    return imports ? new WebAssembly.Instance(mod, imports) : mod;
  }
}
```

shellscape:
We're closing this issue as stale as it's more than 30 days without activity, and without an associated Pull Request. Please feel free to continue discussion. We'll reopen this issue if anything actionable is posted.

vedantroy:
@shellscape I believe this issue was fixed with a PR https://github.com/rollup/plugins/pull/334 :smile: 

But, yeah, makes sense to close it.

PR #334, opened 2020-04-25
Make wasm plugin work with service workers

vedantroy:
The code generated by @rollup/plugin-wasm breaks when executed in a
service worker because window is undefined.

## Rollup Plugin Name: `wasm`

This PR contains:

- [X] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [] no

~~This is a draft PR because I'm not sure how to write a test that runs code inside a service worker. Once I figure that out, I will add a test and update the above checkbox.~~

Breaking Changes?

- [ ] no

List any relevant issue numbers: #333 

### Description

~~This PR adds a check for whether `window` is undefined.~~ This PR makes the loader function use `globalThis` instead of `window`. Prior to this, `@rollup/plugin-wasm` would break when the resulting generated code was run in a service worker, because it assumed `window` was defined.

### DIFF
--- a/packages/wasm/src/index.js
+++ b/packages/wasm/src/index.js
@@ -30,7 +30,9 @@ export default function wasm(options = {}) {
         if (isNode) {
           buf = Buffer.from(src, 'base64')
         } else {
-          var raw = window.atob(src)
+          var raw;
+          if (typeof window === 'object') raw = window.atob(src)
+          else raw = self.atob(src)
           var rawLength = raw.length
           buf = new Uint8Array(new ArrayBuffer(rawLength))
           for(var i = 0; i < rawLength; i++) {
--- a/packages/wasm/src/index.js
+++ b/packages/wasm/src/index.js
@@ -30,9 +30,7 @@ export default function wasm(options = {}) {
         if (isNode) {
           buf = Buffer.from(src, 'base64')
         } else {
-          var raw;
-          if (typeof window === 'object') raw = window.atob(src)
-          else raw = self.atob(src)
+          var raw = globalThis.atob(src)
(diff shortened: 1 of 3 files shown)


shellscape:
Thanks for taking this on. We'll leave this open and pending until you give the go-ahead to review.

vedantroy:
@shellscape I added a test and refactored test.js because I needed to pass the code generated by rollup into a `Worker` instead of `testBundle`.

Note: The test is a bit fragile because I have to fake the browser environment (`atob` does not exist on `globalThis` in node, so I polyfilled it, and I undefined `process` to force the wasm loader to go into the browser section). However, I think it's the most reasonable test that is possible.

shellscape:
Thanks for the PR!

@ColinEberhardt and @jamen would you two be willing to take a look at this one?

vedantroy:
@jamen The gap of support seems to show that WebAssembly is less supported than `globalThis` in all cases (except for Opera Mobile, which supports wasm but not `globalThis`). However, if this is an issue, I can add a check so it tries using `window` before `globalThis`.

jamen:
What I mean is the browser versions. WebAssembly came first and globalThis came later, so there are versions where WebAssembly exists but globalThis doesn't. I believe that is a breaking change, because the build would produce a more limited range of support than WebAssembly itself.

I don't think its an issue, but I think it should be in a major version. I haven't been very involved anymore though, so someone else can make that call. But also, a build concerned about this would be using something like `babel-preset-env` anyways.

shellscape:
Thanks everyone!

PR #335, opened 2020-04-25
feat(virtual): improve virtual entry input option support

Autarc:
## Rollup Plugin Name: `virtual`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] no

### Description

The `virtual` plugin allows [to specify an inline script as the `input` entry point of a bundle](https://github.com/rollup/plugins/blob/master/packages/virtual/README.md#using-the-plugin-for-bundle-input). For handling virtual files it keeps a mapping of related module ids. Since [the `options` hook](https://github.com/rollup/rollup/blob/master/docs/05-plugin-development.md#options) is executed initially to handle input option modifications this PR extends the plugin to update the values so they can be further used in the next steps. Furthermore the example in the `README` was updated to avoid confusion by referring the correct `input` field.

While the change alone might not seem very helpful it allows other plugins to better handle the case of a virtual `input` entry point. In another PR the `@rollup/plugin-run` will be updated based on that.

### DIFF
--- a/packages/virtual/README.md
+++ b/packages/virtual/README.md
@@ -73,9 +73,9 @@ export default {
   plugins: [
     virtual({
       entry: `
-        import batman from 'batcave';
-        console.log(batman);
-      `
+import batman from 'batcave';
+console.log(batman);
+`
     })
   ]
 };
--- a/packages/virtual/README.md
+++ b/packages/virtual/README.md
@@ -43,12 +43,12 @@ Create a `rollup.config.js` [configuration file](https://www.rollupjs.org/guide/
 import virtual from '@rollup/plugin-virtual';
 
 export default {
-  entry: 'src/entry.js',
+  input: 'src/input.js',
   // ...
   plugins: [
     virtual({
-      batman: `export default 'na na na na na'`,
-      'src/robin.js': `export default 'batmannnnn'`
+      batman: `export default 'na na na na na';`,
+      'src/robin.js': `export default 'batmannnnn';`
     })
   ]
 };
@@ -73,9 +73,9 @@ export default {
   plugins: [
     virtual({
       entry: `
-import batman from 'batcave';
-console.log(batman);
-`
+        import batman from 'batcave';
+        console.log(batman);
+      `
     })
   ]
 };
--- a/packages/virtual/src/index.js
+++ b/packages/virtual/src/index.js
@@ -4,6 +4,7 @@ import path from 'path';
 const PREFIX = `\0virtual:`;
 
 export default function virtual(modules) {
+  const virtualInputIds = [];
   const resolvedIds = new Map();
 
   Object.keys(modules).forEach((id) => {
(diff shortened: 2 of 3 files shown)


packages/virtual/README.md
```diff
@@ -73,9 +73,9 @@ export default {
   plugins: [
     virtual({
       entry: `
-import batman from 'batcave';
-console.log(batman);
-`
+        import batman from 'batcave';
```
shellscape, line 76: because we're using backticks this will add quite a bit more padding to the source of the virtual module. with this being in the README I'd like to avoid that. please do revert this change.

packages/virtual/test/test.js
```diff
@@ -25,3 +25,35 @@ test('loads an absolute path from memory', (t) => {
...
+
+  const mappedInputOption = {
+    input: {
+      'foo.js': 'lib/foo.js',
+      'src/foo.js': 'lib/index.js'
+    }
+  };
+
+  t.is(plugin.options(singleInputOption).input, '\0virtual:foo.js');
+  t.deepEqual(plugin.options(multipleInputOption).input, [
```
shellscape, line 51: would prefer using snapshots here instead of deepEqual

packages/virtual/src/index.js
```diff
@@ -13,7 +14,49 @@ export default function virtual(modules) {
   return {
     name: 'virtual',
 
+    options(options) {
+      let updatedInput = options.input;
```
shellscape, line 18: @lukastaegert please give this a look. I recall something recently about this hook not being ideal but I'm having a hard time recalling what it was.

packages/virtual/README.md
Autarc: Just thought the example in the `README` looks a bit less aligned compared to the rest. Spacing was added for better readability compared to [the current version](https://github.com/rollup/plugins/blob/master/packages/virtual/README.md#using-the-plugin-for-bundle-input):

Screenshot 2020-04-26 at 01 39 02

But sure it shouldn't have a negative impact so if thats not preferred I can revert it.

shellscape: yes please do revert it

Autarc: Okay changed it back. Hope the other modification of using the `input` field in the first example rather than `entry` was okay 😅

packages/virtual/src/index.js
```diff
@@ -4,6 +4,7 @@ import path from 'path';
 const PREFIX = `\0virtual:`;
```
lukastaegert, line 4: I see we are doing all this because virtual ids do have this prefix. I know it has been there before but I think it should generally only be a last resort change to update `input`. The reason is that in the future, `input` might be extended in unexpected ways, which means that this plugin needs to play catch-up and will become more brittle.

Moreover I think the  issue is that IMO we should NOT add this prefix. Adding a `\0` prefix signifies to other plugins that they should skip this plugin e.g. in the `transform` hook. This means, however, that neither e.g. `rollup-plugin-typescript` nor `rollup-plugin-babel` will touch these modules, which I do not think is what users will expect. The only issue I foresee at the moment is that another plugin might catch the `load` hook, but I am not sure which plugin that could be. To prevent this, users should be instructed that this plugin should ideally be placed before other plugins that load modules.

What I would recommend instead (and yes, this will be a breaking change, but worthwhile IMO):
- Do not update `input` in the options hook.
- Change the `resolveId` hook like this:
  ```js
  resolveId(id, importer) {
      if (modules.hasOwnProperty(id)) {
        return id;
      }

      if (importer) {
        const resolved = path.resolve(path.dirname(importer), id);
        if (resolvedIds.has(resolved)) return resolved;
      }
  }
  ```
- Change the load hook like this:
  ```js
  load(id) {
      if (modules.hasOwnProperty(id)) {
        return modules[id];
      }
      if (resolvedIds.has(id)) {
        return resolvedIds.get(id);
      }
  }
  ```

Opinions?

Another approach might be to use a prefix without a leading `\0` but as I said, I am not 100% happy about modifying `input`.

shellscape: This is actually then insight I was looking for. WRT prefixes - won't we need something to differentiate virtual modules from static/file-based modules to prevent collisions? Or should that be left up to the user?

lukastaegert: The only situation I could think of is if a plugin implements a custom load hook. But most that I am aware of are not problematic as they just load themselves plugin-specific virtual files that have their own prefix. The prefixes have always been there (https://github.com/rollup/rollup-plugin-virtual/commit/6d7f6d08dd2803e9d103d26c13561080521f70d0#diff-1fdf421c05c1140f6d71444ea2b27638R3) and I guess they have not been added to solve a particular problem but I could be wrong. Or do you remember something I miss here @Rich-Harris (as it was your initial commit)?

shellscape:
@Autarc last call before close this as abandoned.

shellscape:
Closing as abandoned.

PR #336, opened 2020-04-25
 feat(run): use resolved input if file exists

Autarc:
## NOTE
This PR is is based on #335 as changes there are connected with this use case. It will be rebased as there are updates and also [refer then in the test to the `@rollup/plugin-virtual` module rather than the relative path.

The modified files are:
- [packages/run/rollup.config.js](https://github.com/Autarc/plugins/blob/7fb9cf1ed96f546beae6caa1613db59c40316b5c/packages/run/rollup.config.js)
- [packages/run/src/index.ts](https://github.com/Autarc/plugins/blob/7fb9cf1ed96f546beae6caa1613db59c40316b5c/packages/run/src/index.ts)
- [packages/run/test/test.js](https://github.com/Autarc/plugins/blob/7fb9cf1ed96f546beae6caa1613db59c40316b5c/packages/run/test/test.js) (will be updated with a new version of `@rollup/plugin-virtual`)
- [packages/run/package.json](https://github.com/Autarc/plugins/blob/7fb9cf1ed96f546beae6caa1613db59c40316b5c/packages/run/package.json) (will be updated with a new version of `@rollup/plugin-virtual`)
- [pnpm-lock.yaml](https://github.com/Autarc/plugins/blob/7fb9cf1ed96f546beae6caa1613db59c40316b5c/pnpm-lock.yaml) (will be updated with a new version of `@rollup/plugin-virtual`)

## Rollup Plugin Name: `run`

This PR contains:

- [x] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

The `run` plugin allows to run a bundle directly after its output is written to the disk. For figuring out the correct entry filename it uses the `input` entry point. As [the `virtual` plugin allows to specify inline scripts](https://github.com/rollup/plugins/blob/master/packages/virtual/README.md#using-the-plugin-for-bundle-input) instead of real paths [the matching is currently failing](https://github.com/rollup/plugins/blob/master/packages/run/src/index.ts#L47). To solve the issue its initially checked if the `input` entry is pointing to an existing file, otherwise the value is taken as is.

Along with the update of the `virtual` plugin in #335 this allows to successfully run the output of bundles initiated through virtual files.

### DIFF
--- a/packages/run/package.json
+++ b/packages/run/package.json
@@ -50,6 +50,7 @@
   },
   "devDependencies": {
     "@rollup/plugin-typescript": "^3.0.0",
+    "@rollup/plugin-virtual": "^2.0.1",
     "del": "^5.1.0",
     "rollup": "^2.0.0",
     "sinon": "8.0.4"
--- a/packages/run/rollup.config.js
+++ b/packages/run/rollup.config.js
@@ -5,7 +5,7 @@ import pkg from './package.json';
 export default {
   input: 'src/index.ts',
   plugins: [typescript()],
-  external: Object.keys(pkg.dependencies).concat(['path', 'child_process']),
+  external: Object.keys(pkg.dependencies).concat(['path', 'child_process', 'fs']),
   output: [
     { format: 'cjs', file: pkg.main },
     { format: 'esm', file: pkg.module }
--- a/packages/run/src/index.ts
+++ b/packages/run/src/index.ts
@@ -1,5 +1,6 @@
 import { ChildProcess, fork } from 'child_process';
 import * as path from 'path';
+import * as fs from 'fs';
 
 import { Plugin, RenderedChunk } from 'rollup';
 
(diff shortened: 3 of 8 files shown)


shellscape:
Thanks for the PR, but I'm a bit confused. You're referencing an unknown plugin named "build," but we don't have any plugins named that here. Did you mean the `run` plugin?

It looks like this is dependent on #335, so we'll put a hold label on this until that one is resolved.

Autarc:
@shellscape Yeah sorry for the confusion. As you said the `run` plugin was meant so I've updated it in the header. It clearly depends on #335 so it is certainly better on hold until we know how to proceeded there. Perhaps the change in this PR might afterwards be even redundant. Just though pushing it for feedback might be an idea since I already had it locally working.

shellscape:
Totally, thanks for clarifying. We'll tackle this once #335 is resolved.

shellscape:
Closing as abandoned as it depends on #335 which was closed as abandoned.

Issue #209, opened 2020-02-13
resolveOnly option prevent local file resolution


arantes555:
- Rollup Plugin Name: node-resolve
- Rollup Plugin Version: 7.1.0 & 7.1.1
- Rollup Version: 1.31.0
- Operating System (or Browser): macOS
- Node Version: v13.7.0

### How Do We Reproduce?

```bash
git clone https://github.com/arantes555/node-resolve-local-files.git
cd node-resolve-local-files
npm install
node build.js
```

### Expected Behavior

Using node-resolve 7.1.x with the `resolveOnly` option should prevent resolving other node modules, but should not prevent resolving local files. 

### Actual Behavior

Using node-resolve 7.1.x with the `resolveOnly` option prevents resolving local files.

shellscape:
Thanks for reporting the issue. We'd be happy to review a PR that resolves this.

shellscape:
Closed by #337

PR #337, opened 2020-04-26
fix(node-resolve): Resolve local files if `resolveOption` is set

JaviIG:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

This PR fixes the issue #209, where using the `resolveOnly` option at the `node-resolve` plugin prevented the resolution of local modules.

### DIFF
--- a/packages/node-resolve/src/index.js
+++ b/packages/node-resolve/src/index.js
@@ -111,6 +111,7 @@ export const nodeResolve = (opts = {}) => {
 
       const parts = importee.split(/[/\\]/);
       let id = parts.shift();
+      let isRelativeImport = false;
 
       if (id[0] === '@' && parts.length > 0) {
         // scoped packages
@@ -118,11 +119,15 @@ export const nodeResolve = (opts = {}) => {
       } else if (id[0] === '.') {
         // an import relative to the parent dir of the importer
         id = resolve(basedir, importee);
+        isRelativeImport = true;
       }
 
       const input = normalizeInput(rollupOptions.input);
-
-      if (resolveOnly.length && !resolveOnly.some((pattern) => pattern.test(id))) {
+      if (
+        !isRelativeImport &&
+        resolveOnly.length &&
+        !resolveOnly.some((pattern) => pattern.test(id))
+      ) {
         if (input.includes(id)) {
           return null;
         }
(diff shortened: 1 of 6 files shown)


shellscape:
Thanks for the PR!

I might be missing something here, entirely possible, but if this change were successful, wouldn't the deepEqual fail for the `only` test?

```
  t.deepEqual(imports, ['@scoped/foo', '@scoped/bar']);
```

JaviIG:
Hello @shellscape. I'm the one probably missing something here. For what I understood about the issue #209 is that the `resolveOnly` option shouldn't prevent resolving local modules. 
So in the `only.js` tests, as the `only-local.js` fixture is a local module, it should be resolved always, and never imported. If that assumption is right, shouldn't the `node-resolve` plugin configured either with the `test` string, or the `/^@scoped\/.*$/` regex, never import the `only-local.js` file?

shellscape:
Yeah I think I mistook the differences between imports and resolved.

I would like to try and get some test in place that has something we can check that asserts that the local file did resolve. It looks like the lack of warnings is the signal that it's successful but that feels brittle. Nothing to block this PR on but if you can think of anything that'd be cool.

I'd like @lukastaegert to put some quick eyes on this, but it looks good to me so far.

JaviIG:
Great! I don´t have a deep understanding of the Rollup internal types and properties, so I might be wrong, but after a fast check, I thing we can use the ´PreRenderedChunk.modules´ property, and test if the modules that should be resolved are contained in that dictionary keys. If it works as I think, this might give the tests a little more confidence.

Other option I can think of is searching inside the generated code, but it is far more ugly and fragile than searching inside the imported or resolved modules.

If you have any more ideas I would be more than happy to try them.

shellscape:
@JaviIG would you like to try Lukas' test suggestions before we merge?

JaviIG:
Thank you both! I´ve just uploaded the improved tests. That was the idea I had, but I wasn´t 100% sure about the behavior of the `modules` property. If there is anything more you think I can improve just tell me :)

shellscape:
thanks!

PR #308, opened 2020-04-14
feat(html): support custom meta element creation

mseeley:
This allows for convenient creation of `meta` elements without reimplementing all template function responsibilities.

## Rollup Plugin Name: `html`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] yes (_bugfixes and features will not be merged without tests_)
- [ ] no

Breaking Changes?

- [ ] yes (_breaking changes will not be merged unless absolutely necessary_)
- [x] no

If yes, then include "BREAKING CHANGES:" in the first commit message body, followed by a description of what is breaking. 

List any relevant issue numbers: n/a

### Description

The `html` plugin doesn't allow for injection of `meta` elements without re-implementing all of the template function responsibilities. This is bunch of relative complexity to take on in order to, in my case, inject our old friendly viewport meta below:

```
<meta
  name="viewport"
  content="minimum-scale=1, initial-scale=1, width=device-width"
/>
```

Instead I added support for a backwards compatible `meta` option.
- when omitted the current `utf-8` charset meta will be inserted.
- when provided the charset meta must be supplied manually. To me this was the simplest implementation and an easy to doc/grok all or nothing approach.

~~Running tests locally some snapshots are failing with `null` vs `undefined` values for `map`. I'm running 12.15.0 using nvm. Perhaps there's a behavior changes between `12` and `12.xx`. Tests passed for changed code, so, here we go. :tada:~~

PR #339, opened 2020-04-27
feat(html): Allow adding arbitrary HTML to body and head

yagebu:
## Rollup Plugin Name: `{html}`

This PR contains:

- [ ] bugfix
- [x] feature
- [x] refactor
- [x] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: n/a

### Description

I recently started using this plugin and discovered the just merged #308, which allows adding additional `<meta>` tags to the HTML `<head>`. This will nearly allow me to get rid of a custom template function for adding a `viewport` meta tag. However, I'd still need a custom template function to add a `<link>` for a favicon or a link to a webmanifest.

This PR adds two more options to add arbitrary strings at the end of the head and body tag respectively in the template function. This might be considered bloat of the options of this plugin, but I'd consider them generic and useful enough to be worth the addition.

Also, the first three commits contain a minor refactor to get rid of an eslint-disable and fix up some minor mistakes in the README.

### DIFF
--- a/packages/html/README.md
+++ b/packages/html/README.md
@@ -92,7 +92,7 @@ const template = ({ attributes, bundle, files, publicPath, title }) => { ... }
 
 - `attributes`: Corresponds to the `attributes` option passed to the plugin
 - `bundle`: An `Object` containing key-value pairs of [`AssetInfo` or `ChunkInfo`](https://rollupjs.org/guide/en/#generatebundle)
-- `files`: An `Array` of `AssetInfo` or `ChunkInfo` containing any entry (`isEntry: true`) files, and any asset (`isAsset: true`) files in the bundle that will be emitted
+- `files`: An `Object` mapping file types to `Array`s of `AssetInfo` or `ChunkInfo` of entry (`isEntry: true`) files, and asset (`isAsset: true`) files in the bundle that will be emitted
 - `publicPath`: Corresponds to the `publicPath` option passed to the plugin
 - `title`: Corresponds to the `title` option passed to the plugin
 
@@ -134,7 +134,7 @@ Consumes an object with key-value pairs that represent an HTML element attribute
 const { makeHtmlAttributes } = require('@rollup/plugin-html');
 
 makeHtmlAttributes({ lang: 'en', 'data-batcave': 'secret' });
-// -> 'lang="en" data-batcave="secret"'
+// -> ' lang="en" data-batcave="secret"'
 ```
 
 ## Supported Output Formats
--- a/packages/html/README.md
+++ b/packages/html/README.md
@@ -57,12 +57,26 @@ Specifies additional attributes for `html`, `link`, and `script` elements. For e
 
 _Note: If using the `es` / `esm` output format, `{ type: 'module'}` is automatically added to `attributes.script`._
 
+### `body`
+
+Type: `String`<br>
+Default: `''`
+
+Specifies any extra HTML to insert into the `<body>` tag.
+
 ### `fileName`
 
 Type: `String`<br>
 Default: `'index.html'`
 
 Specifies the name of the HTML to emit.
+
+### `head`
+
+Type: `String`<br>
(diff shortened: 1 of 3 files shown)


yagebu:
I didn't find out from the READMEs how to update the snapshots, so there's one failing snapshot where I test the new two options. How can I update the snapshots?

shellscape:
Thanks for opening a PR and for your interest in the plugin. I had feared that #308 would be a slippery slope and potentially open us up to people wanting to inject all manner of things into the default document. I was hoping that wouldn't happen, but here we are 😄 

I apologize for dismissing this so quickly, but we have to pass on this addition. We'll be directing folks proposing similar features in the future to use the template function. That exists for customization of the default template, and folks should really use that. If repetition over multiple domains is a concern, I'd recommend creating a small package that can be reused across domains that exports the same template function. 

Thanks again for your time on this.

yagebu:
As I wrote, I understand that might seem to lead to bloat, but I think you're a bit quick to dismiss it completely. I'd argue for reverting #308 and just adding the `head` option. It covers the "I need additional <meta> tags" and all my use cases (various types of links). Also, for simply adding HTML tags, I think the object syntax of the `meta` option is more verbose and harder to understand than a simple string.

@mseeley: What do you think? Would your use cases also be addressed by a `head` option that allows adding a string into the head tag in the template?

shellscape:
Thanks for the reply. I disagree that it was closed too quickly, however. I had already prepared a response should a similar feature be proposed. I'm afraid this isn't one I'm willing to budge on.

yagebu:
Thanks for the reply.

>  I'm afraid this isn't one I'm willing to budge on.

I'm sad to hear that. I think the meta->head change wouldn't hurt users of the meta option and a net improvement overall.

Thanks for maintaining these plugins!

vperron:
I'm also very interested in something close to this. I'm trying to use rollup as a new build environment for SPAs and this HTML plugin proves quite frustrating to customize :)

A patch enabling the customization of `meta` tags a few days ago was most welcome and enabled me to have a viewport-ready HTML. There was just no way to insert those, unless entirely redefining the `template` function in my own code and all of its internal code to transform "files" into "links" and "scripts".

My new issue is to add something as basic as a favicon to my HTML. Any feature that would let me add links, or any tags to the head would be so useful. I think that is a sad and hard limitation. 

I see maybe two way outs there:
- either add "links" and "scripts", array-based configuration entries, just like the "meta" one that has just been added;
- change the `template` function signature to include directly "links"  and "scripts" instead of the "bundle" and "files" entries. This way we could just add tags on the fly and call the original function to do the job properly.

What do you think ? I could contribute either one of those.

shellscape:
Neither of those options are very appealing to me. The default template (https://github.com/rollup/plugins/blob/233c7c19801d055963dcdd258a6508038781c311/packages/html/lib/index.js#L27-L61) isn't very complicated, nor are the methods for placing `files`, `links`, and the rest. Our html plugin follows the minimalistic nature of `mini-html-webpack-plugin`, and I spoke with the author at length about the ideas behind offering a different html plugin for webpack.

This boils down to a difference in ideology: "make every possible allowance for the user" vs "provide a simple baseline and empower the user to meet their needs". You'll notice this a lot when comparing webpack and its components to rollup and its components. We don't want to get into the business of trying to meet every single user need with specific features for each one of those needs. Instead, we try to provide a reasonable default baseline, and make sure we have the right features in place to let users extend that to their heart's content.

If the `html` plugin isn't doing it for you, there are a few more listed here: https://github.com/rollup/awesome that might prove to meet your wants a little better. (We maintain that list to lift other plugin and utility authors up, even without them being official plugins).

That said, there won't be any more extensions to the `template` function option made in the near future. If exporting any of the plugin's internal methods would aid you in building your custom template, please open a separate issue and we can talk about making that happen.

PR #340, opened 2020-04-27
refactor(html): remove eslint disable

yagebu:
## Rollup Plugin Name: `{html}`

This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [x] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

List any relevant issue numbers: #339 

### Description

Extracted fixes from #339:

[...] a minor refactor to get rid of an eslint-disable and fix up some minor mistakes in the README.

### DIFF
--- a/packages/html/README.md
+++ b/packages/html/README.md
@@ -92,7 +92,7 @@ const template = ({ attributes, bundle, files, publicPath, title }) => { ... }
 
 - `attributes`: Corresponds to the `attributes` option passed to the plugin
 - `bundle`: An `Object` containing key-value pairs of [`AssetInfo` or `ChunkInfo`](https://rollupjs.org/guide/en/#generatebundle)
-- `files`: An `Array` of `AssetInfo` or `ChunkInfo` containing any entry (`isEntry: true`) files, and any asset (`isAsset: true`) files in the bundle that will be emitted
+- `files`: An `Object` mapping file types to `Array`s of `AssetInfo` or `ChunkInfo` of entry (`isEntry: true`) files, and asset (`isAsset: true`) files in the bundle that will be emitted
 - `publicPath`: Corresponds to the `publicPath` option passed to the plugin
 - `title`: Corresponds to the `title` option passed to the plugin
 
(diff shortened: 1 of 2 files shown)


shellscape:
Thanks for the PR

I don't think the README changes are an improvement, and contain a mistake if their own. And I'm not a fan of style change refactors without benefit. This one is a :-1: for me but we'll leave this open for a while to let other maintainers comment.

yagebu:
> I don't think the README changes are an improvement, and contain a mistake if their own.

Could you point out the mistake? Some of the changes are really fixes though...

shellscape:
> Specifies the name of the HTML to emit.

I believe you left out "file" in that sentence.

This is overall a subjective style change. What you're perceiving as mistakes are just another way to say the same thing. In the future I'd recommend opening a documentation issue and discussing there first.

yagebu:
> I believe you left out "file" in that sentence.

I copied the existing line...

image

That last line being in the wrong place doesn't seem to be a subjective style change to me.

yagebu:
> The documentation changes seem unneeded.

I had assumed they would be obvious as just fixes (with minimal style changes), but that assumption was apparently incorrect so I'll elaborate:

- The first commit moves an incorrectly placed line (Specfies the name[...]) so that it's placed with the actual option it describes.
- The `file` property of the object passed to the template function is currently described as `An Array of [...]`, however it's an Object. I'm not perfectly happy with the new wording either, so if you have any better suggestions, please change it to something better. But I find it odd to call an Object an "Array" and dismiss a fix to change that as a "subjective style change".
- The last change shows the literal output of the `makeHtmlAttributes` function. That output starts with a space, so that should be reflected in the displayed output in the README.

shellscape:
> I had assumed they would be obvious as just fixes (with minimal style changes), but that assumption was apparently incorrect so I'll elaborate

I _really_ don't appreciate the tone of the language chosen here. This can remain open and discussed by users and other maintainers but this isn't a PR that I'm going to approve, and will be stepping away from it.

yagebu:
> I really don't appreciate the tone of the language chosen here.

@shellscape: I'm sorry about my wording, it was not my intention to offend. I was trying to express that I was wrong to not provide more detailed descriptions of the changes in this PR.

yagebu:
These changes are overall of minor value, so I'll close this PR. Sorry to have caused such a fuss.

Thanks for your comments and the maintenance of these plugins!

Issue #322, opened 2020-04-21
Memory leak in typescript plugin (4.0.0+)


Tronil:
- Rollup Plugin Name: typescript
- Rollup Plugin Version: 4.1.1
- Rollup Version: 2.6.1
- Operating System (or Browser): Linux
- Node Version: 12.16.2

### How Do We Reproduce?
Clone https://github.com/Tronil/rollup-ts-mem-leak
Run:
npm install
npm run watch
Repeatedly trigger watch build by saving src/index.ts

### Expected Behavior
Memory usage should not grow until rollup crashes with out-of-memory.

### Actual Behavior
Memory usage grows with about 25MB for every triggered build, eventually running out of memory and crashing.

Note that this repo only contains a "hello world" file - for real projects this memory leak is massive. In our project the leak was around 300 MB for every build, making the process crash after 6 or 7 builds. The 3.x versions do not appear to have this issue, so it is likely related to the 4.0 rewrite.

shellscape:
Have you verifies that it's a problem with that plugin specifically and not rollup core?

Tronil:
If I use version 3.1.1 of the typescript plugin, the memory leak seems to disappear - so I assume the leak is in the plugin. Also the leak did not appear before adding the typescript plugin (our project was originally in pure javascript).

Tronil:
Looked through the source and this part caught my eye:
```
buildStart() {
      emitParsedOptionsErrors(ts, this, parsedOptions);

      program = createWatchProgram(ts, this, {
        formatHost,
        resolveModule,
        parsedOptions,
        writeFile(fileName, data) {
          emittedFiles.set(fileName, data);
        }
      });
    },
```
`program` is being replaced every time buildStart() is called, which looks a little odd. After I put in a check for whether it is already set, the leak disappears...
So this seems to fix it:
```
buildStart() {
      emitParsedOptionsErrors(ts, this, parsedOptions);
      if (!program) {
        program = createWatchProgram(ts, this, {
          formatHost,
          resolveModule,
          parsedOptions,
          writeFile(fileName, data) {
            emittedFiles.set(fileName, data);
          }
        });
      }
    },
```

NotWoods:
Would you be interested in submitting a PR with the change? Otherwise I can add it in later.

Tronil:
I don't know when I'll have some time to look at a PR, so I wouldn't wait for me to create one.

VitorLuizC:
I can make this PR.

PR #352, opened 2020-04-29
fix(typescript): fix a memory leak (#322)

VitorLuizC:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

rollup/plugins#322

### Description

Just check if program already exists before creating a new one.

### DIFF
--- a/packages/typescript/src/index.ts
+++ b/packages/typescript/src/index.ts
@@ -34,14 +34,17 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
     buildStart() {
       emitParsedOptionsErrors(ts, this, parsedOptions);
 
-      program = createWatchProgram(ts, this, {
-        formatHost,
-        resolveModule,
-        parsedOptions,
-        writeFile(fileName, data) {
-          emittedFiles.set(fileName, data);
-        }
-      });
+      // Fixes a memory leak https://github.com/rollup/plugins/issues/322
+      if (!program) {
+        program = createWatchProgram(ts, this, {
+          formatHost,
+          resolveModule,
+          parsedOptions,
+          writeFile(fileName, data) {
+            emittedFiles.set(fileName, data);
+          }
+        });
+      }
     },
 
     buildEnd() {
--- a/packages/data-uri/test/test.js
+++ b/packages/data-uri/test/test.js
@@ -5,7 +5,8 @@ import { rollup } from 'rollup';
 
 import { testBundle } from '../../../util/test';
 
-import dataUri from '..';
+// eslint-disable-next-line import/no-unresolved, import/extensions
+import dataUri from '../';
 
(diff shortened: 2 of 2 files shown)


packages/typescript/src/index.ts
```diff
@@ -34,14 +34,17 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
...
-        formatHost,
-        resolveModule,
-        parsedOptions,
-        writeFile(fileName, data) {
-          emittedFiles.set(fileName, data);
-        }
-      });
+      // Fixes a memory leak https://github.com/rollup/plugins/issues/322
+      if (!program) {
+        program = createWatchProgram(ts, this, {
```
shellscape, line 39: @NotWoods how can we best add a test for this to protect against regressions?

NotWoods: Unfortunately I'm not sure how to test for memory leaks short of re-running the test many times (which would have a large impact on our CI times). I think we should merge this in for now without an added test.

shellscape: right on.

shellscape:
ping @NotWoods think this is good to go?

shellscape:
thanks!

finico:
@VitorLuizC @shellscape @NotWoods This change fixed the memory leaks, but led to a race between typescript's and rollup's watchers.
Now when saving the file, sometimes the bundle builds by rollup before typescript compiles the the file.
It's because independent watchers are used to track files and they don't wait for each other.

shellscape:
@finico could you open that as a new issue with a reproduction?

VitorLuizC:
@shellscape @NotWoods

I can make a new PR that closes current program and then open a new one, and it'll fix both issues.

```js
    buildStart() {
      emitParsedOptionsErrors(ts, this, parsedOptions);

      if (program) {
        program.close();
      }

      program = createWatchProgram(ts, this, {
          formatHost,
          resolveModule,
          parsedOptions,
          writeFile(fileName, data) {
            emittedFiles.set(fileName, data);
          }
        });
    }
```

finico:
@VitorLuizC It has no sense to use WatchProgram that way - unnecessary overhead (about 700ms on my machine) over creation a new instance

shellscape:
'Gents. It's best to open this on a _new issue_ for tracking. A reproduction is also essential because tests can be created from one. We let this PR in without tests due to it's nature, but I'm going to insist that any further PRs that modify how `program` is used be accompanied by a test. That is also something I would rather see discussed on a new issue. This PR is merged and closed, please move to a new issue.

milky2028:
I'm seeing the race condition issue mentioned by @finico. I created an issue #418.

PR #361, opened 2020-05-01
feat(node-resolve): Add default export

NotWoods:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no *(this is a pretty straightforward change)*

Breaking Changes?

- [x] yes (_breaking changes will not be merged unless absolutely necessary_) Since we're including breaking changes anyways, I'm going to fix the Typescript interface name so it matches other plugins.
- [ ] no

### Description

Building on #301, this updates the documentation but also re-adds a default export. When the plugin is used with ES modules, the previous change is no longer a breaking change. However, CommonJS modules still must switch to named exports.

For consistency we should always used named exports in the documentation, but this helps users migrate to the new format.

### DIFF
--- a/packages/node-resolve/src/index.js
+++ b/packages/node-resolve/src/index.js
@@ -29,7 +29,7 @@ const defaults = {
 };
 export const DEFAULTS = deepFreeze(deepMerge({}, defaults));
 
-export const nodeResolve = (opts = {}) => {
+export function nodeResolve(opts = {}) {
   const options = Object.assign({}, defaults, opts);
   const { customResolveOptions, extensions, jail } = options;
   const warnings = [];
@@ -258,4 +258,6 @@ export const nodeResolve = (opts = {}) => {
       return idToPackageInfo.get(id);
     }
   };
-};
+}
+
+export default nodeResolve;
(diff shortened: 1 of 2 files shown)


TrySound:
Do you mean "Add named export"?

NotWoods:
We've switched to named exports already here, this adds back an export named "default".

packages/alias/README.md
```diff
@@ -140,9 +140,9 @@ Example:
 ```javascript
 // rollup.config.js
 import alias from '@rollup/plugin-alias';
-import resolve from '@rollup/plugin-node-resolve';
+import { nodeResolve } from '@rollup/plugin-node-resolve';
```
shellscape, line 143: let's keep this scoped to node-resolve for now, and update the plugins individually. we'll want to do a patch release for each so the docs on npm match the repo. for that we need individual commits for each plugin to keep the changelogs and publishing automation correct.

shellscape: @NotWoods ping. can we do one PR per plugin or a separate commit per plugin? With the path forward in #360 I'm thinking that a separate PR might be easier to digest.

NotWoods: I'll revert these doc changes since the path forward should allow us to keep the same import format.

shellscape: Awesome

PR #363, opened 2020-05-01
feat(wasm): Switch to TypeScript & named exports

NotWoods:
BREAKING CHANGES: Named exports are used for CJS

## Rollup Plugin Name: `wasm`

This PR contains:

- [ ] bugfix
- [x] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [ ] no

List any relevant issue numbers: #360

### Description

Switch to Typescript for the wasm plugin. I chose to also switch to named exports here, but that could be pulled out to a separate PR.

### DIFF
--- a/packages/wasm/rollup.config.js
+++ b/packages/wasm/rollup.config.js
@@ -1,13 +1,9 @@
-const { builtinModules } = require('module');
+import typescript from '@rollup/plugin-typescript';
+import { createConfig } from '../../shared/rollup.config';
 
-const pkg = require('./package.json');
+import pkg from './package.json';
 
-const dependencies = Object.keys(pkg.dependencies || {});
-
-export default [
-  {
-    input: 'src/index.js',
-    output: { exports: 'named', file: 'dist/index.js', format: 'cjs' },
-    external: [...builtinModules, ...dependencies]
-  }
-];
+export default {
+  ...createConfig(pkg),
+  plugins: [typescript()],
+}
--- a/packages/wasm/types/index.d.ts
+++ b/packages/wasm/types/index.d.ts
@@ -1,0 +1,14 @@
+import { Plugin } from 'rollup';
+
+export interface RollupWasmOptions {
+  /**
+   * Specifies an array of strings that each represent a WebAssembly file to load synchronously.
+   */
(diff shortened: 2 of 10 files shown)


packages/wasm/rollup.config.js
```diff
@@ -1,13 +1,15 @@
...
 export default [
   {
-    input: 'src/index.js',
-    output: { exports: 'named', file: 'dist/index.js', format: 'cjs' },
-    external: [...builtinModules, ...dependencies]
+    input: 'src/index.ts',
+    plugins: [typescript()],
+    external: ['fs', 'path'],
+    output: [
+      { format: 'cjs', file: pkg.main, exports: 'named' },
```
shellscape, line 11: Let's follow the path forward that @lukastaegert laid out here: https://github.com/rollup/plugins/issues/360#issuecomment-622997038

This is also an opportunity to create a "base" rollup config for plugins. Should help reduce duplication across the repo.

NotWoods: Should we place a base config in root or next to the test utils?

shellscape: `/shared/rollup.confg.js` would probably work at the root.

packages/wasm/types/index.d.ts
```diff
@@ -0,0 +1,14 @@
+import { Plugin } from 'rollup';
+
+export interface RollupWasmOptions {
```
shellscape, line 3: Do we need this with the latest version of the typescript plugin, or will that do definition file output for us now?

NotWoods: At the moment the types folder is a bit cleaner. I'm still refining the declaration output for the typescript plugin.

shellscape:
Gotta fix that pnpm lock and we should be good to merge

shellscape:
@NotWoods your commit message on merge didn't include the breaking changes, so the publish script won't pick up the breaking changes. If we update and force push that's going to cause trouble with pending PRs. Please make sure you're adding `BREAKING CHANGES: ...` to the merge commit messages before merging PRs that have them.

Issue #239, opened 2020-03-04
Plugin sucrase is not resolving directories


Vehmloewff:
- Rollup Plugin Name:  sucrase
- Rollup Plugin Version: 3.0.0
- Rollup Version: 1.32.0
- Operating System (or Browser): Ubuntu 19.10
- Node Version: 12.16.1

### How Do We Reproduce?

1. See the [repl](https://repl.it/@Vehmloewff/rollup-plugin-sucrase-no-resolve)
2. Click "run"
3. Yikes!

### Expected Behavior

It is expected that sucrase will resolve `./src` to `./src/index.ts` because `./src` is a directory.

### Actual Behavior

```
input.ts → output/bundle.js...
[!] Error: Could not resolve './src' from input.ts
Error: Could not resolve './src' from input.ts
    at error (/home/runner/rollup-plugin-sucrase-no-resolve/node_modules/rollup/dist/rollup.js:5365:30)
    at ModuleLoader.handleMissingImports (/home/runner/rollup-plugin-sucrase-no-resolve/node_modules/rollup/dist/rollup.js:12249:17)
    at ModuleLoader.<anonymous> (/home/runner/rollup-plugin-sucrase-no-resolve/node_modules/rollup/dist/rollup.js:12140:30)
    at Generator.next (<anonymous>)
    at fulfilled (/home/runner/rollup-plugin-sucrase-no-resolve/node_modules/rollup/dist/rollup.js:41:28)

exit status 1
```

shellscape:
If you use surcase outside of Rollup, does it resolve the directories as expected?

Vehmloewff:
Yes

```shell
▶ npx sucrase ./src -d ./dist --transforms typescript,imports         
src/index.ts -> dist/index.js
src/interfaces.ts -> dist/interfaces.js
src/lib/tracker.ts -> dist/lib/tracker.js
```

The resolve part for this plugin is done entirely 'in-house'.

https://github.com/rollup/plugins/blob/master/packages/sucrase/src/index.js#L14-L22

larrybotha:
@Vehmloewff this should have been resolved by #390

shellscape:
Just published this morning. I'll close this one.

PR #387, opened 2020-05-12
fix(sucrase): fix aliases not resolving

larrybotha:
## Rollup Plugin Name: `sucrase`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

- https://github.com/rollup/plugins/issues/239
	- I've written the fixtures so that both directories and filenames are resolved, which appears to resolve this issue

### Description

The plugin was only resolving imported files that started with `.`. 

When using an alias without an extension (TypeScript requires that no extension is appended):

- `@rollup/plugin-alias` resolves the full path to the file without the extension
- `@rollup/plugin-sucrase`'s resolver ignores the imported path because it doesn't begin with a `.`
- `sucrase` throws as it's unable to resolve an extensionless filename

This fix ensures the resolver processes file ids that have been processed by `@rollup/plugin-alias`.

~Possible caveats:~

- ~the regex I've used only checks for file ids beginning with `.` and `/`~
  - ~I'm not sure how this will behave on Windows, as I'm unfamiliar with how paths are resolved in Node on Windows~


larrybotha:
This PR doesn't actually resolve #239

larrybotha:
@shellscape I've resolved the conflicts on this PR

shellscape:
thanks!

PR #419, opened 2020-05-26
feat(pluginutils): add native node es modules support

TrySound:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

- https://github.com/rollup/plugins/issues/412
- https://github.com/rollup/plugins/pull/413

### Description

Enable native node es modules via "exports" field in package.json.
Added custom plugin to generate nested package.json with `{"type": "module"}`
as an alternative to mjs extension usage. This works similar to rollup distribution.
https://unpkg.com/browse/rollup@2.10.9/dist/es/

This change is necessary before migrating all other plugins.

### DIFF
--- a/packages/pluginutils/package.json
+++ b/packages/pluginutils/package.json
@@ -12,7 +12,7 @@
   "bugs": {
     "url": "https://github.com/rollup/plugins/issues"
   },
-  "main": "dist/index.js",
+  "main": "./dist/cjs/index.js",
   "engines": {
     "node": ">= 8.0.0"
   },
@@ -75,12 +75,17 @@
       "!**/types.ts"
     ]
   },
-  "module": "dist/index.es.js",
+  "exports": {
+    "require": "./dist/cjs/index.js",
+    "import": "./dist/es/index.js"
+  },
+  "module": "./dist/es/index.js",
   "nyc": {
     "extension": [
       ".js",
       ".ts"
     ]
   },
+  "type": "commonjs",
   "types": "types/index.d.ts"
 }
--- a/packages/pluginutils/rollup.config.js
+++ b/packages/pluginutils/rollup.config.js
@@ -2,20 +2,9 @@ import commonjs from '@rollup/plugin-commonjs';
 import resolve from '@rollup/plugin-node-resolve';
 import typescript from '@rollup/plugin-typescript';
 
-import pkg from './package.json';
+import { emitModulePackageFile } from '../../shared/rollup.config';
 
-function emitModulePackageFile() {
-  return {
-    name: 'emit-module-package-file',
-    generateBundle() {
-      this.emitFile({
-        type: 'asset',
-        fileName: 'package.json',
-        source: `{"type":"module"}`
-      });
-    }
-  };
-}
+import pkg from './package.json';
 
 export default {
   input: 'src/index.ts',
--- a/packages/pluginutils/rollup.config.js
+++ b/packages/pluginutils/rollup.config.js
@@ -17,7 +17,7 @@ export default {
   ],
   external: [...builtinModules, 'picomatch'],
   output: [
-    { format: 'cjs', file: pkg.main, exports: 'named' },
+    { file: pkg.main, format: 'cjs', exports: 'named' },
     { file: pkg.module, format: 'es', plugins: [emitModulePackageFile()] }
   ]
 };
--- a/packages/pluginutils/rollup.config.js
+++ b/packages/pluginutils/rollup.config.js
@@ -1,3 +1,5 @@
+import { builtinModules } from 'module';
+
 import commonjs from '@rollup/plugin-commonjs';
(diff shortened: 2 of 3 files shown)


packages/pluginutils/rollup.config.js
```diff
@@ -4,6 +4,19 @@ import typescript from '@rollup/plugin-typescript';
 
 import pkg from './package.json';
 
+function emitModulePackageFile() {
```
shellscape, line 7: let's export this output plugin from https://github.com/rollup/plugins/blob/master/shared/rollup.config.js so we can reuse it

TrySound:
Looks like there is a blocker
https://github.com/Rich-Harris/estree-walker/pull/23

packages/pluginutils/rollup.config.js
```diff
@@ -13,14 +15,7 @@ export default {
...
-      format: 'cjs',
-      file: pkg.main,
-      exports: 'named'
-    },
-    {
-      format: 'es',
-      file: pkg.module
-    }
+    { format: 'cjs', file: pkg.main, exports: 'named' },
+    { file: pkg.module, format: 'es', plugins: [emitModulePackageFile()] }
```
abdonrd, line 19:
```suggestion
    { format: 'cjs', file: pkg.main, exports: 'named' },
    { format: 'es', file: pkg.module, plugins: [emitModulePackageFile()] }
```

To have both outputs the same format order.

abdonrd:
```suggestion
    { file: pkg.main,  format: 'cjs', exports: 'named' },
    { file: pkg.module, format: 'es', plugins: [emitModulePackageFile()] }
```

If you prefer the other order.

abdonrd: Also I recommend the same in this PR: https://github.com/rollup/plugins/pull/413

TrySound:
@shellscape Bundling estree-walker would eliminate the problem. What do you think?

shellscape:
@TrySound it's got a pretty small bundle footprint of 2.9kb raw. That should be fine.

packages/pluginutils/package.json
```diff
@@ -75,12 +75,17 @@
       "!**/types.ts"
     ]
   },
-  "module": "dist/index.es.js",
+  "exports": {
+    "require": "./dist/cjs/index.js",
+    "import": "./dist/es/index.js"
+  },
+  "module": "./dist/es/index.js",
```
shellscape, line 82: question about the choice to nest these within a directory: our prior file format was `index.[format].[extension]` for output files. e.g. `index.es.js`. what's the reasoning behind the choice to put them into directories?

TrySound: es/package.json with {type:module}

TrySound: As an alternative to .mjs extension.

shellscape: OK. Can you explain in the PR description why we need mjs alternative? I've added a bunch more people to review. We're going to need to get majority buy-in to move forward with this since it's a pretty big paradigm shift.

danielgindi:
@TrySound can you please explain why we need to shift from `mjs`?

TrySound:
I just don't like it. And to be consistent with rollup. So no strong opinion.

shellscape:
> And to be consistent with Rollup

That seems the most compelling reason. I really don't like how it litters `dist` but if that stops the complaining about Node 14+ imports then I guess that's what we do.

If we're going to adopt this, we should also revisit comments by @Andarist where `dist` files were named `pluginutils.js` instead of `index.js`, as that would also follow parent Rollup's convention.

packages/pluginutils/package.json
```diff
@@ -75,12 +75,17 @@
...
   "nyc": {
     "extension": [
       ".js",
       ".ts"
     ]
   },
+  "type": "commonjs",
```
guybedford, line 89: This will treat `dist/es/index.js` as CommonJS - are you sure that is the intention here?

If not it is probably advisable to change `dist/es/index` to use the `.mjs` extension, or switch things around to the `.cjs` extension.

Nested package.json files can form sub-boundaries within packages, but its not necessarily recommended.

TrySound: Nested package.json is what I do here
https://github.com/rollup/plugins/pull/419/files#diff-bdf5352c3e711371ce12e70c4f962f4dR27-R38

guybedford: Ah right, didn't see that!

shellscape:
I'll be following up with a PR to address the comment from above:

> If we're going to adopt this, we should also revisit comments by @Andarist where dist files were named pluginutils.js instead of index.js, as that would also follow parent Rollup's convention.

PR #425, opened 2020-05-28
fix(typescript): sync rollup and typescript file watch

### LINKED ISSUE #418: @rollup/plugin-typescript breaks rollup watch
- Rollup Plugin Version: 4.1.2
- Rollup Version: 2.10.9
- Operating System (or Browser): macOS 10.15.4 and debian 10
- Node Version: seen in most recent versions of both node 10 and node 12

### How Do We Reproduce?

I've created a [minimal repo](https://github.com/milky2028/rollup-bug). 

```
npm install
```
Then,
```
npm run watch
```
to watch for changes. Make some changes to either of the files in the src directory, and you'll see that build files do not appropriately update. I was originally going to file a bug with the main rollup repo, but after testing with rollup-plugin-terser and @rollup/plugin-babel, I did not see this behavior with either of those plugins.

### Expected Behavior
I expect that build files in the dist folder of the included minimal repo should actively respond to changes and each built file should reflect the changes in source files.

### Actual Behavior
Build files do not appropriately respond and seem to be about one change behind.

dionysiusmarquis:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [ ] no

- #418 

### Description

This will fix #418. As discussed in #352 there is only one Typescript watcher program running. This program is out of sync with the `Rollup` hooks. I utillized the watcher status callback to listen to the diagnostic status codes. With this status it's possible to know whenever Typescript detects file changes or finished compiling. 

Sometimes `Rollup` triggers the watch hook while the Typescript watcher didn't recognize any changes (Typescript seems to be right in that cases). Because of that, I needed to race the Typescript watcher start promise against a timeout to always fulfill the start promise, even when `Rollup` gave "false alarm". I testet it quiet heavily. As far as I can tell every change found its way to the bundle and I didn't run into unfulfilled promises.

It seems pretty performant. Another approach would be to stop uing the Typescript watcher and instead creating a new compiler program everytime something changes. Then we wouldn't need any additional promise logic.

BREAKING CHANGES:
Added async logic


packages/typescript/src/index.ts
```diff
@@ -42,11 +46,40 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
...
+      // Race watcher start promise against a timeout in case Typescript and Rollup change detection is not in sync
+      tsWatcherStartPromise = Promise.race<Promise<void | boolean>>([
+        new Promise(resolve => setTimeout(resolve, 1000, true)),
+        new Promise(resolve => tsWatcherStartResolve = resolve)
+      ]);
+      tsWatcherFinishPromise = new Promise(resolve => tsWatcherFinishResolve = resolve);
+    },
+
     buildEnd() {
       if (process.env.ROLLUP_WATCH !== 'true') {
```
dionysiusmarquis, line 84: Btw: This doesn't work if you start 'rollup.watch' process.env.ROLLUP_WATCH will be undefined

packages/typescript/src/index.ts
```diff
@@ -42,11 +46,40 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
           parsedOptions,
           writeFile(fileName, data) {
             emittedFiles.set(fileName, data);
+          },
+          status(diagnostic) {
+            // Fullfil watcher promises by Typescript diagnostic message codes
+            if (diagnostic.category === ts.DiagnosticCategory.Message) {
+              switch(diagnostic.code) {
+                case 6032:
```
NotWoods, line 54: Constants should be used for these codes.

dionysiusmarquis: Typescript didn't include constants for the codes. I will add the relevant.

packages/typescript/src/watchProgram.ts
```diff
@@ -3,20 +3,113 @@ import { PluginContext } from 'rollup';
 import { buildDiagnosticReporter } from './diagnostics/emit';
 import { DiagnosticsHost } from './diagnostics/host';
 import { Resolver } from './moduleResolution';
+import { DiagnosticCategory } from 'typescript';
```
dionysiusmarquis, line 6: This doesn't work:
```js 
import('typescript'). DiagnosticCategory
```
`import('typescript')` doesn't have the property `DiagnosticCategory`

`Property 'DiagnosticCategory' does not exist on type 'Promise<{ default: typeof ts; …`

packages/typescript/src/index.ts
dionysiusmarquis: https://github.com/rollup/rollup/pull/3616 will fix it.

Should be `if (this.meta.watchMode !== true)` then

packages/typescript/src/watchProgram.ts
```diff
@@ -3,20 +3,113 @@ import { PluginContext } from 'rollup';
...
+}
+
+/**
+ * Typescript watch program helper to sync Typescript watch status with Rollup hooks.
+ */
+export class WatchProgramHelper {
+  private _startDeferred: Deferred | null = null;
+  private _finishDeferred: Deferred | null = null;
+
+  watch (timeout: number = 1000) {
```
NotWoods, line 60: Please make sure to run prettier to fix the styling. (We should make sure the CI is running it later.) There shouldn't be spaces between function names and parenthesis.

PR #449, opened 2020-06-10
fix(typescript): utilize 'this.meta.watchMode'

dionysiusmarquis:

This PR contains:

- [x] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [ ] no

BREAKING CHANGES: Changed `rollup` peer dependency to `rollup@^2.14.0`

https://github.com/rollup/rollup/issues/3608
#418
Fix: #478

### Description

I applied `prettier` and added the new `this.meta.watchMode` (see https://github.com/rollup/rollup/pull/3616). Checking `process.env.ROLLUP_WATCH` will only work in `rollup -w` context. Checking for `this.meta.watchMode` will also work in `rollup.watch` context. In the future plugins should preferable utilize `this.meta.watchMode`.

### DIFF
--- a/packages/typescript/src/index.ts
+++ b/packages/typescript/src/index.ts
@@ -58,7 +58,7 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
     },
 
     buildEnd() {
-      if (process.env.ROLLUP_WATCH !== 'true') {
+      if (this.meta.watchMode !== true) {
         // ESLint doesn't understand optional chaining
         // eslint-disable-next-line
         program?.close();
--- a/packages/typescript/src/index.ts
+++ b/packages/typescript/src/index.ts
@@ -93,7 +93,7 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
     async load(id) {
       if (!filter(id)) return null;
 
-      await watchProgramHelper.wait()
+      await watchProgramHelper.wait();
 
       const output = findTypescriptOutput(ts, parsedOptions, id, emittedFiles);
 
--- a/packages/typescript/package.json
+++ b/packages/typescript/package.json
@@ -44,7 +44,7 @@
     "es2015"
   ],
   "peerDependencies": {
-    "rollup": "^1.20.0||^2.0.0",
+    "rollup": "^2.14.0",
     "tslib": "*",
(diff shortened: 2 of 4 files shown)


dionysiusmarquis:
@shellscape How can I ensure the tests are running against `rollup@^2.14.0`? I did not add any new functionality. I just changed one variable to test against before killing the Typescript watch program.

shellscape:
@dionysiusmarquis the same as you would any other package - you need to ensure that the right version is installed in devDeps. Because this will rely upon a particular patch version, the PR contains breaking changes, and that means it must be bumped to a new major version. The breaking change is that anyone using prior versions must now have a peerDep of rollup 2.14.0+

dionysiusmarquis:
@shellscape I don't know if I understand you correctly. 

Actually this won‘t change or break anything if you use an older `rollup` version. It will behave exactly like it would before this PR. This PR will  reenable the ability to use the `watch` Hook in `rollup.watch` context. You need `rollup@^2.14.0` if you want to utilize the Typescript Interface of `PluginContextMeta`. Which, as far as I can tell only necessary in the test environment.
If you use this plugin with an older `rollup` version from `node_modules` `this.meta.watchMode` will be `undefined` and everything will behave just like before this PR.

If you mean we have to ensure that the `watch` hook will always be utilized in any `rollup` watch context coherently. Then yes, we should change the peerDep.

packages/typescript/src/index.ts
```diff
@@ -58,7 +58,11 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
     },
 
     buildEnd() {
-      if (process.env.ROLLUP_WATCH !== 'true') {
+      /** TODO: remove 'process.env.ROLLUP_WATCH' and only check for 'this.meta.watchMode' in the future
+       * 'this.meta.watchMode' is supported since rollup@2.14.0
+       * @see https://github.com/rollup/rollup/pull/3616
+       **/
+      if (process.env.ROLLUP_WATCH !== 'true' && this.meta.watchMode !== true) {
```
NotWoods, line 65: If we're increasing the peer dependency, is there any reason to leave the env check?

PR #455, opened 2020-06-14
Html plugin: add templatePath option

joeuy:
## Rollup Plugin Name: `html`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description
It is very common that custom html is required for different frameworks. Example for `react` a `<div id="root"></div>` is required to render the application. Another use case could be to include ganalytics. 

This change allows the html plugin to receive a templatePath, so it can work with custom html. 

Existing `template` option is very verbose, as there is a lot of logic that could not be re-used.


joeuy:
If this change makes sense I will write some docs and fix the tests. Thank you!

shellscape:
Thanks for opening a PR. As a friendly tip, it's usually better to open a feature issue to discuss if the project is open to a particular feature before implementing it. 

I vote 👎  for this feature. We haven't seen any calls for it in issues, and I'm not comfortable with adding one-off features that won't get widespread use. Adding to that, this introduces a fair amount of refactoring and complexity that users could easily manage. Users of the plugins using `readFile` in the `template` option would then have unlimited flexibility, and would avoid calls for additional features around loading the the template from file. Ideally, reading from the filesystem should not be the job of the `html` plugin.

We'll see how other maintainers weigh in before making a decision.

joeuy:
@shellscape appreciate the quick response! I opened the PR because I wrote this code for my project, so It didn't take me a lot of extra time. 

Could we `export` something like `getTemplateData` ? its supper handy for custom `templates`

joeuy:
One unrelated question, since I got your attention. Given the following configuration: 

```
export default {
  input: 'main.js',
  output: {
    file: 'public/bundle.js',
    format: 'es',
  },
  manualChunks: {
    vendor: ['react', 'react-dom'],
  },
  plugins: [
    resolve(), // tells Rollup how to find date-fns in node_modules
    commonjs(), // converts date-fns to ES modules
    production && terser() // minify, but only in production
  ]
}
};```

Is it possible to have `vendor` chunk rendered directly into the html by the `html` plugin?

PR #456, opened 2020-06-15
add .nodeResolve change to the readme

heavyk:
just ran into what's described in #412

this updates the readme to reflect the latest version of the module

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### DIFF
--- a/packages/node-resolve/README.md
+++ b/packages/node-resolve/README.md
@@ -28,7 +28,7 @@ npm install @rollup/plugin-node-resolve --save-dev
 Create a `rollup.config.js` [configuration file](https://www.rollupjs.org/guide/en/#configuration-files) and import the plugin:
 
 ```js
-import resolve from '@rollup/plugin-node-resolve';
+import { nodeResolve } from '@rollup/plugin-node-resolve';
 
 export default {
   input: 'src/index.js',
@@ -36,7 +36,7 @@ export default {
     dir: 'output',
     format: 'cjs'
   },
-  plugins: [resolve()]
+  plugins: [nodeResolve()]
 };
 ```
 


vladshcherbin:
Why?

Default export works fine in `.js` configs, it's an issue with using `.mjs` configs which should be solved so that default exports work there too.

heavyk:
right, I'm using require. all the other plugins export the function, cept this one which I have to get `.nodeResolve` or `.default` eg.

```js
const plugin_json = require('@rollup/plugin-json') // works
const plugin_inject = require('@rollup/plugin-inject') // works
const plugin_node_resolve = require('@rollup/plugin-node-resolve').nodeResolve // this one needs .nodeResolve
const plugin_commonjs = require('@rollup/plugin-commonjs') // works
```

vladshcherbin:
I believe the real issue here is that it works differently and should be solved by updating its code to work like the other ones and not by updating readme, but I might be wrong 😉

vladshcherbin:
@heavyk can you check node-resolve v8.1.0?

I believe it should work fine now and there is no need to change readme.

XhmikosR:
The issue is still not fixed. `const { resolve } = require('@rollup/plugin-node-resolve')` doesn't work even in v8.1.0.

shellscape:
> The issue is still not fixed. const { resolve } = require('@rollup/plugin-node-resolve') doesn't work even in v8.1.0.

@XhmikosR the code cited doesn't exist in the readme at present.

XhmikosR:
True, but the way listed here seems to still work as a replacement for the old way of `require`ing packages.

heavyk:
@vladshcherbin with the latest 8.1.0, it still exports an object instead of a function:

```
> require('@rollup/plugin-node-resolve')
{
  DEFAULTS: {
    customResolveOptions: {},
    dedupe: [],
    extensions: [ '.mjs', '.js', '.json', '.node' ],
    resolveOnly: []
  },
  default: [Function: nodeResolve],
  nodeResolve: [Function: nodeResolve]
}
> require('@rollup/plugin-commonjs')
[Function: commonjs]
```

shellscape:
Thanks for weighing in everyone.

PR #462, opened 2020-06-19
feat(babel): add typings

AndrewLeedham:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

Adds TypeScript typings to the `@rollup/plugin-babel` so the plugin works nicely in TypeScript environments. I used the `@rollup/plugin-commonjs` typings as a template for this :)


packages/babel/test/types.ts
```diff
@@ -0,0 +1,22 @@
+import babel from '../types';
```
NotWoods, line 1:
```suggestion
import babel from '../';
```

packages/babel/types/index.d.ts
```diff
@@ -0,0 +1,38 @@
...
+   */
+  skipPreflightCheck?: boolean;
+}
+
+/**
+ * A Rollup plugin for seamless integration between Rollup and Babel.
+ * @param options - Plugin options.
+ * @returns Plugin instance.
+ */
+export default function babel(options?: RollupBabelOptions): Plugin;
```
NotWoods, line 38: It looks like we're missing some of the exports from the plugin.

In addition to `default`, the following should be exported:
- getBabelInputPlugin
- getBabelOutputPlugin
- createBabelInputPluginFactory 
- createBabelOutputPluginFactory
- babel

packages/babel/types/index.d.ts
```diff
@@ -0,0 +1,38 @@
+import { Plugin } from 'rollup';
+import { FilterPattern } from '@rollup/pluginutils';
+import { TransformOptions } from '@babel/core';
+
+interface RollupBabelOptions extends Omit<TransformOptions, 'include' | 'exclude'> {
```
Andarist, line 5: this should be named something like `RollupBabelInputPluginOptions`

packages/babel/package.json
```diff
@@ -58,6 +60,7 @@
     "@babel/plugin-transform-runtime": "^7.7.4",
     "@babel/preset-env": "^7.9.0",
     "@rollup/plugin-json": "^4.0.0",
+    "@types/babel__core": "^7.1.8",
```
Andarist, line 63: according to Maël Nison (Yarn2 maintainer) this should be also be specified as optional peer dependency, which can be done by adding this to package.json:
```
  "peerDependenciesMeta": {
    "@types/babel__core": {
      "optional": true
    }
  },
```

This might create some usability issues though as one doesn't necessarily will have those types installed and thus our types for `@rollup/plugin-babel` won't work. Maybe I'm worried about this prematurely though, I'm ok with shipping this (with this optional peer dep specified) and just seeing what happens.

packages/babel/test/types.ts
AndrewLeedham: image

packages/babel/package.json
AndrewLeedham: Annoyingly this makes the typings for the `test:ts` run become `any`. Hence the tests are passing, even though Ideally we need https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45613

NotWoods: At least for local tests, we could add a devDependency too.

packages/babel/types/index.d.ts
```diff
@@ -0,0 +1,130 @@
...
+export function createBabelOutputPluginFactory(
+  customCallback?: CustomOutputPluginBuilder
+): typeof getBabelOutputPlugin;
+
+/**
+ * A Rollup plugin for seamless integration between Rollup and Babel.
+ * @param options - Plugin options.
+ * @returns Plugin instance.
+ */
+export default function babel(options?: RollupBabelInputPluginOptions): Plugin;
```
NotWoods, line 130: `babel` is also a named export

```suggestion
export function babel(options?: RollupBabelInputPluginOptions): Plugin;
export default babel;
```

packages/babel/package.json
AndrewLeedham: Good idea 👍

AndrewLeedham:
FYI this shouldn't be merged until https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45613 is included as it breaks locally!

packages/babel/package.json
```diff
@@ -44,7 +46,8 @@
   ],
   "peerDependencies": {
     "@babel/core": "^7.0.0",
-    "rollup": "^1.20.0||^2.0.0"
+    "rollup": "^1.20.0||^2.0.0",
+    "@types/babel__core": "^7.1.8"
```
Andarist, line 50: q: do we require this version specifically? wouldn't things work for `^7.0.0`?

packages/babel/types/index.d.ts
```diff
@@ -0,0 +1,131 @@
...
+  skipPreflightCheck?: boolean;
+}
+
+interface RollupBabelOutputPluginOptions
+  extends Omit<babelCore.TransformOptions, 'include' | 'exclude'> {
+  /**
+   * A minimatch pattern, or array of patterns, which specifies the files in the build the plugin should operate on. When relying on Babel configuration files you cannot include files already excluded there.
+   * @default undefined;
+   */
+  include?: FilterPattern;
```
Andarist, line 40: `exclude`, `include` & `extensions` should not be included here:
https://github.com/rollup/plugins/blob/1523383fdf69a11480bd8094232d065c4ea44126/packages/babel/src/index.js#L180

packages/babel/types/index.d.ts
```diff
@@ -0,0 +1,131 @@
...
+  /**
+   * A boolean value indicating whether to bundle in the Babel helpers.
+   * @default false
+   */
+  externalHelpers?: boolean;
+  /**
+   * An array which gives explicit control over which babelHelper functions are allowed in the bundle.
+   * @default undefined
+   */
+  externalHelpersWhitelist?: string[];
```
Andarist, line 60: Similarly, I don't think `externalHelpers` & `externalHelpersWhitelist` are valid for output plugins - they are just unpacked & discarded here:
https://github.com/rollup/plugins/blob/1523383fdf69a11480bd8094232d065c4ea44126/packages/babel/src/index.js#L194-L203
without being used anywhere in the output plugins.

packages/babel/types/index.d.ts
```diff
@@ -0,0 +1,131 @@
...
+  /**
+   * @default false
+   */
+  runtimeHelpers?: boolean;
```
Andarist, line 64: This option doesn't exist for any type of this plugin in the current major version.

packages/babel/types/index.d.ts
```diff
@@ -0,0 +1,131 @@
...
+  /**
+   * Use with other formats than UMD/IIFE.
+   * @default false
+   */
+  allowAllFormats?: boolean;
+}
+
+type CustomInputPluginOptions = (
+  options: RollupBabelInputPluginOptions & Record<string, any>
```
Andarist, line 73: TS is not my strongest suit so take my comment with a grain of salt - maybe instead of accepting `& Record<string, any>` here this should be parametrized with a generic?

packages/babel/types/index.d.ts
```diff
@@ -0,0 +1,131 @@
+import { Plugin } from 'rollup';
+import { FilterPattern } from '@rollup/pluginutils';
+import * as babelCore from '@babel/core';
+
+interface RollupBabelInputPluginOptions
```
Andarist, line 5: shouldnt those interfaces be exported?

packages/babel/types/index.d.ts
AndrewLeedham: I didn't think the `customOptions` should be a 1-1 mapping, as that would limit the custom plugin, so they could not remove or add options. I think this way the user could build stronger on top if necassary.

packages/babel/package.json
AndrewLeedham: It need to be at least `7.1.9`, so the `hasFilesystemConfig` option exists.

AndrewLeedham:
> FYI this shouldn't be merged until [DefinitelyTyped/DefinitelyTyped#45613](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45613) is included as it breaks locally!

This was published and I updated the version here, so good to merge now :)

shellscape:
@Andarist if you think this looks good to go, we'll merge this. As the custodian of this plugin, I'd like to get your thumbs up first.

shellscape:
ping @Andarist

packages/babel/package.json
```diff
@@ -44,7 +46,8 @@
   ],
   "peerDependencies": {
     "@babel/core": "^7.0.0",
-    "rollup": "^1.20.0||^2.0.0"
+    "rollup": "^1.20.0||^2.0.0",
+    "@types/babel__core": "^7.1.9"
```
Andarist, line 50: nit: those keys here should be sorted alphabetically

packages/babel/test/types.ts
```diff
@@ -0,0 +1,149 @@
...
+
+import babelPlugin, {
+  babel,
+  getBabelInputPlugin,
+  getBabelOutputPlugin,
+  createBabelInputPluginFactory,
+  createBabelOutputPluginFactory
+} from '../types';
+
+const rollipConfig: import('rollup').RollupOptions = {
```
Andarist, line 11: typo: rollip 😅

packages/babel/types/index.d.ts
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * An array of file extensions that Babel should transpile. If you want to transpile TypeScript files with this plugin it's essential to include .ts and .tsx in this option.
+   * @default ['.js', '.jsx', '.es6', '.es', '.mjs']
+   */
+  extensions?: string[];
+  /**
+   * It is recommended to configure this option explicitly (even if with its default value) so an informed decision is taken on how those babel helpers are inserted into the code.
+   * @default 'bundled'
+   */
+  babelHelpers?: 'bundled' | 'runtime' | 'inline' | 'extends';
```
Andarist, line 26:
```suggestion
  babelHelpers?: 'bundled' | 'runtime' | 'inline' | 'external';
```

packages/babel/types/index.d.ts
```diff
@@ -0,0 +1,102 @@
...
+  customOptions: Record<string, any>;
+  pluginOptions: RollupBabelInputPluginOptions;
+};
+export type CustomOutputPluginOptions = (
+  options: RollupBabelOutputPluginOptions & Record<string, any>
+) => {
+  customOptions: Record<string, any>;
+  pluginOptions: RollupBabelOutputPluginOptions;
+};
+export type CustomPluginConfig = (
```
Andarist, line 55: Hm, I find it a little bit odd that types related to the most common exports have very specific names like `RollupBabelInputPluginOptions` but those related to more esoteric exports have rather generic names like `CustomPluginConfig`. I believe that we should unify those - one way or another.

I understand those specific names are getting really mouthful though - what is the convention used by other plugins in this monorepo? maybe exporting a `RollupBabel` namespace would make this slightly more approachable? with that we could have types like `RollupBabel.InputPluginOptions` etc

AndrewLeedham: 👍  Done

packages/babel/types/index.d.ts
AndrewLeedham: 👍  Done

packages/babel/test/types.ts
AndrewLeedham: 👍  Done

packages/babel/package.json
AndrewLeedham: 👍  Done

PR #469, opened 2020-06-22
docs(node-resolve): Fix mainFields valid values

dasa:

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

The valid value for `mainFields` should be `jsnext:main` instead of `jsnext`.

### DIFF
--- a/packages/node-resolve/README.md
+++ b/packages/node-resolve/README.md
@@ -107,7 +107,7 @@ Locks the module search within specified path (e.g. chroot). Modules defined out
 
 Type: `Array[...String]`<br>
 Default: `['module', 'main']`<br>
-Valid values: `['browser', 'jsnext', 'module', 'main']`
+Valid values: `['browser', 'jsnext:main', 'module', 'main']`
 
 Specifies the properties to scan within a `package.json`, used to determine the bundle entry point. The order of property names is significant, as the first-found property is used as the resolved entry point. If the array contains `'browser'`, key/values specified in the `package.json` `browser` property will be used.
 


shellscape:
> bugfixes and features will not be merged without tests

Also, please cite your sources for asserting that this new default value is correct versus the old.

dasa:
Example usage:

https://github.com/rollup/plugins/blob/ee5be6a7d8709e42821f6aeabbcad18a4af4dc66/packages/node-resolve/test/test.js#L18

shellscape:
@dasa unfortunately that's not what I asked for. if you're going to claim that the current implementation is wrong, you need to cite sources to back that claim up. citing your own code is not citing a source.

dasa:
> citing your own code is not citing a source.

That code is from this repo.

Here's a blog article: https://shuheikagawa.com/blog/2017/01/05/main-jsnext-main-and-module/

Here's where I ran into this: https://github.com/moment/moment/blob/develop/package.json#L26

shellscape:
Thanks for adding that info. We'll have some maintainers take a look.

Andarist:
@shellscape actually any strings are valid, you can make the node-resolve plugin to load packages based on any, arbitrary, "main" fields in package.json manifests. However, indeed `jsnext:main` has been used in the wild (and not `"jsnext"`) in the past, for example this page mentions that: https://github.com/rollup/rollup/wiki/pkg.module

dasa:
Perhaps the `Valid values` heading should be changed as well then to something else like `Common values`?

shellscape:
Thanks for weighing in everyone

PR #471, opened 2020-06-23
fix(strip): support specifying both functions and labels

dgkimpton:
## Rollup Plugin Name: `strip`

This PR contains:
- [ ] bugfix
- [x] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?
- [x] yes
- [ ] no

Breaking Changes?
- [x] yes 
- [ ] no

#167

### Description

When trying to use the **strip** plugin, I was for ages unable to get it to remove my labelled code. After some investigation, it became clear that the 'first pass filter' was rejecting labels when functions were specified.

This pull request fixes the exclusive nature of labels and functions. It is now possible to specify both, either, or neither.

I later found this 'closed' issue that reports the same problem, #167, it appears to have been started and then abandoned - hope I'm not treading on toes by fixing it.

Along the way, I had some trouble with the unit tests and got annoyed that I couldn't see the input and expected output together, so I refactored them and then added some more.

During testing, I further discovered that labels and functions were being ignored when spaces/newlines were included around the punctuation, so I fixed that too. Probably that should have been a different PR, but since my use-case had spaces in it, I just kept going until it worked. Sorry about that!


dgkimpton:
I see your point. That would be a bit of a hairy build process, but I suppose it is conceivable.  The only way I can think to resolve that is to add a new option with a different name.  
E.g. labels:[] has the current functionality and labeled:[] works in the none exclusive mode.  
That can definitely be done but explodes the complexity quite a bit... is that your preferred option, or can you think of an alternative?

pnevares:
I would probably prefer marking it as a breaking change and a major version bump for the package, instead of adding the technical debt of the newly-named option. Somebody would have an item backlogged to clean that up for the next version anyway.

packages/strip/README.md
```diff
@@ -36,59 +36,88 @@ export default {
     dir: 'output',
     format: 'cjs'
   },
-  plugins: [
-    strip({
-      labels: ['unittest']
-    })
-  ]
+  plugins: [strip({})]
```
shellscape, line 39: please revert this change

packages/strip/README.md
```diff
@@ -36,59 +36,88 @@ export default {
...
 };
 ```
 
 Then call `rollup` either via the [CLI](https://www.rollupjs.org/guide/en/#command-line-reference) or the [API](https://www.rollupjs.org/guide/en/#javascript-api).
 
+### Default Options
```
shellscape, line 45: please remove this section. the defaults are spelled out in the options below

packages/strip/README.md
```diff
@@ -36,59 +36,88 @@ export default {
...
+
+By default the **strip** plugin applies the following options, you can override by specifing options manually.
+
+> consider files matching **'\*\*/\*.js'**  
+> remove functions matching '**console.\***' and '**assert.\***'  
+> remove any **debugger** statements
+
+### Custom Options Example
```
shellscape, line 53: please remove this section. you've added examples in the options below and that's sufficient

packages/strip/README.md
```diff
@@ -36,59 +36,88 @@ export default {
...
 
 ### `debugger`
 
 Type: `Boolean`<br>
-Default: `true`
+Default: `true`<br>
+Example: `debugger: false,`<br>
 
-If `true`, instructs the plugin to remove debugger statements.
+If `true` or unspecified, instructs the plugin to remove debugger statements.
```
shellscape, line 92: `or unspecified` is not not necessary and doesn't add any value here, please revert.

packages/strip/src/index.js
```diff
@@ -38,21 +38,34 @@ export default function strip(options = {}) {
   const sourceMap = options.sourceMap !== false;
 
   const removeDebuggerStatements = options.debugger !== false;
-  const functions = (options.functions || ['console.*', 'assert.*']).map((keypath) =>
```
shellscape, line 41: please avoid "personal preference" refactors like this. please merge your changes from lines 45-47 back to this code block and remove the extra `functionsPatterns` variable. this make look cleaner in your eyes, but it adds extra complexity to the code that's not necessary.

packages/strip/src/index.js
```diff
@@ -38,21 +38,34 @@ export default function strip(options = {}) {
...
   const labels = options.labels || [];
 
-  const firstpass = new RegExp(`\\b(?:${functions.join('|')}|debugger)\\b`);
-  const pattern = new RegExp(`^(?:${functions.join('|')})$`);
+  const functionsPatterns = functions.map((f) =>
+    f.replace(/\*/g, '\\w+').replace(/\./g, '\\.\\s*')
+  );
+  const labelsPatterns = labels.map((l) => `${l}\\s*:`);
+
+  const firstPassPatterns = removeDebuggerStatements
```
shellscape, line 50: this ternary and the following array assignment within is unnecessary. 

```js
const firstPass = [...functionsPatterns, ...labelsPatterns];
if (removeDebuggerStatements) {
  firstPass.push('debugger\\b');
}
```

packages/strip/src/index.js
```diff
@@ -38,21 +38,34 @@ export default function strip(options = {}) {
...
+    ? [...functionsPatterns, ...labelsPatterns, 'debugger\\b']
+    : [...functionsPatterns, ...labelsPatterns];
+
+  const functionsRE = new RegExp(`^(?:${functionsPatterns.join('|')})$`);
```
shellscape, line 54: while we're in here, please use the naming pattern `reFunctions` and `reFirstpass`

packages/strip/README.md
dgkimpton: ok

packages/strip/README.md
dgkimpton: I added this because I wished it had been here when I started trying to use this plugin, but ok.

packages/strip/README.md
dgkimpton: ok

packages/strip/README.md
dgkimpton: ok

packages/strip/src/index.js
dgkimpton: ok

packages/strip/src/index.js
dgkimpton: Ooops, that would be my C++ background where const is deep. I'd forgotten in JS const is shallow and still allows the object to be mutated. Done.

packages/strip/src/index.js
dgkimpton: done

dgkimpton:
OK, I figured that would probably be the case. I modified the tests to make it easier for me to verify the fix and then decided just to offer them up as-is in case you preferred them (no point doing extra work if not needed). Changed.

shellscape:
Thanks. Please clean up those file mode changes so they're not part of the PR and we should be good to merge.

packages/strip/README.md
```diff
@@ -51,44 +51,54 @@ Then call `rollup` either via the [CLI](https://www.rollupjs.org/guide/en/#comma
...
 
 ### `sourceMap`
 
 Type: `Boolean`<br>
-Default: `true`
+Default: `true`<br>
+Example: `sourceMap: false,`<br>
 
-If `true`, instructs the plugin to update source maps accordingly after removing configured targets from the bundle.
+If `true` instructs the plugin to update source maps accordingly after removing configured targets from the bundle.
```
shellscape, line 101:
```suggestion
If `true`, instructs the plugin to update source maps accordingly after removing configured targets from the bundle.
```

packages/strip/src/index.js
```diff
@@ -38,21 +38,34 @@ export default function strip(options = {}) {
   const sourceMap = options.sourceMap !== false;
 
   const removeDebuggerStatements = options.debugger !== false;
+
```
shellscape, line 41:
```suggestion
```

packages/strip/src/index.js
```diff
@@ -39,20 +39,32 @@ export default function strip(options = {}) {
...
+  if (removeDebuggerStatements) {
+    firstPass.push('debugger\\b');
+  }
+
+  const reFunctions = new RegExp(`^(?:${functions.join('|')})$`);
+  const reFirstpass = new RegExp(`\\b(?:${firstPass.join('|')})`);
+
+  const firstPassFilter = firstPass.length > 0 ? (code) => reFirstpass.test(code) : () => false;
+
+  const UNCHANGED = null;
```
shellscape, line 59:
```suggestion
  const reFirstpass = new RegExp(`\\b(?:${firstPass.join('|')})`);
  const firstPassFilter = firstPass.length > 0 ? (code) => reFirstpass.test(code) : () => false;
  const UNCHANGED = null;
```

shellscape:
thanks!

dasa:
It looks like there was a publishing problem and this PR wasn't included in the 2.0.0 release: https://unpkg.com/@rollup/plugin-strip@2.0.0/dist/index.es.js

For example, the published file is still using `if (!edited) return null;`.

dasa:
Fixed in 2.0.1: https://unpkg.com/@rollup/plugin-strip@2.0.1/dist/index.es.js

👍🏻

PR #487, opened 2020-07-06
feat(node-resolve): preserve search params and hashes

LarsDenBakker:

This PR contains:

- [ ] bugfix
- [X] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [X] no

List any relevant issue numbers: Fixes https://github.com/rollup/plugins/issues/486

### Description

This makes node-resolve preserve any search params or hashes if they were in the import path.

Code:

```js
import foo from 'my-package?foo=bar';
```

Before this change it would throw, saying it can't find `my-package?foo=bar'`.

After this change, it gets resolved to something like:

```js
import foo from '/absolute/file/system/path/to/my-package?foo=bar';
```

This would still throw as rollup tries to find this file, but a rollup plugin can set up a special load for a module with this parameter.

### DIFF
--- a/packages/node-resolve/src/index.js
+++ b/packages/node-resolve/src/index.js
@@ -90,6 +90,12 @@ export function nodeResolve(opts = {}) {
       // ignore IDs with null character, these belong to other plugins
       if (/\0/.test(importee)) return null;
 
+      // strip hash and query params from import
+      const [withoutHash, hash] = importee.split('#');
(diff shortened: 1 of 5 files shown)


shellscape:
@LarsDenBakker we'll give this another day or so for folks to put eyes on it before we merge.

PR #515, opened 2020-07-28
fix(babel): Pass chunk filename to output plugin transform

evocateur:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: N/A

### Description

This _appears_ to be an oversight, as the `filename` is passed at the same place in the input plugin transform Babel options. Without this field, a Babel error including text `"Preset /* your preset */ requires a filename to be set when babel is
called directly"` is thrown when using a custom preset.

    [!] (plugin babel) Error: [BABEL] unknown: Preset /* your preset */ requires a filename to be set when babel is called directly,
    ```
    babel.transform(code, { filename: 'file.ts', presets: [/* your preset */] });
    ```
    See https://babeljs.io/docs/en/options#filename for more information.
    Error: [BABEL] unknown: Preset /* your preset */ requires a filename to be set when babel is called directly,
    ```
    babel.transform(code, { filename: 'file.ts', presets: [/* your preset */] });
    ```
    See https://babeljs.io/docs/en/options#filename for more information.
        at validateIfOptionNeedsFilename ($PWD/node_modules/@babel/core/lib/config/full.js:274:11)
        at options.overrides.forEach.overrideOptions ($PWD/node_modules/@babel/core/lib/config/full.js:286:52)
        at Array.forEach (<anonymous>)
        at validatePreset ($PWD/node_modules/@babel/core/lib/config/full.js:286:25)
        at loadPresetDescriptor ($PWD/node_modules/@babel/core/lib/config/full.js:293:3)
        at loadPresetDescriptor.next (<anonymous>)
        at recurseDescriptors ($PWD/node_modules/@babel/core/lib/config/full.js:107:30)
        at recurseDescriptors.next (<anonymous>)
        at recurseDescriptors ($PWD/node_modules/@babel/core/lib/config/full.js:128:34)
        at recurseDescriptors.next (<anonymous>)

(stack trace is using `@babel/core@7.10.5`)

The test is not ideal, but without replicating a huge stack of other Rollup plugins and complicated Babel config, it's unclear how else to regress this in a focused, minimal way. A local edit of the Rollup Babel plugin in `node_modules` of the offending build does indicate the error is mitigated, however.

### DIFF
--- a/packages/babel/src/index.js
+++ b/packages/babel/src/index.js
@@ -203,7 +203,13 @@ function createBabelOutputPluginFactory(customCallback = returnObject) {
         } = unpackOutputPluginOptions(pluginOptionsWithOverrides, outputOptions);
         /* eslint-enable no-unused-vars */
 
-        return transformCode(code, babelOptions, overrides, customOptions, this);
+        return transformCode(
+          code,
+          { ...babelOptions, filename: chunk.fileName },
+          overrides,
+          customOptions,
+          this
+        );
       }
     };
   };
--- a/packages/babel/test/as-output-plugin.js
+++ b/packages/babel/test/as-output-plugin.js
@@ -332,7 +332,7 @@ test('allows using Babel to transform to other formats', async (t) => {
       exports: {}
     };
     factory();
-    global.unknown = mod.exports;
+    global.main = mod.exports;
   }
 })(typeof globalThis !== "undefined" ? globalThis : typeof self !== "undefined" ? self : this, function () {
   "use strict";
@@ -356,4 +356,21 @@ const answer = Math.pow(42, 2);
 console.log(\`the answer is \${answer}\`);
 `
   );
+});
+
+test('passes chunk filename to transform', async (t) => {
+  t.plan(2);
+
+  const customOutputPlugin = createBabelOutputPluginFactory(() => {
+    return {
+      config(cfg) {
+        // eslint-disable-next-line no-undefined
+        t.not(cfg.options.filename, undefined, 'cfg.options.filename wasn’t set');
+        t.is(nodePath.relative(cfg.options.cwd, cfg.options.filename), 'main.js');
+        return cfg.options;
+      }
+    };
+  });
+  const bundle = await rollup({ input: 'fixtures/basic/main.js' });
+  await getCode(bundle, { format: 'cjs', plugins: [customOutputPlugin()] });
 });


evocateur:
Ooh, I think I figured out the trigger, and it's much simpler than I thought: `@babel/preset-typescript` is [configuring the `test` field](https://github.com/babel/babel/blob/6373eb5d82d527f9b86a0145bdd2c4c9970a16eb/packages/babel-preset-typescript/src/index.js#L49), which is what triggers [validateIfOptionNeedsFilename](https://github.com/babel/babel/blob/601c824873a2606b0292ea1460a3a79f5f704215/packages/babel-core/src/config/full.js#L325) to explode. The workaround probably involves setting a preset option like `allExtensions`, but at least it promises to make the test here a bit more specific and less brittle.

evocateur:
Additionally, in a repo where `@babel/preset-typescript` is _not_ enabled, it looks like a Babel config file with `overrides` will trip up `getBabelOutputPlugin()` too:
```
[!] (plugin babel) Error: Configuration contains string/RegExp pattern, but no filename was passed to Babel
Error: Configuration contains string/RegExp pattern, but no filename was passed to Babel
    at matchPattern ($PWD/node_modules/@babel/core/lib/config/config-chain.js:430:11)
    at patterns.some.pattern ($PWD/node_modules/@babel/core/lib/config/config-chain.js:417:35)
    at Array.some (<anonymous>)
    at matchesPatterns ($PWD/node_modules/@babel/core/lib/config/config-chain.js:417:19)
    at configFieldIsApplicable ($PWD/node_modules/@babel/core/lib/config/config-chain.js:399:10)
    at configIsApplicable ($PWD/node_modules/@babel/core/lib/config/config-chain.js:394:41)
    at forEach ($PWD/node_modules/@babel/core/lib/config/config-chain.js:255:13)
    at Array.forEach (<anonymous>)
    at $PWD/node_modules/@babel/core/lib/config/config-chain.js:252:42
    at Generator.next (<anonymous>)
```

The `.babelrc.js` in question:
```js
module.exports = {
    presets: [
        [
            '@babel/preset-env',
            {
                targets: {
                    ie: 11,
                },
            },
        ],
    ],
    overrides: [
        {
            test: [
                './src/stuff.js',
                './src/things/*.js',
            ],
            presets: [
                [
                    '@babel/preset-env',
                    {
                        exclude: ['@babel/plugin-transform-classes'],
                        targets: {
                            node: 'current',
                        },
                    },
                ],
            ],
        },
    ],
};
```

Andarist:
This was not implemented on purpose - see https://github.com/rollup/rollup-plugin-babel/pull/333#discussion_r346966187 . I'm too tired today to analyze this issue as a whole but maybe you could look at the linked comments to see how this applies to your case (or how it doesn't).

evocateur:
@Andarist Will do, thanks for the pointer!

evocateur:
Well, turns out I was barking up the wrong tree entirely. Output plugins were not the solution, I needed to pass an array of configs from my `rollup.config.js` instead so the per-config input plugin could do the right thing w/r/t the `supportsStaticESM` caller property.

Thanks for your responsiveness!

PR #525, opened 2020-08-03
Correct minimatch to picomatch in the docs

AlexVipond:
## Rollup Plugin Name: `createFilter`

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [X] documentation
- [ ] other

Are tests included?

- [X] no

Breaking Changes?

- [X] no

### Description

The current docs say that `createFilter` uses minimatch, but it actually uses picomatch.

https://github.com/rollup/plugins/blob/5fa15590e452d568239a6d8e1c7018865c027fe4/packages/pluginutils/src/createFilter.ts#L3

The differences are pretty subtle, but I think it's important to document them for people with more specific use cases, and for people creating packages that depend on `createFilter`.

### DIFF
--- a/packages/pluginutils/README.md
+++ b/packages/pluginutils/README.md
@@ -99,14 +99,16 @@ export default function myPlugin(options = {}) {
 
 Constructs a filter function which can be used to determine whether or not certain modules should be operated upon.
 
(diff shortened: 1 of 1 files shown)


shellscape:
thanks!

PR #529, opened 2020-08-05
refactor(node-resolve): remove deep-freeze and deepmerge from dependencies

alan-agius4:
This 2 dependencies are currently used to deepfreeze an exported symbol. Which seems to be an wasteful to add 2 extra dependencies for something trivial.

Also `deep-freeze` is licensed as Public Domain, which might be problematic for some 3rd parties such as the Angular CLI.

In Angular CLI we have a license validator that validates direct and transitive dependencies, and Public Domain is a problematic license becuse it falls under the "unencumbered' group which requires legal audit.

More context: https://opensource.google/docs/thirdparty/licenses/#unencumbered

### DIFF
--- a/packages/node-resolve/src/index.js
+++ b/packages/node-resolve/src/index.js
@@ -2,7 +2,6 @@
 import { dirname, normalize, resolve, sep } from 'path';
 
 import builtinList from 'builtin-modules';
-import deepFreeze from 'deep-freeze';
 import deepMerge from 'deepmerge';
 import isModule from 'is-module';
 
@@ -19,6 +18,17 @@ import {
 const builtins = new Set(builtinList);
 const ES6_BROWSER_EMPTY = '\0node-resolve:empty.js';
 const nullFn = () => null;
+const deepFreeze = object => {
+  Object.freeze(object);
+
+  for (const value of Object.values(object)) {
+    if (typeof value === 'object' && !Object.isFrozen(value)) {
+      deepFreeze(value);
+    }
+  }
+
+  return object;
+};
 const defaults = {
   customResolveOptions: {},
   dedupe: [],
--- a/packages/node-resolve/package.json
+++ b/packages/node-resolve/package.json
@@ -52,7 +52,6 @@
     "@rollup/pluginutils": "^3.1.0",
     "@types/resolve": "1.17.1",
     "builtin-modules": "^3.1.0",
-    "deep-freeze": "^0.0.1",
     "deepmerge": "^4.2.2",
     "is-module": "^1.0.0",
     "resolve": "^1.17.0"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -291,7 +291,6 @@ importers:
       '@rollup/pluginutils': 3.1.0_rollup@2.23.0
       '@types/resolve': 1.17.1
       builtin-modules: 3.1.0
-      deep-freeze: 0.0.1
       deepmerge: 4.2.2
       is-module: 1.0.0
       resolve: 1.17.0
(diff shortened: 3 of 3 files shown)


packages/node-resolve/src/index.js
```diff
@@ -19,15 +17,15 @@ import {
 const builtins = new Set(builtinList);
 const ES6_BROWSER_EMPTY = '\0node-resolve:empty.js';
 const nullFn = () => null;
-const defaults = {
-  customResolveOptions: {},
-  dedupe: [],
+const defaults = Object.freeze({
```
shellscape, line 20: I'd rather see a simple convenience method (`const deepFreeze = () =>`) in here to take the place of the removed dep on line 30, rather than a buncha calls to `Object.freeze` that we'd have to remember to add to anything new to `defaults`. A depth of 4 seems like it would be sufficient. The code _is_ [rather simple](https://github.com/substack/deep-freeze/blob/master/index.js).

TrySound:
Isn't it much easier to just duplicate defaults definition twice than adding a bunch of packages for very doubtful use case (laziness).

alan-agius4:
I was actually considering copy and pasting the defaults.

@shellscape, what do you think?

shellscape:
> Isn't it much easier to just duplicate defaults definition twice than adding a bunch of packages for very doubtful use case (laziness)?

man, I really take exception to that language. 

@alan-agius4 I'd ask that you follow the guidance offered in my review

TrySound:
By laziness I mean initial issue with adding ts extensions. It's really the only use case and requires two dependencies though it's not a big deal to list all necessary extensions.

shellscape:
> `deepmerge` is a guard and unfortunately I won't approve removing that.

Sorry guys, not backing down from that.

alan-agius4:
@shellscape, I added the deep freeze implementation.

alan-agius4:
deepmerge package seems to be unmaintained. They do have a PR (https://github.com/substack/deep-freeze/pull/20 ) to change the license type but is staled.

alan-agius4:
Can you explain what do you mean by deepmerge being a guard?

shellscape:
Dude, please use the edit feature rather than three consecutive replies. No one likes their notifications to go bananas. 

It's a guard to prevent mutating the exported defaults. Because this is compiled down to CJS, defaults could inadvertently be overwritten, causing program failure. Providing what is essentially a deep `Object.assign` copy of the object to the user prevents the plugin's defaults from being mutated even if they mess with the exports. 

That code was put in place precisely because someone did exactly what is described above and encountered errors.

alan-agius4:
@shellscape, apologies for the extra notifications.

I believe that `Object.freeze` should be sufficient here. Also adding 2 extra dependencies for a single use case for defensive programming seems an overkill, but that’s just my option.

If it makes you happy and we can merge this PR I can add back deepmerge, but again it seem redundant to me considering you are freezing the object itself and all of it’s sub properties.

shellscape:
> I believe that Object.freeze should be sufficient here.

Unfortunately that's not true. This can be simply demonstrated.

```js
// a.js
module.exports = { defaults: Object.freeze({ batman: 'batman' }) };
```

```console
→ node 
Welcome to Node.js v12.18.1.
> require('./a').defaults
{ batman: 'batman' }
> require('./a').defaults = null
null
> require('./a').defaults
null
```

alan-agius4:
@shellscape, I see what you mean now. Thanks for the explanation.

But honestly if someone is doing that, they should be punished and not handle such cases. I cannot think of a genuine use case that it would be okay to that.

Also possible in this case a top level `Object.assign({}, defaults)` should suffice.

shellscape:
> But honestly if someone is doing that, they should be punished and not handle such cases.

Thankfully, we have empathy for users in this project.

Change looks good. I'm on my phone right now and will merge when I return home.

shellscape:
@alan-agius4 next time please use our Pull Request template. normally a script I run on my machine would have picked up that up and closed the PR citing the missing template. Guess you got lucky :)

PR #534, opened 2020-08-09
docs(babel): Add warning to tell about configuration name change runtimeHelpers -> babelHelpers

renoirb:

This PR contains:

- [x] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

When a person upgrades from earlier version of `@rollup/plugin-babel` (say, [from the time of `rollup-plugin-babel` for a use-case such as the one in this issue](https://github.com/rollup/rollup-plugin-babel/issues/322)) which used `runtimeHelpers`, gets into error somewhere between other errors and unexpected outputs with a behavior change.

I wished, as a user of this fine package, that the error message would've told me about the variable name change.

### DIFF
--- a/packages/babel/src/index.js
+++ b/packages/babel/src/index.js
@@ -30,6 +30,21 @@ const unpackOptions = ({
 };
 
 const unpackInputPluginOptions = ({ skipPreflightCheck = false, ...rest }, rollupVersion) => {
+  if ('runtimeHelpers' in rest) {
+    const valueAsString = `runtimeHelpers: ${rest.runtimeHelpers}`;
+    const suggestion =
+      rest.runtimeHelpers === true
+        ? `try changing your configuration to babelHelpers: 'runtime'. `
+        : '';
+    // eslint-disable-next-line no-console
+    console.warn(
+      '"runtimeHelpers" is now called "babelHelpers". ' +
+        `At the moment, your configuration is set as ${valueAsString}. ` +
+        suggestion +
+        'Refer to the documentation to adjust to your use-case. ' +
+        'https://rollupjs.org/guide/en/#babel'
+    );
+  }
   if (!rest.babelHelpers) {
     // eslint-disable-next-line no-console
     console.warn(
--- a/packages/babel/src/index.js
+++ b/packages/babel/src/index.js
@@ -29,23 +29,26 @@ const unpackOptions = ({
   };
 };
 
+const warnAboutDeprecatedHelpersOption = ({ deprecatedOption, suggestion }) => {
+  // eslint-disable-next-line no-console
+  console.warn(
+    `\`${deprecatedOption}\` has been removed in favor a \`babelHelpers\` option. Try changing your configuration to \`${suggestion}\`. ` +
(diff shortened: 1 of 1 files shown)


packages/babel/src/index.js
```diff
@@ -31,11 +31,19 @@ const unpackOptions = ({
 
 const unpackInputPluginOptions = ({ skipPreflightCheck = false, ...rest }, rollupVersion) => {
   if (!rest.babelHelpers) {
-    // eslint-disable-next-line no-console
-    console.warn(
-      "babelHelpers: 'bundled' option was used by default. It is recommended to configure this option explicitly, read more here: " +
-        'https://github.com/rollup/plugins/tree/master/packages/babel#babelhelpers'
-    );
+    if ('runtimeHelpers' in rest) {
```
Andarist, line 34: IMHO this could be moved to the top of this function - one probably won't have both `babelHelpers` & `runtimeHelpers` set at the same time, so it doesn't matter much but less indentation the better.

packages/babel/src/index.js
```diff
@@ -31,11 +31,19 @@ const unpackOptions = ({
...
+      // eslint-disable-next-line no-console
+      console.warn(
```
Andarist, line 36: We can give a concrete recommendation on how the code should be adjusted, it's just a matter of changing `runtimeHelpers: true` to `babelHelpers: 'runtime'`. The link to the docs is still nice - but should just act as "learn more" kind of thing.

Andarist: Also - the same warning should be introduced for `externalHelpers: true` which is equivalent to `babelHelpers: 'external'`

renoirb: Yeah, I'm unsure of the mapping.

As I understood it, for me, `babelHelpers: true` was [about inlining dependencies](https://github.com/renoirb/experiments-201908-rush-typescript-just-bili-monorepo/blob/20200303-version-policies/conventions/use-bili/src/plugins/babel.ts#L37-L72).

So, at least a "learn more" is better than a warning about something I didn't know changed is good.

renoirb: Oh, the home page has *bit rot*, maybe we should fix that too.

Screen Shot 2020-08-10 at 6 55 49 PM

renoirb: Oh. I see.
*externalHelpers* and *runtimeHelpers* are mutually exclusive and now instead of two attributes with boolean, they're now as a string under *babelHelpers*.

Andarist: Yes, that's basically the idea behind those - the meaning of them hasn't changed much, it's only a couple of booleans converted to a single enum.

shellscape:
@Andarist this looks more like a `fix` or `chore` than a `docs` issue now. would you agree?

Andarist:
@shellscape right, the deprecation/migration warnings have been added to the runtime code. I'm not well-versed with semantic commit types, does `chore` ok for this? It ain't definitely fix, just a small DX improvement for migrating from v4

shellscape:
`chore` works great!

Issue #249, opened 2020-03-09
Second build fails when `incremental` set to `true`


Moeriki:
- Rollup Plugin Name: typescript
- Rollup Plugin Version: 4.0.0
- Rollup Version: 2.0.2
- Operating System (or Browser): macOS 10.15.3 (19D76)
- Node Version: 12.16.0

### How Do We Reproduce?

```
git clone git@github.com:Moeriki/bug-rollup-ts-incremental.git
cd bug-rollup-ts-incremental/
npm install
npm run build
npm run build
```

### Expected Behavior

Build succeeds two times.

### Actual Behavior

Build fails the second time.

### Work-around

```
npm run build
npm run clean
npm run build
```

Or set `compilerOptions.incremental` to `false`.

shellscape:
@Moeriki thanks for the issue. Please verify that this also happens when using NPM. NPM is our baseline in the project for issues and repo-based reproductions.

Moeriki:
Tested and verified the same issue with npm.

The issue though may be related to my compiler options. I'm still testing some things.

VitorLuizC:
Maybe this is the cause of https://github.com/rollup/plugins/issues/287

VitorLuizC:
Same happens here and we're using PNPM.

shellscape:
Please use the "edit" function on replies to add info and cut down on noise.

shellscape:
Hey folks. This issue hasn't received any traction for more than 90 days, so we're going to close this for housekeeping. If this is still an ongoing issue, please do consider contributing a Pull Request to resolve it. Further discussion is always welcome even with the issue closed. If anything actionable is posted in the comments, we'll consider reopening it.

VitorLuizC:
This was fixed?

shellscape:
@VitorLuizC please re-read my previous comment.

VitorLuizC:
I don't understand the concept of housekeeping. You're closing an issue that is not fixed to reduce number of open issues? I can't see the advantages in doing that.

shellscape:
It's quite simple really; we have finite resources and must focus on issues that have activity and active triage. Housekeeping is very common in large projects (most even have bots that handle it for the project). We haven't received any help from the community to resolve this, which is unfortunate but OK, so we're closing this until such time that someone would like to pick up the effort.

If you're interested in contributing a fix for this issue, please do open a PR and we'll get some folks to review it.

VitorLuizC:
[...]

PR #535, opened 2020-08-10
feat(typescript): Implement cached incremental code

mrwho:

This PR contains:

- [X] bugfix
- [X] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [X] no

### Description
  
  When running consecutive `incremental` builds in typescript any build after the first one will fail. This is due to the fact that the typescript compiler will not emit the code for cached files. When this happens the typescript files will be sent to the next hop in the compilation pipeline as javascript files.

  This PR adds caching when `incremental` or `composite` - this last one will enable `incremental` - so that any consecutive build will be able to use the cached compiled files.

  This is related to [Issue 249](https://github.com/rollup/plugins/issues/249)


packages/typescript/src/index.ts
```diff
@@ -42,6 +46,9 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
           resolveModule,
           parsedOptions,
           writeFile(fileName, data) {
+            if (parsedOptions.options.composite || parsedOptions.options.incremental) {
+              tsCache.cacheCode(fileName, data);
```
NotWoods, line 50: Is there a Typescript API available for this functionality? I find it odd that these flags require us to create our own cache implementation.

mrwho: Well, the issue is this: `writeFile` overwrites the default typescript output.
  
  So, normally: 

  [fs] main.ts => tsc => `writeFile` => [fs] main.js

this plugin changes the `writeFile` to work like this:

  [fs] main.ts => tsc => `writeFile` => [emittedFiles] main.js

   When dealing with an already compiled file that has not changed - according to the `tsbuildinfo` - `tsc` will not write the file, since it's already on disk. 

  With the plugin - during an incremental build - the first time we run rollup we will emit the files correctly into the next step of the pipeline - correctly transpiled into javascript. The second time we run rollup the plugin will not see the file in the `emittedFiles` - since it hasn't changed and `tsc` didn't call `writeFile` - and it will pass the un-transpiled typescript file to the next step of the pipeline. This will fail, since it isn't  a javascript file.

  This PR creates a local cache of the files emitted by tsc during incremental runs, the next time rollup runs the transpiled files will be picked up from the cache and sent forward.

  In other words: yes - typescript has a cache: the transpiled files, but we are not saving them - since we only want to transpile on the fly and move the on with our lives. This will break when using incremental builds.

  I hope I have clarified a bit more than I have confused.

KennyVaneetvelde (approved):

packages/typescript/src/index.ts
ghepesdoru: Hello everyone, sorry for jumping into the conversation, but I wanted to highlight a few questions/concernes regarding the cache implementation:
- there is no **process awareness** in default caching destination: can be **easy to mitigate** by **using separate cache paths**/process or build. This can be a major problem when dealing with multiple concurrent builds and would highly suggest **documenting the edge case**.
- TSCache works on **file names only** and is **unawareness of contents changes**: can be **easy to mitigate** by **deleting the cache folder** before new builds, but beats the purpose in my view. One of incremental's advantages is lowering the "cold" build (as opposed to watch mode) duration, and deletion of cache will negate that particular advantage. I would highly suggest using hashes based on file's contents or at least file stats (mtime, ctime etc) to mittigate the issue, this would also work well with common use cases such as: changing a branch, pull-ing from version control etc.

Please take into consideration that my last point could be muted, by the typescript compiler determining changes and emitting new files (this should be the expected behavior, and emitted files have precedence over cached versions in current implementation); but I wanted to raise awareness.

shellscape:
@mrwho @ghepesdoru @NotWoods don't want to let this one get too stale. where are we currently at with it?

NotWoods (approved):

shellscape:
@mrwho would you be able to resolve the conflicts?

PR #540, opened 2020-08-13
feat(node-resolve): support package entry points

LarsDenBakker:

This PR contains:

- [ ] bugfix
- [X] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] no

List any relevant issue numbers: Fixes https://github.com/rollup/plugins/issues/362

### Description

This implements package entrypoints as specified in https://nodejs.org/api/esm.html#esm_package_entry_points. 

For any bare import that is resolved we first look up the package.json. If the package.json has an export map we use that for resolving instead of the regular node resolve logic. If a path is not found in the export map, an error is thrown. I've tried to match the implementation of node, and also throw the same informative error messages as node js to ensure a consistent experience. @guybedford I would appreciate it if you would be able to verify the implementation. Consistency is very important here.

We need make a decision on how to the `exports` field should relate to other main fields and the `browser` field. The way I implemented it right if a package has package entrypoints defined we ignore any main fields configuration, but we do allow the `browser` field to override the export map result. This isn't really a conscious decision, it's just a result of how the code is structured right now.

This is a breaking change because it changes the way modules are resolved.

### DIFF
--- a/packages/node-resolve/README.md
+++ b/packages/node-resolve/README.md
@@ -34,15 +34,26 @@ export default {
   input: 'src/index.js',
   output: {
     dir: 'output',
-    format: 'cjs'
+    format: 'cjs',
   },
-  plugins: [nodeResolve()]
+  plugins: [nodeResolve()],
 };
 ```
 
 Then call `rollup` either via the [CLI](https://www.rollupjs.org/guide/en/#command-line-reference) or the [API](https://www.rollupjs.org/guide/en/#javascript-api).
 
 ## Options
+
+### `exportConditions`
+
+Type: `Array[...String]`<br>
+Default: `[]`
+
+Additional conditions of the package.json exports field to match when resolving modules. By default, this plugin looks for the `['default', 'module', 'import']` conditions when resolving imports.
+
+When using `@rollup/plugin-commonjs` v16 or higher, this plugin will use the `['default', 'module', 'require']` conditions when resolving require statements.
+
+Setting this option will add extra conditions on top of the default conditions. See https://nodejs.org/api/packages.html#packages_conditional_exports for more information.
 
 ### `browser`
 
@@ -164,9 +175,9 @@ export default {
   output: {
     file: 'bundle.js',
     format: 'iife',
-    name: 'MyModule'
+    name: 'MyModule',
   },
-  plugins: [resolve(), commonjs()]
+  plugins: [resolve(), commonjs()],
 };
 ```
 
@@ -185,6 +196,19 @@ export default ({
   external: builtins,
   output: ...
 })
+```
+
+## Resolving require statements
+
+According to [NodeJS module resolution](https://nodejs.org/api/packages.html#packages_package_entry_points) `require` statements should resolve using the `require` condition in the package exports field, while es modules should use the `import` condition.
+
+The node resolve plugin uses `import` by default, you can opt into using the `require` semantics by passing an extra option to the resolve function:
+
+```js
+this.resolve(importee, importer, {
+  skipSelf: true,
+  custom: { 'node-resolve': { isRequire: true } },
+});
 ```
 
 ## Meta
--- a/packages/node-resolve/src/index.js
+++ b/packages/node-resolve/src/index.js
@@ -7,13 +7,8 @@ import isModule from 'is-module';
 
 import { isDirCached, isFileCached, readCachedFile } from './cache';
 import { exists, readFile, realpath } from './fs';
-import {
-  getMainFields,
-  getPackageInfo,
-  getPackageName,
-  normalizeInput,
-  resolveImportSpecifiers
-} from './util';
+import { resolveImportSpecifiers } from './resolveImportSpecifiers';
+import { getMainFields, getPackageInfo, getPackageName, normalizeInput } from './util';
 
 const builtins = new Set(builtinList);
 const ES6_BROWSER_EMPTY = '\0node-resolve:empty.js';
@@ -29,6 +24,10 @@ const deepFreeze = (object) => {
 
   return object;
 };
+
+const baseConditions = ['default', 'module'];
+const baseConditionsEsm = [...baseConditions, 'import'];
+const baseConditionsCjs = [...baseConditions, 'require'];
 const defaults = {
   customResolveOptions: {},
   dedupe: [],
@@ -42,6 +41,8 @@ export const DEFAULTS = deepFreeze(deepMerge({}, defaults));
 export function nodeResolve(opts = {}) {
   const options = Object.assign({}, defaults, opts);
   const { customResolveOptions, extensions, jail } = options;
+  const conditionsEsm = [...baseConditionsEsm, ...(options.exportConditions || [])];
+  const conditionsCjs = [...baseConditionsCjs, ...(options.exportConditions || [])];
   const warnings = [];
   const packageInfoCache = new Map();
   const idToPackageInfo = new Map();
@@ -93,7 +94,7 @@ export function nodeResolve(opts = {}) {
       isDirCached.clear();
     },
 
-    async resolveId(importee, importer) {
+    async resolveId(importee, importer, opts) {
       if (importee === ES6_BROWSER_EMPTY) {
         return importee;
       }
@@ -222,7 +223,16 @@ export function nodeResolve(opts = {}) {
       importSpecifierList.push(importee);
       resolveOptions = Object.assign(resolveOptions, customResolveOptions);
 
-      let resolved = await resolveImportSpecifiers(importSpecifierList, resolveOptions);
+      const warn = (...args) => this.warn(...args);
+      const isRequire =
+        opts && opts.custom && opts.custom['node-resolve'] && opts.custom['node-resolve'].isRequire;
+      const exportConditions = isRequire ? conditionsCjs : conditionsEsm;
+      let resolved = await resolveImportSpecifiers(
+        importSpecifierList,
+        resolveOptions,
+        exportConditions,
+        warn
+      );
       if (!resolved) {
         return null;
       }
--- a/packages/node-resolve/src/resolveImportSpecifiers.js
+++ b/packages/node-resolve/src/resolveImportSpecifiers.js
@@ -1,0 +1,197 @@
+import fs from 'fs';
+import path from 'path';
+import { promisify } from 'util';
+
+import resolve from 'resolve';
+
+import { getPackageName } from './util';
+import { exists, realpath } from './fs';
+
+const resolveImportPath = promisify(resolve);
+const readFile = promisify(fs.readFile);
+
+const pathNotFoundError = (subPath, pkgPath) =>
+  new Error(`Package subpath '${subPath}' is not defined by "exports" in ${pkgPath}`);
+
+function findExportKeyMatch(exportMap, subPath) {
+  for (const key of Object.keys(exportMap)) {
+    if (key.endsWith('*')) {
+      // star match: "./foo/*": "./foo/*.js"
+      const keyWithoutStar = key.substring(0, key.length - 1);
+      if (subPath.startsWith(keyWithoutStar)) {
+        return key;
+      }
+    }
+
+    if (key.endsWith('/') && subPath.startsWith(key)) {
+      // directory match (deprecated by node): "./foo/": "./foo/.js"
+      return key;
+    }
+
+    if (key === subPath) {
+      // literal match
+      return key;
+    }
+  }
+  return null;
+}
+
+function mapSubPath(pkgJsonPath, subPath, key, value) {
+  if (typeof value === 'string') {
+    if (typeof key === 'string' && key.endsWith('*')) {
+      // star match: "./foo/*": "./foo/*.js"
+      const keyWithoutStar = key.substring(0, key.length - 1);
+      const subPathAfterKey = subPath.substring(keyWithoutStar.length);
+      return value.replace(/\*/g, subPathAfterKey);
+    }
+
+    if (value.endsWith('/')) {
+      // directory match (deprecated by node): "./foo/": "./foo/.js"
+      return `${value}${subPath.substring(key.length)}`;
+    }
+
+    // mapping is a string, for example { "./foo": "./dist/foo.js" }
+    return value;
+  }
+
+  if (Array.isArray(value)) {
+    // mapping is an array with fallbacks, for example { "./foo": ["foo:bar", "./dist/foo.js"] }
+    return value.find((v) => v.startsWith('./'));
+  }
+
+  throw pathNotFoundError(subPath, pkgJsonPath);
+}
+
+function findEntrypoint(pkgJsonPath, subPath, exportMap, conditions, key) {
+  if (typeof exportMap !== 'object') {
+    return mapSubPath(pkgJsonPath, subPath, key, exportMap);
+  }
+
+  // iterate conditions recursively, find the first that matches all conditions
+  for (const [condition, subExportMap] of Object.entries(exportMap)) {
+    if (conditions.includes(condition)) {
+      const mappedSubPath = findEntrypoint(pkgJsonPath, subPath, subExportMap, conditions, key);
+      if (mappedSubPath) {
+        return mappedSubPath;
+      }
+    }
+  }
+  throw pathNotFoundError(subPath, pkgJsonPath);
+}
+
+export function findEntrypointTopLevel(pkgJsonPath, subPath, exportMap, conditions) {
+  if (typeof exportMap !== 'object') {
+    // the export map shorthand, for example { exports: "./index.js" }
+    if (subPath !== '.') {
+      // shorthand only supports a main entrypoint
+      throw pathNotFoundError(subPath, pkgJsonPath);
+    }
+    return mapSubPath(pkgJsonPath, subPath, null, exportMap);
+  }
+
+  // export map is an object, the top level can be either conditions or sub path mappings
+  const keys = Object.keys(exportMap);
+  const isConditions = keys.every((k) => !k.startsWith('.'));
+  const isMappings = keys.every((k) => k.startsWith('.'));
+
+  if (!isConditions && !isMappings) {
+    throw new Error(
+      `Invalid package config ${pkgJsonPath}, "exports" cannot contain some keys starting with '.'` +
+        ' and some not. The exports object must either be an object of package subpath keys or an object of main entry' +
+        ' condition name keys only.'
+    );
+  }
+
+  let key = null;
+  let exportMapForSubPath;
+
+  if (isConditions) {
+    // top level is conditions, for example { "import": ..., "require": ..., "module": ... }
+    if (subPath !== '.') {
+      // package with top level conditions means it only supports a main entrypoint
+      throw pathNotFoundError(subPath, pkgJsonPath);
+    }
+    exportMapForSubPath = exportMap;
+  } else {
+    // top level is sub path mappings, for example { ".": ..., "./foo": ..., "./bar": ... }
+    key = findExportKeyMatch(exportMap, subPath);
+    if (!key) {
+      throw pathNotFoundError(subPath, pkgJsonPath);
+    }
+    exportMapForSubPath = exportMap[key];
+  }
+
+  return findEntrypoint(pkgJsonPath, subPath, exportMapForSubPath, conditions, key);
+}
+
+async function resolveId(importPath, options, exportConditions, warn) {
+  const pkgName = getPackageName(importPath);
+  if (pkgName) {
+    let pkgJsonPath;
+    let pkgJson;
+    try {
+      pkgJsonPath = await resolveImportPath(`${pkgName}/package.json`, options);
+      pkgJson = JSON.parse(await readFile(pkgJsonPath, 'utf-8'));
+    } catch (_) {
+      // if there is no package.json we defer to regular resolve behavior
+    }
+
+    if (pkgJsonPath && pkgJson && pkgJson.exports) {
+      try {
+        const packageSubPath =
+          pkgName === importPath ? '.' : `.${importPath.substring(pkgName.length)}`;
+        const mappedSubPath = findEntrypointTopLevel(
+          pkgJsonPath,
+          packageSubPath,
+          pkgJson.exports,
+          exportConditions
+        );
+        const pkgDir = path.dirname(pkgJsonPath);
+        return path.join(pkgDir, mappedSubPath);
+      } catch (error) {
+        warn(error);
+        return null;
+      }
+    }
+  }
+
+  return resolveImportPath(importPath, options);
+}
+
+// Resolve module specifiers in order. Promise resolves to the first module that resolves
+// successfully, or the error that resulted from the last attempted module resolution.
+export function resolveImportSpecifiers(
+  importSpecifierList,
+  resolveOptions,
+  exportConditions,
+  warn
+) {
+  let promise = Promise.resolve();
+
+  for (let i = 0; i < importSpecifierList.length; i++) {
+    // eslint-disable-next-line no-loop-func
+    promise = promise.then(async (value) => {
+      // if we've already resolved to something, just return it.
+      if (value) {
+        return value;
+      }
+
+      let result = await resolveId(importSpecifierList[i], resolveOptions, exportConditions, warn);
+      if (!resolveOptions.preserveSymlinks) {
+        if (await exists(result)) {
+          result = await realpath(result);
+        }
+      }
+      return result;
+    });
+
+    // swallow MODULE_NOT_FOUND errors
+    promise = promise.catch((error) => {
+      if (error.code !== 'MODULE_NOT_FOUND') {
+        throw error;
+      }
+    });
+  }
+
+  return promise;
+}
--- a/packages/node-resolve/test/browser.js
+++ b/packages/node-resolve/test/browser.js
@@ -190,7 +190,7 @@ test('supports `false` in browser field', async (t) => {
   await testBundle(t, bundle);
 });
 
-test.only('pkg.browser with mapping to prevent bundle by specifying a value of false', async (t) => {
+test('pkg.browser with mapping to prevent bundle by specifying a value of false', async (t) => {
   const bundle = await rollup({
     input: 'browser-object-with-false.js',
     plugins: [nodeResolve({ browser: true }), commonjs()]
--- a/packages/node-resolve/test/browser.js
+++ b/packages/node-resolve/test/browser.js
@@ -190,7 +190,7 @@ test('supports `false` in browser field', async (t) => {
   await testBundle(t, bundle);
 });
 
-test('pkg.browser with mapping to prevent bundle by specifying a value of false', async (t) => {
+test.only('pkg.browser with mapping to prevent bundle by specifying a value of false', async (t) => {
   const bundle = await rollup({
     input: 'browser-object-with-false.js',
     plugins: [nodeResolve({ browser: true }), commonjs()]
--- a/packages/node-resolve/types/index.d.ts
+++ b/packages/node-resolve/types/index.d.ts
@@ -10,6 +10,17 @@ export const DEFAULTS: {
 
 export interface RollupNodeResolveOptions {
   /**
+   * Additional conditions of the package.json exports field to match when resolving modules.
+   * By default, this plugin looks for the `'default', 'module', 'import']` conditions when resolving imports.
+   *
+   * When using `@rollup/plugin-commonjs` v16 or higher, this plugin will use the
+   * `['default', 'module', 'import']` conditions when resolving require statements.
+   *
+   * Setting this option will add extra conditions on top of the default conditions.
+   * See https://nodejs.org/api/packages.html#packages_conditional_exports for more information.
+   */
+  exportConditions?: string[];
+  /**
    * If `true`, instructs the plugin to use the `"browser"` property in `package.json`
    * files to specify alternative files to load for bundling. This is useful when
    * bundling for a browser environment. Alternatively, a value of `'browser'` can be
--- a/packages/node-resolve/package.json
+++ b/packages/node-resolve/package.json
@@ -65,7 +65,7 @@
     "@babel/core": "^7.10.5",
     "@babel/plugin-transform-typescript": "^7.10.5",
     "@rollup/plugin-babel": "^5.1.0",
-    "@rollup/plugin-commonjs": "^14.0.0",
+    "@rollup/plugin-commonjs": "^16.0.0",
     "@rollup/plugin-json": "^4.1.0",
     "es5-ext": "^0.10.53",
     "rollup": "^2.23.0",
--- a/packages/node-resolve/src/util.js
+++ b/packages/node-resolve/src/util.js
@@ -1,13 +1,8 @@
 import { dirname, extname, resolve } from 'path';
-import { promisify } from 'util';
 
 import { createFilter } from '@rollup/pluginutils';
 
-import resolveModule from 'resolve';
-
-import { exists, realpath, realpathSync } from './fs';
-
-const resolveId = promisify(resolveModule);
+import { realpathSync } from './fs';
 
 // returns the imported package name for bare module imports
 export function getPackageName(id) {
@@ -157,37 +152,4 @@ export function normalizeInput(input) {
 
   // otherwise it's a string
   return [input];
-}
-
-// Resolve module specifiers in order. Promise resolves to the first module that resolves
-// successfully, or the error that resulted from the last attempted module resolution.
-export function resolveImportSpecifiers(importSpecifierList, resolveOptions) {
-  let promise = Promise.resolve();
-
-  for (let i = 0; i < importSpecifierList.length; i++) {
-    // eslint-disable-next-line no-loop-func
-    promise = promise.then(async (value) => {
-      // if we've already resolved to something, just return it.
-      if (value) {
-        return value;
-      }
-
-      let result = await resolveId(importSpecifierList[i], resolveOptions);
-      if (!resolveOptions.preserveSymlinks) {
-        if (await exists(result)) {
-          result = await realpath(result);
-        }
-      }
-      return result;
-    });
-
-    // swallow MODULE_NOT_FOUND errors
-    promise = promise.catch((error) => {
-      if (error.code !== 'MODULE_NOT_FOUND') {
-        throw error;
-      }
-    });
-  }
-
-  return promise;
 }
--- a/packages/node-resolve/test/fixtures/browser-object-with-false.js
+++ b/packages/node-resolve/test/fixtures/browser-object-with-false.js
@@ -12,7 +12,7 @@ const clientHttp = new Client('http:');
 t.is(clientWs.name, 'websocket-tracker');
 t.is(clientHttp.name, 'NULL');
 t.is(HTTPTracker, ES6_BROWSER_EMPTY);
-t.is(HTTPTrackerWithSubPath, ES6_BROWSER_EMPTY);
+t.deepEqual(HTTPTrackerWithSubPath, { default: {} });
 
 // expose
 export default 'ok';
--- a/packages/node-resolve/test/fixtures/exports-cjs.js
+++ b/packages/node-resolve/test/fixtures/exports-cjs.js
@@ -1,0 +1,3 @@
+const main = require('exports-cjs');
+
+module.exports = main;
--- a/packages/node-resolve/test/fixtures/exports-conditions-fallback.js
+++ b/packages/node-resolve/test/fixtures/exports-conditions-fallback.js
@@ -1,0 +1,3 @@
+import main from 'exports-conditions-fallback';
+
+export default main;
--- a/packages/node-resolve/test/fixtures/exports-directory.js
+++ b/packages/node-resolve/test/fixtures/exports-directory.js
@@ -1,0 +1,5 @@
+import a from 'exports-directory/foo/a.js';
+import b from 'exports-directory/foo/b.js';
+import c from 'exports-directory/foo/nested/c.js';
+
+export default { a, b, c };
--- a/packages/node-resolve/test/fixtures/exports-main-directory.js
+++ b/packages/node-resolve/test/fixtures/exports-main-directory.js
@@ -1,0 +1,5 @@
+import a from 'exports-main-directory/a.js';
+import b from 'exports-main-directory/foo/b.js';
+import c from 'exports-main-directory/foo/nested/c.js';
+
+export default { a, b, c };
--- a/packages/node-resolve/test/fixtures/exports-mappings-and-conditions.js
+++ b/packages/node-resolve/test/fixtures/exports-mappings-and-conditions.js
@@ -1,0 +1,5 @@
+import main from 'exports-mappings-and-conditions';
+import foo from 'exports-mappings-and-conditions/foo';
+import bar from 'exports-mappings-and-conditions/bar';
+
+export default { main, foo, bar };
--- a/packages/node-resolve/test/fixtures/exports-nested-conditions.js
+++ b/packages/node-resolve/test/fixtures/exports-nested-conditions.js
@@ -1,0 +1,3 @@
+import main from 'exports-nested-conditions';
+
+export default main;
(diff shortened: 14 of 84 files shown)


packages/node-resolve/src/index.js
```diff
@@ -30,6 +30,7 @@ const deepFreeze = object => {
   return object;
 };
 const defaults = {
+  exportConditions: ['module', 'default'],
```
LarsDenBakker, line 33: @lukastaegert if I understand correctly this is the approach webpack will take. this seems like a good default for rollup as well

packages/node-resolve/src/resolveId.js
```diff
@@ -0,0 +1,121 @@
...
+
+function mapSubPath(pkgJsonPath, subPath, value) {
+  if (typeof value === 'string') {
+    // mapping is a string, for example { "./foo": "./dist/foo.js" }
+    return value;
+  }
+
+  if (Array.isArray(value)) {
+    // mapping is an array with fallbacks, for example { "./foo": ["foo:bar", "./dist/foo.js"] }
+    return value.find((v) => v.startsWith('./'));
```
LarsDenBakker, line 23: @guybedford any advice how to test for valid import specifiers?

packages/node-resolve/src/index.js
lukastaegert: I would also add the `import` condition as a default because within Rollup, the assumption is that all modules are ESM. The only exception which is the commonjs plugin should receive its own logic IMO, or better, we should add an API to the node-resolve plugin that the commonjs plugin can access. But I would see this beyond this PR. Otherwise, either Rollup unnecessarily picks a default branch that is CommonJS or worse, it does not pick any branch at all when presented with the choice between `import` and `require`.

Another question: Why are you listing `default` as a condition? To my understanding it is a catch all that always applies.

LarsDenBakker: Good point about `import`, I'll add it in. 

In the way package entrypoints work, `default` is a condition like any other in a technical sense. We need to iterate the array in order and find the best match, ending up with default as the last. However, I understand the confusion since it's listed under the plugin options. We should just add it to the array in the code, a user should not be able to remove it.

LarsDenBakker:
@lukastaegert in my opinion the whole main fields thing is very confusing to most developers. Package entrypoints is a clean break, and perhaps the best chance we have right now at a standard format for describing the export structure of a package. Therefore I think it's best that if a package has specified `exports` it should take priority over other main fields.

We could add options to disable to tweak the behavior, but I'd like to avoid that if possible. You can't disable this behavior in node, making this logic consistent is important I think. Package authors know best what applied to what environment. Power users can write a plugin with `resolveId` if they *really* need to customize it.

lukastaegert:
> Power users can write a plugin with resolveId if they really need to customize it.

Makes sense. Still I think it would be sensible to add a note to the documentation of the `mainFields` option that it will be ignored for packages that have an `exports` field.

LarsDenBakker:
Yeah, good point. I've added it in the readme.

shellscape:
_A note from the sidelines to everyone participating: Please do make sure to let us know when this is ready! There are a lot of moving parts to this and several simultaneous conversations, and I want to make sure it doesn't go stale or get missed. :beers:_

LarsDenBakker:
I just realized I'm missing support for exporting entire directories:

```
{
  "exports": {
    "./lib/": "./lib/",
  }
}
```

LarsDenBakker:
It would be great to get some 👀  on this. @guybedford since you worked on this on the node side, would you be able to double check this?

packages/node-resolve/README.md
```diff
@@ -34,16 +34,23 @@ export default {
...
 ```
 
 Then call `rollup` either via the [CLI](https://www.rollupjs.org/guide/en/#command-line-reference) or the [API](https://www.rollupjs.org/guide/en/#javascript-api).
 
 ## Options
 
+### `exportConditions`
+
+Type: `Array[...String]`<br>
+Default: `['module', 'import']`
```
guybedford, line 50: Where is `"module"` defined here? I don't know exactly what its definition is or where that is supported.

packages/node-resolve/src/resolveId.js
```diff
@@ -0,0 +1,130 @@
...
+  if (typeof exportMap !== 'object') {
+    return mapSubPath(pkgJsonPath, subPath, key, exportMap);
+  }
+
+  // iterate conditions recursively, find the first that matches all conditions
+  for (const [condition, subExportMap] of Object.entries(exportMap)) {
+    if (conditions.includes(condition)) {
+      const mappedSubPath = findEntrypoint(pkgJsonPath, subPath, subExportMap, conditions, key);
+      if (mappedSubPath) {
+        return mappedSubPath;
```
guybedford, line 46: Note that condition matching will actually fall back if not found.

Eg:

```json
{
  "exports": {
    "node": {
      "require": "./cjs.js"
    },
    "default": "./main.js"
  }
}
```

will use the default in Node.js when using `import`.

packages/node-resolve/src/resolveId.js
guybedford: This works. The other things you can optionally check are that there are no `.`, `..` or `node_modules` path segments. But that level of strictness isn't completely necessary here too since Node.js will cover the full edge cases naturally as a target.

packages/node-resolve/README.md
LarsDenBakker: This is the standard field webpack and rollup will be using.

LarsDenBakker:
Thanks! I didn't see the algorithm before, I will double check if I covered everything there.

I think we can take the `imports` field in a separate PR. Definitely doesn't make resolving imports simpler though :(

packages/node-resolve/README.md
guybedford: I understand it is a useful way to enable CommonJS modules to load ES modules to avoid duplication cases (although these pains are yet to be felt as there dont seem to be that many people shipping dual mode packages that hit these cases yet).

But I'm not sold that it is necessarily the best way to do this, not that I can't be :)

I just don't think RollupJS should ship this without the meaning being very clearly defined for the various use cases.

Let me just try to get the definition straight then at least:

1. Does `"module"` imply modern syntax, or is that not part of its definition?
2. Does `"module"` imply resolution in a bundler like Webpack or Rollup only? Or is it expected that the definition extends to other bundlers or tools?
3. Does `"module"` apply for the Node.js platform target? This can be seen then as something that changes the default Node.js semantics under bundling which breaks the overall goal of RollupJS building for Node as retaining semantics.
4. Does `"module"` apply for any other platforms other than the browser?
5. Finally, do you think Node.js itself should implement the `"module"` field when resolving `import` (and then just not when resolving `require`)? This seems like it might avoid the repetition of defining `"exports": { "import": "./main.mjs", "module": "./main.mjs", "require": "./main.cjs" }` current redundancy for these sorts of packages.

guybedford:
@LarsDenBakker I was only pointing you to it for the long-term roadmap. It's a fringe feature only needed in esoteric use cases like future builtin polyfills.

guybedford:
@LarsDenBakker hope my feedback wasn't stalling here, this would be very nice to merge.

packages/node-resolve/README.md
LarsDenBakker: `import` can point to es module which wrap commonjs modules. This is a problem for pure es module workflows, so there needs to be a way to indicate pure es modules. The name `module` came out of the discussion at https://github.com/rollup/plugins/issues/362 and https://github.com/webpack/webpack/issues/11014#issuecomment-643256943

I'm fine with deferring that to a later PR, but it would be a breaking change later on again. @lukastaegert what do you think?

packages/node-resolve/src/resolveId.js
LarsDenBakker: I'll keep this as is for now then.

packages/node-resolve/README.md
guybedford: Thanks, I'm not against it, I just want to be clear on the meaning. Ok, it seems the key source here is Tobias's proposal. Reading his proposal more clearly, let me try and answer my own questions then:

> Does "module" imply modern syntax, or is that not part of its definition?

Not mentioned in the doc

> Does "module" imply resolution in a bundler like Webpack or Rollup only? Or is it expected that the definition extends to other bundlers or tools?

Not mentioned in the doc.

> Does "module" apply for the Node.js platform target? This can be seen then as something that changes the default Node.js semantics under bundling which breaks the overall goal of RollupJS building for Node as retaining semantics.

The doc explicitly mentions that you would want to use it when bundling for a Node.js target so it doesn't seem to be Node or browser specific at all as far as I can tell.

> Does "module" apply for any other platforms other than the browser?

As above, it can apply to Node.js

> Finally, do you think Node.js itself should implement the "module" field when resolving import (and then just not when resolving require)? This seems like it might avoid the repetition of defining "exports": { "import": "./main.mjs", "module": "./main.mjs", "require": "./main.cjs" } current redundancy for these sorts of packages.

Node.js likely would not do this.

Basically it seems that "module" is specifically there to solve the bundler problem for when CJS is able to import from ESM in bundlers only. And RollupJS as a bundler can join Webpack in doing that.

These constraints seem clear to me, I hope we can give users a little guidance on when to use "module" mainly.

@lukastaegert will entirely defer to you on this one, I'm happy either way, just erring on the side of caution in introducing new things.

packages/node-resolve/src/resolveId.js
LarsDenBakker: We handle this case, I added a test to confirm :)

LarsDenBakker:
@guybedford no worries, I didn't have the time yet to look at your comments :) I've updated the PR now and rebased on master.

packages/node-resolve/README.md
lukastaegert: As you correctly stated, this condition was specifically created for the bundler use case and I definitely think we should add it by default. An entirely different question is when library authors should use it. Here, I would see it as a last resort solution to solve e.g. certain dual state hazard situations in a purely ESM way, so there are other patterns I would recommend first.

guybedford:
Thanks @LarsDenBakker, sounds like this is almost good to go and handles the cases well.

There is one more case I think we should alter here - and that is that the `"import"` condition should not match for `require()` resolution if we are implementing the `"module"` condition. The `"import"` condition is strictly defined to mean importing from ESM.

It would be interesting to hear from @sokra if Webpack have explicitly followed this path or not as well. Perhaps bundlers will decide to change the semantic, but at least agree on changing it then!

LarsDenBakker:
I don't think the node resolve has information about whether the module was references from an import or a require. 

But even if it's possible, wouldn't that potentially lead to both the esm and the cjs variant being included in the bundle if it was referenced from both a require and an import?

guybedford:
> I don't think the node resolve has information about whether the module was references from an import or a require.

Yes it does, that is the definition of these conditions. For interest, this was the original use case that allowed us to ship conditional exports at all as Node.js otherwise wouldn't have had need to.

> But even if it's possible, wouldn't that potentially lead to both the esm and the cjs variant being included in the bundle if it was referenced from both a require and an import?

This is why the `"module"` condition was proposed by @sokra, and if we are doubling down on this value proposition that is why I say correctly implementing the `"import"` and `"require"` condition semantics seems important to me, but I'd still be interested to hear from @sokra what Webpack does here too.

LarsDenBakker:
I mean I'm not sure whether the rollup node resolve plugin has access to this information.

guybedford:
Yes it requires more architecture work before this can be merged, if it is deemed important. It's effectively a third parameter to the resolver - `isCommonJSImport` or similar.

lukastaegert:
As I stated above, it is impossible for a non-ES module to be the importer unless plugins are involved. Therefore I strongly vote for keeping the `import` condition by default BUT in a later step, solve `require` resolution entirely from within the commonjs plugin. This plugin is already manually doing its own resolution via `this.resolve`. So at that point, we could implement a mechanism where the commonjs plugin can pass along some resolution information to the node-resolve plugin via a parameter to `this.resolve` and thus `resolveId`. But I would see this is a separate enhancement later unless you think it is impossible to land this without `require` resolution from the start.

lukastaegert:
Of course this would require adding support for parameters in Rollup core first. When thinking about how to pass a parameter and avoid conflicts, I am currently thinking of allowing an options parameter where the keys are the receiving plugin names. Then for `node-resolve`, we could pass again an object that for now contains a single `exportConditions` parameter where you can pass an alternative set of conditions, i.e.

```
{
  'node-resolve': {
    exportConditions: ['require']
  }
}
```

This would potentially allow many other interesting use-cases. The question is how this should interact with the existing `exportConditions`. My feeling is that it should completely replace them, which would mean that the commonjs plugin would need its own e.g. `requireExportConditions` parameter that would define what conditions to use when requiring anything from commonjs so that you can specify custom conditions here as well.

guybedford:
I like the gist of that require condition approach, and I tend to think it would be preferable to land the full functionality from the start to avoid possible breaks.

I think it is possible without arbitrary parameter passing though - if the CommonJS resolver can look at the parent importer and tell if that was a CommonJS or ES module initially and apply the "require" condition as needed then?

Note that conditions specifically are meant to be environment constants - that is they are static for the whole build. It is only these require and import conditions that are unique in changing depending on what you resolve from.

lukastaegert:
>  if the CommonJS resolver can look at the parent importer and tell if that was a CommonJS or ES module initially and apply the "require" condition as needed then?

There is no actual CommonJS resolver, the call to `this.resolve` is still going to the node-resolve plugin. I do not see an alternative without a parameter as
- the commonjs plugin is currently responsible for maintaining the information which module is commonjs
- the node-resolve plugin is responsible for handling package.json files and resolving via the `exports` conditions

So information needs to flow between the plugins one way or another.

guybedford:
@lukastaegert ah of course. Yes cross-plugin metadata for modules is an important loader problem, all the solutions are quite complex though.

Perhaps it's time to consider merging the CommonJS and Node.js resolve plugins? That would avoid any cost here at all.

Alternatively sharing a simple object during the build like `this.metadata.moduleFormatCache = moduleFormatCache` where all plugins share that `this.metadata` object can be a simple mechanism.

Everything else gets into plugin keying / message passing which is quite a complexity cliff, but not out of the question too.

lukastaegert:
Another way of handling the parameters could be to not list the full set of conditions but to list what should change, e.g.

```
{
  'node-resolve': {
    addedExportConditions: ['require'],
    removedExportConditions: ['import', 'module']
  }
}
```

or in a more concise way

```
{
  'node-resolve': {
    updateExportConditions: ['require', '!import', '!module']
  }
}
```

but this would establish a new pattern of "negated" conditions.

guybedford:
@lukastaegert the "require" and "import" conditions ideally would not be possible to disable - they should be the core implementation along with "default" since they are well-defined. Node.js does this for the `conditions` argument in loaders too. I would prefer `customConditions: ['module']` to be the default, where users overriding it need to reinclude the "module" condition to indicate keeping it.

lukastaegert:
> Perhaps it's time to consider merging the CommonJS and Node.js resolve plugins? That would avoid any cost here at all.

Not a fan for various reasons:
- especially the CommonJS plugin is a complicated mess of functionality, I feel better knowing it can be thrown away at any time if we have something better
- I know there is at least [one alternative commonjs plugin](https://github.com/eight04/rollup-plugin-cjs-es), I think it is good that people can choose if they want
- CommonJS is a format that is defined outside the Node context and it should be possible to use it outside the Node context
- The CommonJS plugin makes the whole build noticeably slower due to all the additional parsing necessary, so if I only need the node resolution, then I do not want to be weighed down with the commonjs part
- node-resolve is definitely used without commonjs e.g. by [es-dev-server](https://open-wc.org/developing/es-dev-server.html)
- on the other hand, commonjs is sometimes used separately for file-by-file CommonJS to ESM conversion

lukastaegert:
>  the "require" and "import" conditions ideally would not be possible to disable

You mean the logic that adds these conditions. I do not fully agree here. While the exports logic was defined for the Node ecosystem, my understanding is that it was explicitly designed to support non-Node use-cases as well. It is very conceivable to me that at some point there may be situations where you want neither condition to be applied. For that situation, it should be possible to have full control and not have to wrangle some imposed defaults.

guybedford:
@lukastaegert see https://nodejs.org/dist/latest-v14.x/docs/api/esm.html#esm_conditional_exports and particularly the definitions:

> "import" - matched when the package is loaded via import or import(). Can reference either an ES module or CommonJS file, as both import and import() can load either ES module or CommonJS sources. Always matched when the "require" condition is not matched.
> "require" - matched when the package is loaded via require(). As require() only supports CommonJS, the referenced file must be CommonJS. Always matched when the "import" condition is not matched.

One or the other is defined to always match by these definitions.

lukastaegert:
I understand these things but this was not what I challenged. I want to say that there are other consumers outside the Node ecosystem that may load a package via neither import nor require. Loading via a non-module script tag could be such a use case. This is neither an import nor a require and would indeed warrant a third option: A wrapped IIFE that creates global variables. And there is more possible: Defining special entry points for imports from other module formats such as AMD or SystemJS. Saying that import and require are for all times a binary system may apply to the Node ecosystem but feels short-sighted to me for the JavaScript ecosystem.

guybedford:
@lukastaegert additional conditions can be defined certainly, but they should be matched then in addition to these and not based on choosing not to match these conditions. Multiple conditions are a first-class feature with matching order and composition clearly defined. Arbitrarily deciding not to match a condition based on its defined meaning (or making this easy to do via a configuration error) is exactly the sort of problem we want to avoid though.

lukastaegert:
Ok, so if I want to have a condition for AMD imports, I must expect it to be used together with either `import` or `require` even if it does not make the slightest sense?

sokra:
Here is how we did things in webpack

> I want to have a condition for AMD imports

in weback AMD will also use `require` condition.

> This is why the `"module"` condition was proposed by @sokra, and if we are doubling down on this value proposition that is why I say correctly implementing the `"import"` and `"require"` condition semantics seems important to me, but I'd still be interested to hear from @sokra what Webpack does here too.

The `module` condition is an addon to `import` and `require`. It will always go together with `import` or `require`. When the `module` condition is supported it should go with both of them. As package author the `module` condition must come before `import` and `require` in the exports field either it won't have an effect (object order).

> and that is that the `"import"` condition should not match for `require()` resolution if we are implementing the `"module"` condition. The `"import"` condition is strictly defined to mean importing from ESM.

Yes that's true. `require()` will only use `module + require` conditions.

> It is very conceivable to me that at some point there may be situations where you want neither condition to be applied.

In webpack when using `@import` in css or sass, neither `import` or `require` will be used but `style` resp. `sass`.

---

More details here: https://gist.github.com/sokra/e032a0f17c1721c71cfced6f14516c62#conditions

lukastaegert:
@sokra Thanks for the input, that is very helpful especially with your usage of `module`. So this means we should definitely keep the `module` condition for `require`, which on second thought very much makes sense as this is what the condition was made for 🙄.

@guybedford With regard to my proposal https://github.com/rollup/plugins/pull/540#issuecomment-691811382 you mean to say that removing the `import` condition should happen automatically when I add the `require` condition? I still see benefit in being able to deactivate conditions, though. Or we allow two versions:
- One version that just adds conditions to the existing ones, and there is logic in the node-resolve plugin that handles the binary nature of import/require
- One version that completely replaces all conditions, e.g. to allow plugins to handle the style use-case. For this version, it would be hard to inspect the existing conditions, though

Or we allow passing a function here that is called by node-resolve with the existing conditions and returns a new set of conditions. Maybe this is the most flexible version.

guybedford:
@lukastaegert I'm saying explicitly that like `"default"`, `"import"` and `"require"` are system conditions that should always activate appropriately and not be possible to change. It is simply a binary state in the resolver that can be a third argument carried through or simply contextual for the CJS resolver based on the metadata lookup of whether the parent is CJS.

I'm strongly against any user configurability for these conditions.

guybedford:
> In webpack when using @import in css or sass, neither import or require will be used but style resp. sass.

If one considers this as a sort of resolver call like `import.meta.resolve` the default of it being `"import"` should still apply IMO.

guybedford:
Actually thinking about this further, not setting the `"style"` condition will cause issues in that case.

Perhaps `import.meta.resolve` should be a neither case after all - this way instead of using `"style"` one could use `"import"` and `"default"` to allow a generic asset resolver to work without needing a custom condition (since custom conditions cannot and will never be able to be piped into `import.meta.resolve`).

Regardless, `"import"` and `"require"` being core conditions managed and controlled by the resolver itself is the important property here.

LarsDenBakker:
Would you guys be OK with getting these changes in without handling the `require` clause, and then we implement handling for `require` in a later PR? Or do this entire feature behind a flag?

I'm vey keen on landing this just for being able to define the exports of a package, and preventing imports of unspecified exports. I'd like to start promoting this pattern for web based projects, and we want to start using this inside our company as well. Alternatively we might use a fork temporarily.

ChocolateLoverRaj:
@LarsDenBakker I hope this pull request gets merged. Currently, I'm trying to import `@rollup/plugin-babel`, but there is a problem with the exports. Since the CommonJS file is getting exported, there are no default exports. For example:
```js
import babel from '@rollup/plugin-babel'
```
This isn't working as expected because I'm getting
```js
{
  babel: [Function (anonymous)],
  createBabelInputPluginFactory: [Function: createBabelInputPluginFactory],
  createBabelOutputPluginFactory: [Function: createBabelOutputPluginFactory],
  default: [Function (anonymous)],
  getBabelInputPlugin: [Function (anonymous)],
  getBabelOutputPlugin: [Function (anonymous)]
}
```
I tried explicitly importing from the `dist/index.es.js` file, but that isn't okay with node.js.
```js
import babel from '@rollup/plugin-babel/dist/index.es.js'
```
Then I got the error:
```
import * as babel from '@babel/core';
^^^^^^

SyntaxError: Cannot use import statement outside a module
```
I think node.js is not allowing the `dist/index.es.js` file to use `import`, because it is implicitly using
```json
{ "type": "commonjs" }
```
from the top level `package.json`, which doesn't have a `type` field.

Will this pull request fix the problem I have?

lukastaegert:
> Or to merge the PR as is behind a flag

I like this idea. How about an option `experimentalPackageExports: boolean` that makes clear that this feature is incomplete at the moment? I think there should be nothing to argue against given the name and the fact that it is off be default. In the meantime, we can fix `require` handling before releasing this feature properly without the flag.

lukastaegert:
In the meantime, I will start working on PRs for the message passing infrastructure in Rollup Core and the CommonJS plugin. For the message format, maybe we can go for

```
{
  'node-resolve': {
    updateExportConditions(configuredConditions: string[]): string[] {
      return configuredConditions.filter(condition => condition !== 'import').concat(['require']);
    }
  }
}
```

and then add a validation in `node-resolve` that throws if
- there is no `default` condition (or we always handle this one behind the scenes?)
- there is neither a `require` nor `import` condition
- there is both the `require` and `import` condition (or is this a valid situation?)

I think having a validation makes it easier to have a simple interface.

guybedford:
Ideally the information sharing should just be the format map not the
conditions array, since that is information only the resolve plugin needs.
Does ‘this.parentMeta = map’ work already as an approach? That seems like
it should be relatively straightforward without requiring any new
architecture.

Also, if implementation really is the bottleneck here I can put it on my
list for the next couple of weeks. I would prefer not to ship an incorrect
implementation even flagged.

lukastaegert:
> since that is information only the resolve plugin needs

It would be nice to build a feature that is also usable for other things like resolving script tag/AMD/Electron/style imports or anything that comes to mind. I do not think node-resolve should know about formats at all, especially since an `import()` from commonjs should be using the `import` condition. And if we want AMD to use the `require` condition then an AMD plugin needs to pretend files are commonjs format? This feels to me like it is making stuff complicated and error prone, or at least hard to understand.

No, I am pretty sure that the most future proof way to transfer this information is to pass it as a flag to `this.resolve` when triggering the resolution from the plugin as this approach is very adaptable and would allow other plugins to handle specific import types intelligently as well. That way, all knowledge about CommonJS remains with the commonjs plugin.

Any comments about the validation suggestion?

> ‘this.parentMeta = map’ 

I have not a clue what you are talking about.

guybedford:
@lukastaegert top-level resolve metadata makes sense certainly. 

The concern with the API you proposed is that if we open up conditions too much we risk having them change between resolve calls when they are supposed to be environment constants. Only the import/require is an exception in varying within the same environment. Handling this stuff in a way that doesn't expose users to footguns that change the semantics is quite important I think.

I'd be more than glad to review core APIs here further. My suggestion of parent metadata was a way to handle the switch internally based on checking the module format of the `parentId`. The check is as simple as:

```js
function getConditions (userConditions: string[], parentId: string) {
 return ['default', moduleFormatMap[parentId] === 'commonjs' ? 'require' : 'import', ...userConditions];
}
```

The module format map is the missing data that needs to be piped. Permitting conditions as more generic resolver metadata could be an option, but again I'm not clear how that could be done without creating an API that doubles as a footgun since conditions should really be set once as an input option and then stay constant for that build.

lukastaegert:
One of the things I am trying to tell you is that even within the same Node CommonJS importer, the format is NOT the determining factor!

```js
const foo = require('foo');
console.log(foo);
import('foo').then(console.log);
```

Is a valid CommonJS file in Node, but the first import needs to be resolved with the `require` condition and the second dynamic imports needs to be resolved with the `import` condition. Now of course this can be solved by having node-resolve catch all dynamic imports separately, which may or may not be desireable.

But what about an Electron build where the frontend part needs to follow `browser` semantics while the backend part needs to follow `node` semantics? I think it would be very interesting to be able to produce them in the same build to have shared code between main and render process, though of course the details here would need to be worked out i.e. to decide which semantics a certain module should follow and to make sure the whole build works.

And what about imports from style sheets that I have as assets in my build? It would be nice as well to have special conditions here in the same build.

This is not a common user-facing API, it is a plugin-facing API. We can also limit this API to be able to just add conditions or switch `import` with `require` if this helps to mollify your concerns.

Another concern I have here is architecturally as it is already clear as to your suggestions elsewhere that at some point, the commonjs plugin needs to depend on the node-resolve plugin to be able to correctly infer file types using Node's algorithm instead of the current sniffing algorithm. Having at the same time a dependency the other way feels very problematic to me e.g. with regard to version management.

So another suggestion:

```
{
  'node-resolve': {
    useRequire: true,
    additionalConditions: ['browser']
  }
}
```

lukastaegert:
So what I am also saying is that while a running Node process is homogeneous, a Rollup build does not have to be in that it can contain various more-or-less independent fragments such as assets, various process types that share dependencies, you name it.

guybedford:
@lukastaegert ok, how about we define that it is ok to match _neither_ `"import"` nor "require"` when using `import.meta.resolve` for effectively "asset resolutions"? Tools like RollupJS could apply their own "style" conditions or whatever but Node.js would just say that it is allowed to match neither in these cases. How does that sound as an approach?

I can try follow up with an agenda item for this meeting to see if we can get some agreement here on these definitions. The alternative to me seems to be that `"import"` just matches a style - because you are still **import**ing it.

guybedford:
@lukastaegert posted https://github.com/nodejs/modules/issues/556.

lukastaegert:
Thanks @guybedford for posting these questions to a wider audience because I think some additional feedback from the Node team would go a long way here, also in evolving the concepts.

guybedford:
@lukastaegert if we relax the condition I'm fine with allowing users to provide it but the key requirements are that:

1. The default implementation should respect the definitions of the "require" and "import" condition definitions as used in Node.js.
2. These conditions should remain mutually exclusive, via a validation step or otherwise.
3. If providing custom user conditions, the above defaults for "require" and "import" should still be respected along with those custom conditions, unless the user wants to specifically disable them.

guybedford:
Perhaps RollupJS core could allow metadata to be associated with individual modules, that can be set as a return value of the `resolveId` function (eg `return { id: ..., meta: { format: 'cjs' } }`). This would solve the information problem the CommonJS plugin we already have in needing to maintain side metadata for the format. This metadata object could then be shared with the other hooks, either by an argument, or if that is overcomplicating the existing function definitions, via some `this.getMetadata(parentId)` API.

guybedford:
What's nice about this metadata proposal is that it can integrate with the cache system if metadata is assumed to have the same invalidation properties as the module itself.

lukastaegert:
I think the meta-data id is a really good idea for transferring back information e.g. once we allow relying on the node algorithm to detect file formats. After all, the node-resolve plugin is the sole maintainer of package.json information. I do not think the current sniffing approach for format detection will go away entirely for quite some time, though, because many users will likely not yet have their ES modules flagged as such properly, and we will certainly need it for virtual modules, but I am totally open to make this an opt-in feature rather than the default in a future major version.

guybedford:
@lukastaegert there is nothing wrong with sticking with sniffing, but note that the metadata v sniffing format detection question is completely orthogonal to the "import" / "require" resolution which is about the module format of the parent importer / importer context _not_ about the module format of the target file. It is perfectly valid for `"import"` to point to a `.wasm` file. Similarly it is absolutely fine for `require` to point to an ES module - rather it would throw on execute / instantiate not on resolve.

LarsDenBakker:
I updated this branch to handle the new require marker. However it depends on a release of the commonjs plugin, so the test is currently failing. Will check again after a new release of the commonjs plugin is made.

packages/node-resolve/src/index.js
```diff
@@ -29,6 +29,10 @@ const deepFreeze = (object) => {
 
   return object;
 };
+
+const baseConditions = ['default', 'module']
```
LarsDenBakker, line 33: @lukastaegert should we match `module` also in the case of require? Or only when using import?

lukastaegert: Good question. I know @sokra has written something up how they handle the module condition in webpack but I cannot seem to find the document 🙄. But to my understanding, the main point of the `module` condition is to denote that it is "safe" to use this ES module to resolve a `require` statement, so yes, I'd say the condition should be added.

LarsDenBakker:
@lukastaegert rebased on master after the release of the commonjs plugin. looks like we're good to go now :)

packages/node-resolve/src/resolveId.js
```diff
@@ -0,0 +1,130 @@
...
+const pathNotFoundError = (subPath, pkgPath) =>
+  new Error(`Package subpath '${subPath}' is not defined by "exports" in ${pkgPath}`);
+
+function findExportKeyMatch(keys, subPath) {
+  return keys.find((key) => (key.endsWith('/') ? subPath.startsWith(key) : key === subPath));
+}
+
+function mapSubPath(pkgJsonPath, subPath, key, value) {
+  if (typeof value === 'string') {
+    if (value.endsWith('/')) {
```
guybedford, line 21: Note that subpath patterns (`endsWith('/*')` here) are now the preferred form over the trailing `/`. See https://nodejs.org/dist/latest-v15.x/docs/api/packages.html#packages_subpath_patterns.

The trailing `/` is actually being deprecated in https://github.com/nodejs/node/pull/35747.

No stress getting this into this PR right now, but more just letting you know as I find it much nicer to work with and feedback is welcome further.

packages/node-resolve/src/index.js
```diff
@@ -222,7 +234,15 @@ export function nodeResolve(opts = {}) {
       importSpecifierList.push(importee);
       resolveOptions = Object.assign(resolveOptions, customResolveOptions);
 
-      let resolved = await resolveImportSpecifiers(importSpecifierList, resolveOptions);
+      const warn = (...args) => this.warn(...args);
+      const isRequire = opts && opts.custom && opts.custom['node-resolve'] && opts.custom['node-resolve'].isRequire;
```
guybedford, line 238: Should we warn when not using the RollupJS version that supports this?

packages/node-resolve/src/index.js
```diff
@@ -222,7 +234,15 @@ export function nodeResolve(opts = {}) {
...
+      const exportConditions = isRequire ? conditionsCjs : conditionsEsm;
```
guybedford, line 239: Perfect!

guybedford (approved):
Great work and nice to see the collaboration here in getting this through! Thank you both so much for taking the onerous feedback on board!

packages/node-resolve/src/index.js
LarsDenBakker: We actually need to check the version of the node-resolve plugin, which I think isn't even possible right now.

LarsDenBakker: Sorry I mean we need to check the version of the commonjs plugin.

lukastaegert (commented):
Still testing this, but looks really good. The only thing I am missing are Readme updates:
- document the `exportConditions` option (possibly with a link to the Node documentation so people can read up on this if they do not know it)
- we could extend the short section about the commonjs plugin to note that resolving require conditions requires at least version 16.0.0 of the commonjs plugin
- add a new section that documents the interface the commonjs plugin is using, i.e. the custom `isRequire` condition that can be passed to `this.resolve` to replace the `import` condition with `require`.

packages/node-resolve/src/resolveId.js
```diff
@@ -0,0 +1,130 @@
+import fs from 'fs';
+import path from 'path';
+import { promisify } from 'util';
+
+import resolve from 'resolve';
+
+import { getPackageName } from './util';
```
lukastaegert, line 7: There is a cyclic dependency between this file and the `util` file. While this does not have to be an issue, cyclic dependencies tend to cause confusing errors down the road and I would prefer to avoid them. My suggestion would be to extract `resolveImportSpecifiers` from the `util` file and either move it into its own file or into this one.

lukastaegert:
I just went through https://nodejs.org/api/packages.html#packages_package_entry_points. So it appears nearly everything is working awesomely except pattern conditions like `"./foo/*": "./bar/*.js"`.

Not sure if we want to add them here but they seem very useful.

LarsDenBakker:
The star pattern was added to node after this PR was opened :) I added it now.

I updated the documentation as well.

packages/node-resolve/src/resolveId.js
LarsDenBakker: Removed the circular dependency.

packages/node-resolve/src/resolveId.js
LarsDenBakker: Added this

packages/node-resolve/README.md
```diff
@@ -34,16 +34,27 @@ export default {
...
 Then call `rollup` either via the [CLI](https://www.rollupjs.org/guide/en/#command-line-reference) or the [API](https://www.rollupjs.org/guide/en/#javascript-api).
 
 ## Options
 
+### `exportConditions`
+
+Type: `Array[...String]`<br>
+Default: `[]`
+
+Additional conditions of the package.json exports field to match when resolving modules. By default, this plugin looks for the `'default', 'module', 'import']` conditions when resolving imports.
```
guybedford, line 52:
```suggestion
Additional conditions of the package.json exports field to match when resolving modules. By default, this plugin looks for the `['default', 'module', 'import']` conditions when resolving imports.
```

packages/node-resolve/README.md
```diff
@@ -34,16 +34,27 @@ export default {
...
+
+When using `@rollup/plugin-commonjs` v16 or higher, this plugin will use the `['default', 'module', 'import']` conditions when resolving require statements.
```
guybedford, line 54:
```suggestion
When using `@rollup/plugin-commonjs` v16 or higher, this plugin will use the `['default', 'module', 'require']` conditions when resolving require statements.
```

packages/node-resolve/src/resolveImportSpecifiers.js
```diff
@@ -0,0 +1,197 @@
...
+
+const resolveImportPath = promisify(resolve);
+const readFile = promisify(fs.readFile);
+
+const pathNotFoundError = (subPath, pkgPath) =>
+  new Error(`Package subpath '${subPath}' is not defined by "exports" in ${pkgPath}`);
+
+function findExportKeyMatch(exportMap, subPath) {
+  for (const [key, value] of Object.entries(exportMap)) {
+    if (key.endsWith('*') && value.includes('*')) {
```
guybedford, line 18:
```suggestion
    if (key.endsWith('*')) {
```

`value` doesn't strictly have to include `*` - it is possible to have a static map theoretically.

packages/node-resolve/src/resolveImportSpecifiers.js
```diff
@@ -0,0 +1,197 @@
...
+      // star match: "./foo/*": "./foo/*.js"
+      const keyWithoutStar = key.substring(0, key.length - 1);
+      if (subPath.startsWith(keyWithoutStar)) {
+        return key;
+      }
```
guybedford, line 23: There isn't any requirement that the subpath be the same - it is fine to do `"exports": { "./virtual/*": "./local/*/index.js" }`.

packages/node-resolve/src/resolveImportSpecifiers.js
```diff
@@ -0,0 +1,197 @@
...
+  return null;
+}
+
+function mapSubPath(pkgJsonPath, subPath, key, value) {
+  if (typeof value === 'string') {
+    if (typeof key === 'string' && key.endsWith('*')) {
+      // star match: "./foo/*": "./foo/*.js"
+      const keyWithoutStar = key.substring(0, key.length - 1);
+      const subPathAfterKey = subPath.substring(keyWithoutStar.length);
+      return value.replace('*', subPathAfterKey);
```
guybedford, line 45: Multiple `*` entries can be used on the right hand side, so a replace all is probably best here:

```suggestion
      return value.replace(/\*/g, subPathAfterKey);
```

guybedford (commented):
Left a few implementation comments.

Interesting to see the difference in implementation assumptions around subpath patterns in exports... if you have any feedback further please do share.

packages/node-resolve/src/resolveImportSpecifiers.js
LarsDenBakker: This checks whether the user's import starts with the part that comes before the star in the key. For example:

```json
"exports": { "./virtual/*": "./local/*/index.js" }
```

should match:

`import './virtual/foo`

but not:

`import './foo`

Correct?

guybedford: Ah right, seems good!

LarsDenBakker:
@guybedford thanks for the review!

packages/node-resolve/types/index.d.ts
```diff
@@ -9,6 +9,17 @@ export const DEFAULTS: {
...
+   * Additional conditions of the package.json exports field to match when resolving modules.
+   * By default, this plugin looks for the `'default', 'module', 'import']` conditions when resolving imports.
+   *
+   * When using `@rollup/plugin-commonjs` v16 or higher, this plugin will use the
+   * `['default', 'module', 'import']` conditions when resolving require statements.
+   *
+   * Setting this option will add extra conditions on top of the default conditions.
+   * See https://nodejs.org/api/packages.html#packages_conditional_exports for more information.
+   */
+  exportconditions?: string[];
```
guybedford, line 22: Should be `exportConditions` I think.

guybedford (approved):

packages/node-resolve/types/index.d.ts
LarsDenBakker: Indeed

packages/node-resolve/test/browser.js
```diff
@@ -190,7 +190,7 @@ test('supports `false` in browser field', async (t) => {
   await testBundle(t, bundle);
 });
 
-test('pkg.browser with mapping to prevent bundle by specifying a value of false', async (t) => {
+test.only('pkg.browser with mapping to prevent bundle by specifying a value of false', async (t) => {
```
lukastaegert, line 193:
```suggestion
test('pkg.browser with mapping to prevent bundle by specifying a value of false', async (t) => {
```

Apparently this change has not been comitted to the branch...

lukastaegert (commented):
Great work! This is ok from my side, but we should really get rid of the `.only` :)

lukastaegert (approved):

Issue #324, opened 2020-04-22
wasm plugin shouldn’t use base64 unless a file is marked as `sync`


surma:
- Rollup Plugin Name: @rollup/plugin-wasm 
- Rollup Plugin Version: 3.0.0
- Rollup Version: 2.0
- Operating System (or Browser): Chrome on MacOS
- Node Version: v12.8

### How Do We Reproduce?

https://repl.it/@surma/SaneCarelessSphere

### Expected Behavior

The .wasm file should be emitted as an asset and loaded using `fetch()` when the .wasm file is not marked as `sync.

### Actual Behavior

The .wasm file is inlined using base64, causing unnecessary file bloat.

I am opening an issue first to see if y’all would accept an PR to change the behavior of this plugin. Either me or @rreverser would be happy to whip one up :D

shellscape:
> The .wasm file should be emitted as an asset and loaded using fetch() when the .wasm file is not marked as `sync.

That sounds like a browser environment. When working on a patch for this, it'll be important to consider non-browser environments.

surma:
Sure, I wouldn’t break node support. Although it might be preferrable to make the target environment an option of the plugin to avoid loading unnecessary code on the web.

RReverser:
Note that plugin runtime code already performs different actions for Node.js vs browser, so it would be quite easy to perform `fetch` + `instantiateStreaming` only in browser: https://github.com/rollup/plugins/blob/e9651fb58e01fce6af908c66d5de78542e56a41e/packages/wasm/src/index.js#L30

manzt:
I'm curious if folks here have thoughts around best practices with shipping wasm in a library. I've found `base64` encoding to be very useful for making a wasm-based module npm-installable and easy to handle with bundlers. However, I'm not happy with using `base64` since it adds bloat and feels like a win for DX and not users ultimately.

RReverser:
I don't really see why separate Wasm files are any harder to npm install or use in other apps?

manzt:
Perhaps my nativity is showing here, but I'm not sure if something like `@rollup/plugin-node-resolve` can will end up adding the wasm from an npm package as an asset in the final bundle.

```javascript
// some-npm-module.js
async function init() {
  const { instance }  = await WebAssembly.instantiateStreaming(fetch('simple.wasm'), importObject)
  return instance 
}
export init;
```

```javascript
// my-lib.js
import { init } from 'some-npm-module';
init();
```
[...]

PR #543, opened 2020-08-16
Option to not inline wasm

bminixhofer:
## Rollup Plugin Name: `WASM`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

Discussed in https://github.com/rollup/plugins/issues/324.

### Description

This PR adds support for not inlined WASM files. Currently WASM files are always base64 encoded, which increases size by > 30%. There's some discussion in https://github.com/rollup/plugins/issues/324. This PR addresses this issue by adding a `limit` and `publicPath` option like in the [URL plugin](https://github.com/rollup/plugins/tree/master/packages/url). E. g.

```js
wasm({
  limit: 0,
})
```

would mean that no WASM files are inlined and all are copied to the destination folder and loaded via `fs` (in Node) or `fetch`(in the browser) at runtime.

### Implementation

1.) Switching between `fs` and `fetch`

I decided to include code for both and decide at runtime what to execute. The other option would be having a plugin flag for the target platform. But I opted for this since I didn't see a downside besides the negligible code size impact. 

One noteworthy detail: I loaded `fs` via `eval('require("fs")')` to avoid warnings like https://github.com/darionco/rollup-plugin-web-worker-loader/issues/20. Of course `eval` is typically bad but I don't see an issue here.

2.) Path and name of the output WASM files

I handled file generation similarly to the URL plugin. The differences are:
- I did not add an `emitFiles` option.
- I did not add a `fileName` option, instead the output file name is always `[hash].wasm`.
- I did not add `destDir` and `sourceDir` options.

Let me know if you think these are needed here as well. I didn't add them now because I wanted to make sure that my approach is valid in principle first.

### Tests

I added one test which:
- bundles `fixtures/complex.js` as CJS module.
- asserts that the [hash].wasm file is present in the output folder.
- Runs the `output/bundle.js` with its assertions.


packages/wasm/README.md
```diff
@@ -53,6 +53,22 @@ Default: `null`
 
 Specifies an array of strings that each represent a WebAssembly file to load synchronously. See [Synchronous Modules](#synchronous-modules) for a functional example.
 
+### `limit`
```
shellscape, line 56: could we go with `maxFileSize` here instead?

packages/wasm/src/index.ts
```diff
@@ -40,26 +89,66 @@ export function wasm(options: RollupWasmOptions = {}): Plugin {
...
       if (code && /\.wasm$/.test(id)) {
-        const src = Buffer.from(code, 'binary').toString('base64');
-        const sync = syncFiles.indexOf(id) !== -1;
-        return `export default function(imports){return _loadWasmModule(${+sync}, '${src}', imports)}`;
+        const isSync = syncFiles.indexOf(id) !== -1;
+        const filepath = copies[id] ? `'${copies[id]}'` : null;
+        let src;
+
+        if (filepath === null) {
+          assert(!isSync, 'non-inlined files can not be `sync`.');
```
shellscape, line 108: we really don't use `assert` in plugins. instead, we'll use `this.error` or `this.warn`

packages/wasm/src/index.ts
```diff
@@ -40,26 +89,66 @@ export function wasm(options: RollupWasmOptions = {}): Plugin {
...
+      await makeDir(base);
+
+      await Promise.all(
+        Object.keys(copies).map(async (name) => {
+          const output = copies[name];
+          // Create a nested directory if the fileName pattern contains
+          // a directory structure
+          const outputDirectory = path.join(base, path.dirname(output));
+          await makeDir(outputDirectory);
+          return copy(name, path.join(base, output));
```
shellscape, line 136: the preferred method for copying files to the bundle destination is similar to this: https://github.com/rollup/plugins/blob/23df786d92318d5e142e60c5e980a7698aa2c617/packages/html/lib/index.js#L104-L113

you're already performing a read on line 31 above, so stashing the buffer there to pass to `this.emitFile` here shouldn't cause any trouble.

packages/wasm/README.md
bminixhofer: Sure - I chose `limit` because it is comparable to the `limit` option of the URL plugin.

packages/wasm/src/index.ts
bminixhofer: Ok, I chose this again because of code from the URL plugin:

https://github.com/rollup/plugins/blob/23df786d92318d5e142e60c5e980a7698aa2c617/packages/url/src/index.js#L84-L91

is there any reason for it being used there instead of `this.emitFile`?

shellscape: yeah, a lot of the plugins still have legacy code in them from much earlier versions of rollup. just one of those legacy things that we haven't gotten around to changing. The `fs` methods are still valid, but we prefer `this.emitFile` now.

bminixhofer: Fixed now, I just named it "Rollup WASM Asset".

bminixhofer:
Now that these initial changes are done: Do we need `emitFiles`, `fileName`, `sourceDir`, and `destDir` options like in the URL plugin? (e. g. https://github.com/rollup/plugins/tree/master/packages/url#emitfiles). Would be easy to implement but I'm not sure how useful they are.

shellscape:
> Do we need emitFiles, fileName, sourceDir, and destDir options like in the URL plugin?

I don't see a need right now. Those options in the `url` plugin came about from user feature requests. Unless there's a call from the userbase for options on where to emit the files, I think you're good to go.

packages/wasm/src/index.ts
```diff
@@ -1,34 +1,86 @@
...
+import * as path from 'path';
+import { promisify } from 'util';
+import { createHash } from 'crypto';
 
 import { Plugin } from 'rollup';
 
 import { RollupWasmOptions } from '../types';
 
+const fsStatPromise = promisify(fs.stat);
+const fsReadFilePromise = promisify(fs.readFile);
```
pnevares, line 11: It looks like the fsPromises API for both of these has been around since at least Node 10.x (the minimum engine for plugin-wasm), unless there's a reason they can't be used?

bminixhofer: Oh, sure I'll gladly fix that. This was based on code in the URL plugin:

https://github.com/rollup/plugins/blob/55a9cb20d7686750aa104d6a45a7d6834f57ba4a/packages/url/src/index.js#L10-L11

So I didn't think much of it and thought it was a backwards-compatibility issue.

bminixhofer: Fixed now.

pnevares: Thanks for calling that out, definitely something to keep in mind for cleanup!

entrptaher:
The `publicPath` is not working, if we want to put the generated wasm files in seperate directory, it will not do so. Might as well create a new issue about it.

image

PR #544, opened 2020-08-16
fix(typescript): Change `noEmitOnError` default to false

NotWoods:
BREAKING CHANGE: `noEmitOnError` now defaults to false instead of true. This means that Rollup will show a warning when type errors are encountered but will still compile. The previous behaviour is available by changing the setting to `true`.

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] no

### Description

For watch mode to work, warnings must be used instead of errors. This behaviour better matches how Typescript works, since it only warns for type errors but compiles anyways. However, its different from how bundlers like Webpack work, as they usually fail.

### DIFF
--- a/packages/typescript/README.md
+++ b/packages/typescript/README.md
@@ -13,7 +13,7 @@
 
 ## Requirements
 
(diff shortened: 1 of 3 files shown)


dummdidumm:
Don't wanna rush, but is there anything blocking you from merging this in?

NotWoods:
Waiting for review

benmccann:
Thank you for this!! It will make it so much easier for newcomers to get started

PR #550, opened 2020-08-20
feat(pluginutils): normalizePath

eight04:

This PR contains:

- [x] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

This PR includes:

1. A new utility `normalizePath`.
2. Fixes a bug on Windows. On Windows 10, test failed because `path.resolve` creates patterns including backward slash `\`. This patch adds a normalization similar to `createFilter`:
https://github.com/rollup/plugins/blob/55a9cb20d7686750aa104d6a45a7d6834f57ba4a/packages/pluginutils/src/createFilter.ts#L15-L17
3. Small quotes typo in README.

### DIFF
--- a/packages/pluginutils/README.md
+++ b/packages/pluginutils/README.md
@@ -232,6 +232,22 @@ makeLegalIdentifier('foo-bar'); // 'foo_bar'
 makeLegalIdentifier('typeof'); // '_typeof'
 ```
 
+### normalizePath
+
+Converts path separators to forward slash.
+
+Parameters: `(filename: String)`<br>
+Returns: `String`
+
+#### Usage
+
+```js
+import { normalizePath } from '@rollup/pluginutils';
+
+normalizePath('foo\\bar'); // 'foo/bar'
+normalizePath('foo/bar'); // 'foo/bar'
+```
+
 ## Meta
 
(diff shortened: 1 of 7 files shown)


shellscape:
@eight04 please take a look at the last comment

eight04:
How does typescript plugin use the normalization? If it also relates to picomatch then we can add a `compilePicomatch` function.

Or, do you want a small util like this:

```js
export function normalizePathSep(path) {
  return path
    .split(sep)
    .join('/');
}
```

NotWoods:
Just a small util like that, although I prefer the [implementation in the typescript plugin](https://github.com/rollup/plugins/blob/master/packages/typescript/src/index.ts#L29) as it uses constants from node.

packages/pluginutils/README.md
```diff
@@ -88,9 +88,9 @@ export default function myPlugin(options = {}) {
         },
         leave(node) {
           if (node.scope) scope = scope.parent;
-        }
+        },
```
eight04, line 91: I think these were added by prettier.

NotWoods (approved):

pnevares (approved):

shellscape:
thanks!

PR #564, opened 2020-09-05
fix(node-resolve): do not ignore exceptions

tjenkinson:

This PR contains:

- [x] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] yes _kind of_
- [ ] no

Breaking Changes?

- [x] no

### Description

_recommend viewing with indentation disabled on github_

I was looking to add a feature to return `false` from `resolveId` for builtins when `preferBuiltins` was enabled but noticed code wasn't running that I was expecting. I will follow up with this in a [different PR (which I'll rebase that when this PR is finished)](https://github.com/tjenkinson/plugins/pull/1).

There was a `try/catch` which was ignoring any exception and returning `null` from `resolveId`. I tried removing that block and it caused test failures, and was why my changes weren't working 😱.

I fixed the hidden issues to make the tests pass again. The 'root-dir' test looked like it was a bug in the test itself so I updated that test.

I think it would be better to let any unexpected errors throw out from `resolveId` and fail a build instead of silently converting them to a `null` return. If you'd like that back maybe we can figure out a way to have that logic disabled for tests.


packages/node-resolve/src/index.js
```diff
@@ -100,6 +100,12 @@ export function nodeResolve(opts = {}) {
       // ignore IDs with null character, these belong to other plugins
       if (/\0/.test(importee)) return null;
 
+      if (/\0/.test(importer)) {
+        // handle cases like common-js plugin which has form
+        // \u0000<id>?commonjs-proxy
+        importer = importer.slice(1);
```
tjenkinson, line 106: not sure if there's a better way to handle this case?

If you remove this change you'll see the `pkg.browser with mapping to prevent bundle by specifying a value of false` test fail because the call to `resolveId` with the null byte is not allowed

packages/node-resolve/test/root-dir.js
```diff
@@ -7,11 +7,9 @@ const { testBundle } = require('../../../util/test');
 
 const { nodeResolve } = require('..');
 
-process.chdir(join(__dirname, 'fixtures', 'monorepo-dedupe', 'packages', 'package-a'));
-
 test('deduplicates modules from the given root directory', async (t) => {
   const bundle = await rollup({
-    input: 'index.js',
+    input: './packages/package-a/index.js',
```
tjenkinson, line 12: this was causing an error because the input could not be found. It should be relative to the `rootDir` right?

danielgindi:
I'm a bit confused. I didn't see anywhere that you presented a specific use case, or your specific frontal-accident with this plugin...

Was it trying to somehow resolve modules that are actually internal commonjs proxies? Or was it just the importer being a commonjs proxy that broke things?

And are you now throwing the errors, failing the whole pipeline - or emitting warnings? (Because I see you removed the original try-catch, instead of altering it)

danielgindi:
@tjenkinson so basically avoid having to specify extra `externals` in the rollup level?

About the exceptions- I'm not sure I agree.
In general what you are saying is true.
But there are cases where a warning is more appropriate. If you failed to resolve a module (for whatever unknown reason), you don't necessarily want to stop everything.
Other modules print a warning, suggesting a possible root cause.
This means that if there is a failure in node-resolve, the user in some cases might be able to do nothing to bypass the error if he deems the error "ah, that's fine by me".

Also- removing the catch is a BREAKING CHANGE in that manner, and you did not mark it as one in the intro ;)

danielgindi:
> Happy to put the catch back and log a warning in there if people would prefer that. However IMO as a developer I'd want there to be a blocker if something unexpected went wrong. Especially when some of the options like `jail` seem like they're meant to to prevent things outside a certain location being imported? I'd prefer it to fail safe and stop on an issue then return `null` and hand off to the next resolver.

I guess we need more opinions here

> Yeh I wasn't sure on this. I'm fine with it being breaking (or putting the catch back but with at least a warning log :) ). Also disabling that catch for tests would be nice.

Btw, if only addings warnings to rollup, you could look at other test (i.e in commonjs) - some of them test for the existing of a warning.
So if it's the expected flow, to have a warning - or even an exception - then you should be expecting that in a test and make sure it happens.

LarsDenBakker:
In my opinion a failure to resolve a module should result in an error and not into making the module an external, that's something that confuses users a lot currently. However this is not something that should be handled by the node resolve plugin, it should be on the level of rollup. There could be multiple plugins who resolve an id, and node-resolve shouldn't block further resolving.

LarsDenBakker:
An unexpected error should result in a failure, but I thought the try/catch catches errors thrown by the `resolve` package when a module couldn't by found. That's how it relates to this PR.

packages/node-resolve/src/index.js
lukastaegert: That one is a really good question that got overlooked so far. I do not agree with the solution, though. Files starting with `\0` are under the sole maintenance of some plugin, and we should not make assumptions about what the part behind the `\0` means.
I think a valid approach might be to treat the importer like `undefined`, which is the value that gets passed for entry points. At least for the commonjs plugin, this should work nicely as all imports from those proxy files are fully resolved absolute paths anyway, and I would expect the same from other plugins as well.

lukastaegert:
I agree that this big catch was really bad for testing, but the general idea of course was that when this plugin cannot resolve a module, other resolvers should get a chance to do so before throwing.

So the question to me is that is not entirely clear to me from looking at the code: Will this still be the case for all "relevant" cases? And by that I mean it is definitely ok to throw if there is a weird file system error or some error that results from data in an unexpected format being passed to this plugin.

If you think this is the case, we could try to release this fix but be prepared to do a quick rollback in case we missed something important here.

> In my opinion a failure to resolve a module should result in an error and not into making the module an external, that's something that confuses users a lot currently. However this is not something that should be handled by the node resolve plugin, it should be on the level of rollup. There could be multiple plugins who resolve an id, and node-resolve shouldn't block further resolving.

@LarsDenBakker If you say so, we can talk about this, could be a change in the next major version. Personally, I like this behaviour for quick prototyping, but if it confuses newcomers, then that is bad. Rollup could augment the error with instructions on how to mark the specific import as external to solve the issue.

I also think a lot of the woes that make this feature helpful is that builtins are not marked as external when using the `preferBuiltins` option in this plugin. As I understand @tjenkinson is working on a fix for just that, then one more argument for making this an error in Rollup.

packages/node-resolve/src/index.js
tjenkinson: >I think a valid approach might be to treat the importer like undefined, which is the value that gets passed for entry points

@lukastaegert I pushed a commit to do this

lukastaegert:
> Do we need to check a file at resolved exists here?

Looking at what this option is supposed to be doing and the current implementation, I am leaning towards file-not-found errors should also return `null` instead of throwing here (while other errors should throw), but that is an opinion.

lukastaegert:
Hard to say. I am not aware of any official plugins for which this would be breaking, but it is possible for third party plugins.

lukastaegert (approved):

danielgindi (approved):

shellscape:
thanks!

PR #584, opened 2020-09-23
fix(babel): clone cached helper identifier before returning it

Andarist:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

I don't really have a failing test for this and I don't think it's worth testing any implementation details to just have this verified. Usually one doesn't have to worry about inserting duplicated nodes but it's a good practice to never do this and always insert fresh nodes.


shellscape:
@Andarist moving forward you're welcome to merge these in yourself. We always do a `squash` commit unless the PR affects multiple packages (which is rare)

PR #590, opened 2020-09-28
Feat resolve fallback

chengcyber:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

Assuming a  package has the following package.json

```
{ 
  "main": "entry.main.js",
  "module": "entry.esm.js"
}
```

config `@rollup/plugin-node-resolve` with `mainFields: ['module', 'main']`.

but, `entry.esm.js` does not exist. This leads to a `MODULE_NOT_FOUND` error. After this PR merged, `resolve` can fallback to the next `mainField`. `main` in this case.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
       "prettier --single-quote --write"
     ],
     "*package.json": [
-      "prettier --write --plugin=prettier-plugin-package"
+      "prettier --write --plugin=prettier-plugin-package --with-node-modules"
     ]
   },
   "nyc": {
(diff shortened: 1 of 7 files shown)


chengcyber:
> What real world problems does this solve? Is this not a bug in the package?

Yes, it is a bug in the package, but sometimes it is kind of hard to upgrade a package to the correct version, which is actually my real world problem...
So, there are two ways to make it works:

- write a hard code rollup plugin to handle this buggy package
- node-resolve plugin can handle this case (this PR)

I try to bring the capacity to `node-resolve` plugin, but if it is in a wrong direction in your mind, please close this PR freely. I will pick up the other solution, Thanks.

chengcyber:
Since the unclear decision for this tech choice, i would like to close this PR for now

PR #599, opened 2020-10-08
fix(commonjs): register dynamic modules when a different loader (i.e typescript) loads the entry file

danielgindi:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

I've encountered in a TypeScript project a weird phenomenon, where the entry file does not go through `load` in the `commonjs` plugin.  
As a result, the of prepending the main entry with the registration calls of the dynamic modules - does not take place.  
This means that modules were trying to call dynamic modules, but they were not in the runtime commonjs registry, and so they failed.

I fixed this by moving this phase to the `transform` rather than `load`.  
A small fix that had to go hand in hand, is making sure that these `commonjsRegister` calls are always first as some modules may be appended to the `sources` array before them.

All tests pass, and this works now in typescript project.


danielgindi:
@lukastaegert Do you see any possible issues with this?
I basically just moved things around, so current tests keep passing, and a TS project can now also use this.

lukastaegert (changes requested):
I am a little behind looking through these PRs so please be patient. So you claim that this fixes a bug. Unfortunately I am completely missing a test that was broken without this change and is fixed with this change. This would serve both as documentation and as a guarantee that future refactorings do not break this again. Could you add one?

danielgindi:
> I am a little behind looking through these PRs so please be patient. So you claim that this fixes a bug. Unfortunately I am completely missing a test that was broken without this change and is fixed with this change. This would serve both as documentation and as a guarantee that future refactorings do not break this again. Could you add one?

Done. Now there's a good test for this :-)

danielgindi:
I've rebased on master. Snapshots. Every time...

lukastaegert (approved):
Looks good!

shellscape:
@danielgindi standard ops here is to wait for 2 approvals. please lend that green check mark if you think it's good to go and we'll merge it post haste

danielgindi:
> @danielgindi standard ops here is to wait for 2 approvals. please lend that green check mark if you think it's good to go and we'll merge it post haste

Yeah I know :-) I intentionally tagged @LarsDenBakker. As whatever blocker it is, it's always a good idea to have another set of eyes before publishing to production...

LarsDenBakker:
Fine for me!

danielgindi:
> Fine for me!

Yeah I think @shellscape really wants the green checkmark 🤣

shellscape:
:) it's useful for tracking history

shellscape (approved):

danielgindi:
@shellscape I think that this fix deserves a release, as TS users currently can't fully benefit from commonjs plugin. I bumped into this and using a branch, but everybody else dont :)

shellscape:
@danielgindi we merged this less than 24 hours ago. patience please! 🙏

danielgindi:
> @danielgindi we merged this less than 24 hours ago. patience please! 🙏 

Apologizes, I didn't mean to dictate the exact timeline 😂

PR #607, opened 2020-10-16
refact(wasm): use helper module instead of banner

chengcyber:
## Rollup Plugin Name: `@rollup/plugin-wasm`

This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

For now, `wasm` plugin is using `banner` API to inject helper function `_loadWasmModule`, but it will inject this helper function in every file, even those files is unrelated to wasm things.
After this PR merged, it uses a helper module to inject `_loadWasmModule`, hopes to help removing the redundant code

### DIFF
--- a/packages/wasm/src/helper.ts
+++ b/packages/wasm/src/helper.ts
@@ -1,0 +1,55 @@
+export const HELPERS_ID = '\0wasmHelpers.js';
+
+export const getHelpersModule = () => `
+function _loadWasmModule (sync, filepath, src, imports) {
+  function _instantiateOrCompile(source, imports, stream) {
+    var instantiateFunc = stream ? WebAssembly.instantiateStreaming : WebAssembly.instantiate;
+    var compileFunc = stream ? WebAssembly.compileStreaming : WebAssembly.compile;
+
+    if (imports) {
+      return instantiateFunc(source, imports)
+    } else {
+      return compileFunc(source)
+    }
+  }
+
+  var buf = null
+  var isNode = typeof process !== 'undefined' && process.versions != null && process.versions.node != null
+
+  if (filepath && isNode) {
+    var fs = eval('require("fs")')
+    var path = eval('require("path")')
+
+    return new Promise((resolve, reject) => {
+      fs.readFile(path.resolve(__dirname, filepath), (error, buffer) => {
+        if (error != null) {
+          reject(error)
+        }
+
+        resolve(_instantiateOrCompile(buffer, imports, false))
+      });
+    });
+  } else if (filepath) {
+    return _instantiateOrCompile(fetch(filepath), imports, true)
+  }
+
+  if (isNode) {
+    buf = Buffer.from(src, 'base64')
+  } else {
+    var raw = globalThis.atob(src)
+    var rawLength = raw.length
+    buf = new Uint8Array(new ArrayBuffer(rawLength))
+    for(var i = 0; i < rawLength; i++) {
+       buf[i] = raw.charCodeAt(i)
+    }
+  }
+
+  if(sync) {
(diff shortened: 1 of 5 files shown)


shellscape:
Thanks for the PR. You've checked the box that tests are included but haven't provided any tests in the PR. Please add tests and we'll review.

packages/wasm/src/helper.ts
```diff
@@ -0,0 +1,55 @@
+export const HELPERS_ID = '\0wasmHelpers.js';
```
canadaduane, line 1: What does this null byte prefix do?

chengcyber:
> Thanks for the PR. You've checked the box that tests are included but haven't provided any tests in the PR. Please add tests and we'll review.

I means the PR passes all existing tests, no more tests are added in this PR

packages/wasm/src/helper.ts
chengcyber: I am not 100 percent sure where to find the documentation, but "\0" seems to be a convention when writing rollup plugin to indicate the module comes from a plugin. and if it should not be taken by this plugin, resolveId returns null, otherwise returns the current resolveId

shellscape: It indicates a "virtual module" in the Rollup ecosystem.

shellscape:
Sorry, but we cannot accept pull requests for fixes or features without accompanying tests. You are refactoring some code, but what your refactor attempts to accomplish is a fix. You must write new tests that demonstrate the conditions your changes resolve before we can accept the PR. This requirement exists to protect against future regression.

chengcyber:
> Sorry, but we cannot accept pull requests for fixes or features without accompanying tests. You are refactoring some code, but what your refactor attempts to accomplish is a fix. You must write new tests that demonstrate the conditions your changes resolve before we can accept the PR. This requirement exists to protect against future regression.

@shellscape I totally agree to prevent future regression by adding new tests, but kind of confusing about how to add new tests for this refactor. Since IMO the existing tests cover the functionality of the current codebase, and this PR does not add/remove more functionality...(just moving the code in `banner` to a virtual module). 

Could you kindly help to figure out the direction more clear?

shellscape:
Sure. You have a clear before/after result. Previously the bundle would contain multiple copies of the helper. With your changes it should contain only one. That's the test.

chengcyber:
@shellscape  A test for injecting helper is added, using a custom detect plugin to check whether the module contains inject import. PTAL

packages/wasm/test/test.js
```diff
@@ -121,3 +121,31 @@ try {
...
+        sync: ['fixtures/sample.wasm']
+      }),
+      {
+        name: 'test-detect',
+        transform: (code, id) => {
+          if (id.endsWith('sample.wasm')) {
+            t.true(code.includes(injectImport));
+          }
+          if (id.endsWith('foo.js')) {
+            t.true(!code.includes(injectImport));
```
shellscape, line 143:
```suggestion
            t.false(code.includes(injectImport));
```

shellscape:
@bminixhofer since you recently contributed here, please have a look at this change.

packages/wasm/src/helper.ts
lukastaegert: More precisely it means: Other plugins, please ignore this module and please do not apply any transformations etc.  https://rollupjs.org/guide/en/#conventions

lukastaegert (approved):
Did not test it in detail but looks good to me!

bminixhofer:
Looks good to me as well!

shellscape:
thanks!

PR #612, opened 2020-10-17
docs(alias): add helpful notes

davidrot:
## Rollup Plugin Name: `Alias`

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [x] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

List any relevant issue numbers: -

### Description

Added my thoughts and findings to the readme. Those information would have helped me very much.


packages/alias/README.md
```diff
@@ -27,6 +27,8 @@ import batman from 'batman';
 
 If this seems familiar to Webpack users, it should. This is plugin mimics the `resolve.extensions` and `resolve.alias` functionality in Webpack.
 
+This plugin is working for all kind of files (.ts, .svelte, ...)
```
shellscape, line 30: What was the intent behind adding this line?

packages/alias/README.md
```diff
@@ -67,6 +69,8 @@ module.exports = {
 
 Then call `rollup` either via the [CLI](https://www.rollupjs.org/guide/en/#command-line-reference) or the [API](https://www.rollupjs.org/guide/en/#javascript-api). If the build produces any errors, the plugin will write a 'alias' character to stderr, which should be audible on most systems.
 
+The find input should not end with a slash or it will not work.
```
shellscape, line 72: Please try rephrasing this. I'm not quite sure what you're trying to speak about here.

shellscape (changes requested):
Thanks for opening a PR. Would like some clarity around your changes.

packages/alias/README.md
davidrot: I thought this plugin is not compatible with ts-files but this plugin makes the typescript paths feature (https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping) kind of obsolet.

packages/alias/README.md
davidrot: Maybe you can help me, englisch is not my mother tongue. With this phrase I was trying to describe this (https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping) check. When I was first trying this plugin it was not working for me. After investigating and debugging what took me a while i saw this check.

packages/alias/README.md
shellscape:
```suggestion
This plugin will work for any file type that Rollup natively supports, or those which are [supported by third-party plugins](https://github.com/rollup/awesome#other-file-imports).
```

packages/alias/README.md
shellscape:
```suggestion
```

I'm going to remove this here, and place it in a more appropriate spot.

PR #615, opened 2020-10-21
yaml plugin: provide file id to transform function

janosh:
## Rollup Plugin Name: `YAML`

This PR contains:

- [ ] bugfix
- [x] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

Non-breaking change since people don't have to use the extra transformer argument.

### Description

Reason for this change: I need to apply different content transformations depending on which file the data came from. Without knowing the file, the transformer function has to rely on brittle heuristics to identify the file based on its content (which can obviously change over time).

I think a new test case isn't needed for this but happy to provide one if someone thinks otherwise.

### DIFF
--- a/packages/yaml/test/test.js
+++ b/packages/yaml/test/test.js
@@ -59,8 +59,11 @@ test('resolves extensionless imports in conjunction with nodeResolve plugin', as
 });
 
 test('applies the optional transform method to parsed YAML', async (t) => {
-  const transform = (data) => {
+  const transform = (data, filePath) => {
+    // check that transformer is passed a correct file path
+    t.truthy(typeof filePath === 'string' && filePath.endsWith('.yaml'));
     if (Array.isArray(data)) {
+      t.truthy(filePath.endsWith('array.yaml'));
       return data.filter((datum) => !datum.private);
     }
     Object.keys(data).forEach((key) => {
--- a/packages/yaml/test/test.js
+++ b/packages/yaml/test/test.js
@@ -61,9 +61,9 @@ test('resolves extensionless imports in conjunction with nodeResolve plugin', as
 test('applies the optional transform method to parsed YAML', async (t) => {
   const transform = (data, filePath) => {
     // check that transformer is passed a correct file path
-    t.truthy(typeof filePath === 'string' && filePath.endsWith('.yaml'));
+    t.true(typeof filePath === 'string' && filePath.endsWith('.yaml'), filePath);
     if (Array.isArray(data)) {
-      t.truthy(filePath.endsWith('array.yaml'));
+      t.true(filePath.endsWith('array.yaml'), filePath);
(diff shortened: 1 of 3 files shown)


NotWoods (changes requested):
We still need a test case to make sure the correct path is being sent through. You can modify some of the existing tests to add that check.

janosh:
@NotWoods Odd. If I run `pnpm run test --filter ./packages/yaml` locally, the tests are passing. Am I missing something?

packages/yaml/test/test.js
```diff
@@ -59,8 +59,9 @@ test('resolves extensionless imports in conjunction with nodeResolve plugin', as
 });
 
 test('applies the optional transform method to parsed YAML', async (t) => {
-  const transform = (data) => {
-    if (Array.isArray(data)) {
+  const transform = (data, filePath) => {
+    // check that transformer is passed the right file path
+    if (Array.isArray(data) && filePath.endsWith('/array.yaml')) {
```
NotWoods, line 64:
```suggestion
    t.ok(filePath.endsWith('/array.yaml'), filePath);
    if (Array.isArray(data)) {
```
I'm not sure why the test is failing, but moving the file path check to a separate line should make it easier to read and debug.

janosh:
Ah, different directory separators on UNIX (`'/'`) and Windows (`'\'`) is why it was failing.

packages/yaml/test/test.js
```diff
@@ -59,8 +59,11 @@ test('resolves extensionless imports in conjunction with nodeResolve plugin', as
 });
 
 test('applies the optional transform method to parsed YAML', async (t) => {
-  const transform = (data) => {
+  const transform = (data, filePath) => {
+    // check that transformer is passed a correct file path
+    t.truthy(typeof filePath === 'string' && filePath.endsWith('.yaml'));
```
NotWoods, line 64: Let's just use `t.true` rather than `truthy`. Passing the file path as a second argument will print it if the assertion fails.
```suggestion
    t.true(typeof filePath === 'string' && filePath.endsWith('.yaml'), filePath);
```
Once this change is made to both assertions we're good to merge

janosh: done

NotWoods (approved):
Thanks!

shellscape:
@NotWoods you didn't use conventional commit on that merge. that'll hose our versioning and changelog generation. can you amend that commit and force push?

janosh:
@NotWoods Can you say when this will be published?

shellscape:
@janosh we typically make weekly publish rounds for plugins. we have a lot moving parts this week so don't be surprised if this takes until early next week to be seen on NPM. _(We're not quite ready for automated publishing)_

PR #618, opened 2020-10-23
fix(typescript): add composite to validation checks

dummdidumm:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: -

### Description

TypeScript's `composite` compiler option produces declaration files, too. I stumbled upon this and did get a generic validation error right at the end when the plugin wants to print the files to the outdir but cannot find one. To fix it, I added an additional check to the relevant validation.

### DIFF
--- a/packages/typescript/src/options/validate.ts
+++ b/packages/typescript/src/options/validate.ts
@@ -83,7 +83,7 @@ export function validatePaths(
     }
   }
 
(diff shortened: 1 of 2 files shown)


NotWoods (approved):
LGTM!

PR #619, opened 2020-10-24
docs(typescript): More informative error messages

dummdidumm:
## Rollup Plugin Name: `typescript`

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [x] yes (_bugfixes and features will not be merged without tests_)
- [ ] no

Breaking Changes?

- [ ] yes (_breaking changes will not be merged unless absolutely necessary_)
- [x] no

List any relevant issue numbers: -

### Description

As a Rollup novice I had some problems figuring out what some of the error messages meant. More specifically, I did not know what options they talked about, so I added more context, telling if it's either a Rollup or TS compiler option.

PR #621, opened 2020-10-24
Better error messages

everdimension:

This PR contains:

- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

I saw the original `'dir' must be used when 'outDir' is specified` error and spent a lot of time figuring it out because ts compiler doesn't have this 'dir' option (https://www.typescriptlang.org/docs/handbook/compiler-options.html), but the error sounds as if it should

I changed the message a bit to avoid this confusion.


everdimension:
Just saw https://github.com/rollup/plugins/pull/619/files PR, which is more comprehensive 👍

PR #627, opened 2020-11-01
feat(node-resolve): Mark built-ins as external

tjenkinson:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [x] yes (updated)
- [ ] no

Breaking Changes?

_Not sure on this given previously it was a warning. It would have allowed other plugins to jump in though so maybe it should be 'yes'?_

- [x] yes
- [ ] no

### Description
This continues from https://github.com/rollup/plugins/pull/564.

[I thought that if `useBuiltins` was `true` then any builtins would automatically be marked as external](https://github.com/rollup/rollup/issues/3751#issuecomment-682681151). This was not the case. This PR does that. Let me know if you think it doesn't make sense.

It also makes some of the docs a bit more accurate.

It seems a bit strange that some options (like `jail`) don't mark an import as external, but others (like `resolveOnly` and now `preferBuiltins`) do. I wonder if there is a reason or if we should make them consistent?

There is one commit that updates all the snapshots as I noticed when updating the ones for the tests I updated it changed them all because the ava domain in the comment changed.


packages/node-resolve/README.md
```diff
@@ -118,10 +118,12 @@ DEPRECATED: use "resolveOnly" instead
 ### `preferBuiltins`
 
 Type: `Boolean`<br>
-Default: `true`
+Default: unset
 
 If `true`, the plugin will prefer built-in modules (e.g. `fs`, `path`). If `false`, the plugin will look for locally installed modules of the same name.
 
+If unset the plugin will first look for modules of the same name, and then fall back to built-in modules.
```
lukastaegert, line 125: If we keep the warning about `unset` that still appears to be in the code, I guess this should be mentioned here, otherwise people think they can get away with not configuring this option when they use imports like `fs`. Alternatively, we could allow this as a valid "third"  option, e.g. `"auto"`, where you want to override some builtins with local alternatives but keep others. But I am not sure I can imagine a scenario where people would want this.

lukastaegert (commented):
Looks good so far, would like some clarification about the default behaviour

shellscape:
Thanks for the PR. It looks like there are some tests breaking in CI that will need to be addressed, and we'll need to know if these changes proposed will be breaking changes for users (you didn't select an option in the template)

packages/node-resolve/src/index.js
```diff
@@ -249,13 +249,15 @@ export function nodeResolve(opts = {}) {
...
-          return null;
+          return false;
         } else if (importeeIsBuiltin && preferBuiltins) {
           if (!isPreferBuiltinsSet) {
             this.warn(
               `preferring built-in module '${importee}' over local alternative at '${resolved}', pass 'preferBuiltins: false' to disable this behavior or 'preferBuiltins: true' to disable this warning`
             );
           }
+          return false;
+        } else if (jail && resolved.indexOf(normalize(jail.trim(sep))) !== 0) {
```
tjenkinson, line 260: oops didn't mean to duplicate this line

packages/node-resolve/README.md
tjenkinson: The plugin behaves a bit differently when it's unset. I was confused by this because I though it would default to either the `true` or `false` behaviour, but it's essentially a third option, so this was to try and make that clear.

The way these 2 variables are used makes the third behaviour
```js
  const isPreferBuiltinsSet = options.preferBuiltins === true || options.preferBuiltins === false;
  const preferBuiltins = isPreferBuiltinsSet ? options.preferBuiltins : true;
```

I'd be happy for it to be an explicit option. Are you suggesting that `unset` should become `auto`? or something else?

tjenkinson: Looking at the code again now I'm not actually sure if the line I added here is correct.  It's quite confusing to follow through the logic

lukastaegert: Your example looks very much like the `default` for `preferBuiltins` would be `true`, just with a warning about it.

tjenkinson: Yeh removing this comment. Opened another PR to simplify how we handle `preferBuiltins` a bit

https://github.com/rollup/plugins/pull/637

lukastaegert (approved):
Looks good to me, looking forward to this one. I would still treat it as a breaking change, though, due to the changed default behaviour, even though it is likely a change very much in the user's interest.

shellscape:
thanks!

Issue #630, opened 2020-11-02
improvement to @rollup/plugin-html docs


simonxabris:
- Rollup Plugin Name: @rollup/plugin-html
- Rollup Plugin Version: 0.2.0

Documentation Is:

- [ ] Missing
- [ ] Needed
- [x] Confusing
- [ ] Not Sure?

### Please Explain in Detail...

The docs say that the `template` option receives an object that has a `files` property (this is correct up to this point) that is an `Array` of `AssetInfo` or `ChunkInfo`, which in my experience is not the case. The `files` property actually contains a `js`, `css` and `map` keys which are of type `Array<AssetInfo | ChunkInfo>`.

### Your Proposal for Changes

My proposal is pretty simple, It should be explained in the docs that `files` is an object with the following keys: `js`, `css`, `map` and they are of type `Array<AssetInfo | ChunkInfo>`. If you think this is OK I'll be happy to provide a PR with these changes!

shellscape:
Reopening since we have a PR in progress for this.

PR #631, opened 2020-11-02
docs(html): clarify the template option in the docs for html plugin

simonxabris:

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

#630 

### Description

This PR updates the docs for `@rollup/plugin-html` clarifying the what the `template` options receives as an argument. I explained it more in the above mentioned issue. If you think the wording is not clear enough please make suggestions and I'll make the necessary changes!


shellscape:
Thanks for opening a PR. This is one of those rare times that we need a test to validate the documentation change. We won't ask you to do that, but I will ask you to be patient on getting this PR merged until we've added a test to verify the structure of the parameters.

simonxabris:
Honestly I'd be happy to contribute the tests as well if you could give me some pointers about what you would expect to see. But if it's something you want to do yourself (or someone on the rollup team) I understand that too.

shellscape:
Cool, I just didn't want to put that burden on you. Basically what we need a test to do is snapshot the arguments for both types of array entries in the template handler. That'll be our sanity test in case the upstream API changes (in a new version of Rollup) or the function signature is changed.

simonxabris:
Thanks, I'll try to push the changes as soon as I can.

simonxabris:
I think this should be on hold until this is resolved: https://github.com/rollup/plugins/pull/634#discussion_r516922360

simonxabris:
So I've did some testing and here's what I found. The `files` array that the `template` function receives will be an object, where the keys are the file extensions of the different types of assets the will be written to disk and its value is an array of assets. So for example it can have keys such as `js`, `map`, if you use some kind of plugin to import css, it will have a `css` key, if you use the `wasm` plugin it will have a `wasm` key.

I'll push a new commit that updates the docs according to what I've found.

shellscape:
@simonxabris checking in on this one

shellscape:
Closing as abandoned.

Issue #626, opened 2020-10-30
[TypeScript] Build broken due to type issues


apfelbox:
- Rollup Plugin Name: TypeScript
- Rollup Plugin Version: 6.1.0
- Rollup Version: 2.32.1
- Typescript Version: 4.0.5
- Operating System (or Browser): macOS 
- Node Version: 14.14.0
- Link to reproduction _(⚠️ read below)_: https://github.com/apfelbox/rollup-plugins-issue-626

Using the plugin in a TypeScript build fails. So not the rollup compilation fails, but building the plugin with typescript itself fails.

### Expected Behavior
The plugin should build.

### Actual Behavior
It fails:

```
node_modules/.pnpm/@rollup/plugin-typescript@6.1.0_dc152661769777901a904be96994241b/node_modules/@rollup/plugin-typescript/types/index.d.ts:38:73 - error TS2344: Type 'CustomTransformers[T]' does not satisfy the constraint 'any[]'.
  Type '(TransformerFactory<SourceFile> | CustomTransformerFactory)[] | (CustomTransformerFactory | TransformerFactory<SourceFile | Bundle>)[] | undefined' is not assignable to type 'any[]'.
    Type 'undefined' is not assignable to type 'any[]'.

38 type StagedTransformerFactory<T extends TransformerStage> = ElementType<CustomTransformers[T]>;
                                                                           ~~~~~~~~~~~~~~~~~~~~~

node_modules/.pnpm/@rollup/plugin-typescript@6.1.0_dc152661769777901a904be96994241b/node_modules/@rollup/plugin-typescript/types/index.d.ts:51:20 - error TS2304: Cannot find name 'Program'.

51   factory(program: Program): StagedTransformerFactory<T>;
                      ~~~~~~~

node_modules/.pnpm/@rollup/plugin-typescript@6.1.0_dc152661769777901a904be96994241b/node_modules/@rollup/plugin-typescript/types/index.d.ts:57:24 - error TS2304: Cannot find name 'TypeChecker'.

57   factory(typeChecker: TypeChecker): StagedTransformerFactory<T>;
                          ~~~~~~~~~~~


Found 3 errors.
```

shellscape:
Thanks for opening an issue. Citing the issue template:

>   🚨 Issues WITHOUT a valid reproduction WILL BE CLOSED!
> 
>   Please provide one by:
>   1. Using the REPL.it plugin reproduction template at https://repl.it/@rollup/rollup-plugin-repro
[...]

PR #633, opened 2020-11-03
fix(typescript): add missing imports

apfelbox:
## Rollup Plugin Name: TypeScript

**This PR contains:**

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

**Are tests included?**

- [x] no

**Breaking Changes?**

- [x] no

I have no idea how to test this, as we would need to run the build with both versions of TypeScript and I don't know if sth like this is already set up for this repository).

**List any relevant issue numbers:**
* Fixes part of https://github.com/rollup/plugins/issues/626

### Description

These two imports were missing and fail the TypeScript build of the Typescript rollup plugin with TypeScript 4.0+
Not sure if they failed for ^3.9 as well.

I tested this change with both TypeScript 3.9.7 and 4.0.5 and it works.
Doesn't seem to fix the last issue of https://github.com/rollup/plugins/issues/626 though.


apfelbox:
The second issue is here:

```typescript

type ElementType<T extends Array<any>> = T extends (infer U)[] ? U : never;
                           ^^^^^^^^^^

type StagedTransformerFactory<T extends TransformerStage> = ElementType<CustomTransformers[T]>;
                                                                        ^^^^^^^^^^^^^^^^^^^^^
```

Because as every property of `CustomTransformers` can be undefined, `CustomTransformers[T]` is the `defined type|undefined`.

And therefore the type tries to assign `undefined` to `T extends any[]`, which doesn't work.

There are two possible fixes, but I am (semantically) not sure, which one is the better fit.

### Solution 1: allow undefined in `ElementType`

```diff
- type ElementType<T extends Array<any>> = T extends (infer U)[] ? U : never;
+ type ElementType<T extends (Array<any>|undefined)> = T extends (infer U)[] ? U : never;
```

### Solution 2: making all properties required, before fetching them

```diff
- type StagedTransformerFactory<T extends TransformerStage> = ElementType<CustomTransformers[T]>;
+ type StagedTransformerFactory<T extends TransformerStage> = ElementType<Required<CustomTransformers>[T]>;
```

NotWoods (approved):

apfelbox:
If somebody can give me feedback on the question in https://github.com/rollup/plugins/pull/633#issuecomment-721250077, I can add it (and we can get this merged)

shellscape:
@apfelbox I'm probably not as learned as you with TS, but could using `ElementType<Partial<CustomTransformers[T]>>` work here?

apfelbox:
@shellscape I don't think that that would work. `Partial` makes every property in an interface optional (e.g. `{test: string}` -> `{test?: string}`), but they already are all optional (they all already have a `?:`).

That is the main problem, actually: as they are optional, the type of the index access in `CustomTransformers[T]` can be `undefined`, which is then in turn not accepted in the definition of `ElementType` itself.

shellscape:
@apfelbox does #647 resolve the issue on your end? Seems to be a possible duplication of effort.

apfelbox:
@shellscape yes, I the meantime #647 popped up and seems to be fixing the same issues.
However I would argue (and have in an comment over there) that one of the fixes is wrong.

fernandopasik:
Is this PR the one that fixes this? https://github.com/rollup/plugins/pull/647

PR #632, opened 2020-11-02
feat(html): add type declaration to html package

simonxabris:
## Rollup Plugin Name: `@rollup/plugin-html`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] yes (_bugfixes and features will not be merged without tests_)
- [x] no

Breaking Changes?

- [ ] yes (_breaking changes will not be merged unless absolutely necessary_)
- [x] no

If yes, then include "BREAKING CHANGES:" in the first commit message body, followed by a description of what is breaking. 

List any relevant issue numbers:

### Description

I've added types to the `html` plugin because while using it I saw that they were missing. I've opened another PR about the the docs (#631) and @shellscape said that it needed tests to verify the changes so I'll assume this one needs tests as well. I'll be happy to add them if you could give some hint about what would you like to see in the tests.

PR #634, opened 2020-11-03
refactor(html): refactor to typescript

simonxabris:

This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

As per our discussion with @shellscape in #632 I've refactored the `html` plugin to use typescript.

The latter is now solved and not an issue anymore since @NotWoods solved it with https://github.com/rollup/plugins/pull/634#issuecomment-721418083

However there is one issue that might count as a breaking change. The original package was authored as `cjs` and exported the html plugin as the "default" `module.exports = html` and the `makeHtmlAttributes` as a "named" export `module.exports.makeHtmlAttributes = makeHtmlAttributes`.

As I understand this behavior cannot be reproduced with rollup and esm (https://rollupjs.org/guide/en/#outputexports), because if you use `export default` and `export const ....` in the same package and then later want to use it in a `cjs` environment you would have to import it like this: `const { default } = require('package');`.

What I've done here is made both the `html` and the `makeHtmlAttributes` named exports, because I didn't know what else to do.

Is this actually a breaking change which would prevent this from getting merged or can it be solved somehow?

Thanks for the help in advance!


packages/html/src/index.ts
```diff
@@ -1,20 +1,36 @@
...
-    result[extension] = (result[extension] || []).concat(file);
+
+    switch (extension) {
+      case 'css':
+      case 'js':
+      case 'map':
+        result[extension] = (result[extension] || []).concat(file);
+        break;
+      default:
+        // TODO: verify if this branch can be reached or not.
```
simonxabris, line 25: I wanted to ask a question here because I don't have the rollup knowledge to answer this one. The js code here just grabbed the file extension and put it on the result object without checking what it actually is. Can it happen that rollup emits files that are not `js`, `css` or `map`? For example `wasm` or an `svg` or something?

Because if it can than my question is do we want those on the `files` object and if yes, what type of files can we expect here so I can expand the typing and change the implementation so they actually end up on the `files` object.

simonxabris:
Hmm, It's strange that the lint fails in CI, when I run the same command (`pnpm run ci:lint --filter ./packages/html`) on my machine it runs successfully.

NotWoods:
For CJS exports, we set up a trick in the wasm plugin that you can reuse here. See the [shared config](https://github.com/rollup/plugins/blob/master/shared/rollup.config.js#L17).

simonxabris:
Oh, thanks I didn't know there was a shared config because I was looking at the `typescript` package as an example and that does not use it. I've just updated to the shared config and it solves the problem, so thanks again for that!

shellscape:
@simonxabris the issue with the linting is a sneaky one - because you converted from js to ts, and the tests are trying to import from `..`, and package.json specifies `main`, that means that the ts has to be built before linting on the `test` directory takes place. A few other packages like `alias` have this accounted for like so: `    "ci:lint": "pnpm run build && pnpm run lint",`

simonxabris:
ohh 🤦‍♂️ I should've caught that. Thanks a lot for the help @shellscape!

I left a comment on one of files where I was not sure what the functionality should be, please let me know what you think and I'll make the appropriate changes as soon as I can. I think if that is cleared up than this this PR is pretty close to done.

packages/html/src/index.ts
shellscape: It's a good question - but not one to be tackled during a refactor. Let's focus on parity here and we can open a new issue to discuss how this should be handled moving forward. The intent behind this plugin is to setup some very basic default templating, and anything beyond that, empower the user to manipulate what they need for a more custom or nuanced solution. We don't want to be in the business of anticipating a user's every move. That's how things get to an unmaintainable state.

packages/html/types/index.d.ts
```diff
@@ -0,0 +1,32 @@
+import { Plugin, OutputChunk, OutputAsset, OutputBundle } from 'rollup';
+
+export interface RollupPluginHtmlOptions {
```
shellscape, line 3:
```suggestion
export interface RollupHtmlOptions {
```

Following existing convention, we don't need `Plugin` in the type names.

packages/html/types/index.d.ts
```diff
@@ -0,0 +1,32 @@
...
+  title?: string;
+  attributes?: Record<string, any>;
+  fileName?: string;
+  meta?: Record<string, any>[];
+  publicPath?: string;
+  template?: (templateOptions: RollupPluginHtmlTemplateOptions) => string;
+}
+
+export interface RollupPluginHtmlTemplateOptions {
```
shellscape, line 12:
```suggestion
export interface RollupHtmlTemplateOptions {
```

packages/html/src/index.ts
simonxabris: You're right, I have a PR open that aims to update the docs around this, I think I'll throw together a rollup config that let's me import `svg`s `jpg`s and various file formats, see what ends up in the `files` object and then update that PR with the appropriate documentation changes.

In the meantime I've reverted this back to its original form.

shellscape:
@NotWoods please have a looksie over this. I don't see anything that would prevent merge.

NotWoods (approved):

shellscape:
thanks!

PR #647, opened 2020-11-13
fix(typescript): type declarations

geigerzaehler:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

~~If yes, then include "BREAKING CHANGES:" in the first commit message body, followed by a description of what is breaking.~~

### Description

Before, using type definitions from this package without the `skipLibCheck` option for TypeScript would result in type checking errors. We fix this and to ensure that the types stay correct we have the checked as part of the build process by disabling the `skipLibCheck` option.

### DIFF
--- a/packages/typescript/types/index.d.ts
+++ b/packages/typescript/types/index.d.ts
@@ -38,7 +38,7 @@ export interface RollupTypescriptPluginOptions {
   transformers?: CustomTransformerFactories;
 }
 
-type ElementType<T extends Array<any>> = T extends (infer U)[] ? U : never;
+type ElementType<T extends Array<any> | undefined> = T extends (infer U)[] ? U : never;
 
 export type TransformerStage = keyof CustomTransformers;
 type StagedTransformerFactory<T extends TransformerStage> = ElementType<CustomTransformers[T]>;


shellscape:
Thanks for opening this PR 🎉 

We've got one similar in #633 so I'd like to give the author of that one time to weigh in. Hang in there, we'll get this resolved soon :)

packages/typescript/types/index.d.ts
```diff
@@ -41,7 +41,7 @@ export interface RollupTypescriptPluginOptions {
 type ElementType<T extends Array<any>> = T extends (infer U)[] ? U : never;
 
 export type TransformerStage = keyof CustomTransformers;
-type StagedTransformerFactory<T extends TransformerStage> = ElementType<CustomTransformers[T]>;
+type StagedTransformerFactory<T extends TransformerStage> = ElementType<CustomTransformers[]>;
```
apfelbox, line 44: This seems to be an error. This change means that it isn't the list of a single transform step anymore, but an array of an array of transforms.

The type of `CustomTransformers` is:

```typescript
export interface CustomTransformers {
    /** Custom transformers to evaluate before built-in .js transformations. */
    before?: (TransformerFactory<SourceFile> | CustomTransformerFactory)[];
    /** Custom transformers to evaluate after built-in .js transformations. */
    after?: (TransformerFactory<SourceFile> | CustomTransformerFactory)[];
    /** Custom transformers to evaluate after built-in .d.ts transformations. */
    afterDeclarations?: (TransformerFactory<Bundle | SourceFile> | CustomTransformerFactory)[];
}
```

So `CustomTransformers[T]` already is an array. Making another array from it, removes the error but is probably semantically incorrect.

See https://github.com/rollup/plugins/pull/633#issuecomment-721250077 for a discussion of alternative solutions.

geigerzaehler: I wasn’t quite sure what was happening here so I just tried to make it work. I’ve used the first solution you suggested to fix the issue.

apfelbox:
@shellscape this one looks ready as well and should fix all TypeScript errors 👍

packages/typescript/tsconfig.json
```diff
@@ -1,4 +1,7 @@
 {
   "extends": "../../tsconfig.base.json",
-  "include": ["src/**/*", "types/**/*"]
+  "include": ["src/**/*", "types/**/*"],
+  "compilerOptions": {
+    "skipLibCheck": false
```
benmccann, line 5: I think this is the default, so you shouldn't really need to change it here

https://www.typescriptlang.org/docs/handbook/compiler-options.html

benmccann:
> We fix this and to ensure that the types stay correct we have the checked as part of the build process by disabling the skipLibCheck option

I don't understand how disabling a check would help prevent regressions. What I normally do is rename the `.d.ts` file to `.ts` and put it in the source directory and then let `@rollup/plugin-typescript` generate the `.d.ts` files

packages/typescript/tsconfig.json
geigerzaehler: You’re correct, it is the default for the compiler. But the project `tsconfig` [sets it to `true`](https://github.com/rollup/plugins/blob/a1746415fd032436f23f134a09b1b9cfabf7ceff/tsconfig.base.json#L12). I tried to be unintrusive and not change it for the whole project.

geigerzaehler:
> I don't understand how disabling a check would help prevent regressions. What I normally do is rename the `.d.ts` file to `.ts` and put it in the source directory and then let `@rollup/plugin-typescript` generate the `.d.ts` files

Enabling the lib check will also typecheck the `.d.ts` that is committed to source and force it to be correct.

I agree with you that normally you would want to avoid hand-written declaration files and instead have the compiler generate them for you. But I don’t know why this decision was made for the repo in the first place and I tried to keep the changes to a minimum.

packages/typescript/tsconfig.json
benmccann: Ah, I missed that. Though it looks like we can remove it in the base config without any additional changes, so that might be the better way to go: https://github.com/rollup/plugins/pull/700

benmccann:
> Enabling the lib check will also typecheck the .d.ts that is committed to source and force it to be correct.

Will it? If that's true then this PR isn't necessary because everything builds and all tests pass with `skipLibCheck` removed in https://github.com/rollup/plugins/pull/700

geigerzaehler:
> Will it? If that's true then this PR isn't necessary because everything builds and all tests pass with `skipLibCheck` removed in #700

Great find. I don’t understand the build and check system for this project and I assumed that it would do something that is equivalent to `tsc -p packages/typescript`. Running this command fails for me without the fix but I does not seem to be run on CI so I guess that is something that should be enabled.

packages/typescript/tsconfig.json
fernandopasik: If needed, could we roll back this change in order to merge the change for the type that is breaking in https://github.com/rollup/plugins/issues/626 ?

benmccann: I think the thing to do would be to merge https://github.com/rollup/plugins/pull/700 instead of changing this file, but perhaps @shellscape could weigh in on whether we can merge that one

fernandopasik: #700 is merged, please @geigerzaehler can you revert this change?

shellscape:
@geigerzaehler we merged #700, could you take another look at this?

geigerzaehler:
> @geigerzaehler we merged #700, could you take another look at this?

Cool. I’ve updated the branch and removed `skipLibCheck: false`. So this PR now fixes the issue but the CI still doesn’t not do type checking so this might break again. Since I’m not familiar with the CI setup it would make sense to tackle this separately.

fernandopasik:
Looking at https://github.com/rollup/plugins/blob/master/.circleci/config.yml I don't see any task to do typescript type checks

Each project in the monorepo has the task `test:ts` which does run tsc and check the types

fernandopasik (approved):
In any case that could be an improvement for the CI.
This PR scope is to fix a type issue

shellscape:
@fernandopasik that's correct, plugins should run type checking independently. since the typescript plugin doesn't have that, I'll make a note to add it following merging of this PR. thanks for pointing that out.

fernandopasik:
I'll create an issue and attempt a PR to resolve it.

fernandopasik:
https://github.com/rollup/plugins/issues/779

PR #653, opened 2020-11-17
fix(typescript): normalize returned module ids

### LINKED ISSUE #643: Typescript plugin produces duplicate output when used with multiple entry points on Windows
- Rollup Plugin Version: 6.0.0
- Rollup Version: 2.32.0
- Operating System (or Browser): Windows
- Node Version: 12.16
- Link to reproduction _(⚠️ read below)_: https://github.com/antross/rollup-plugin-typescript-repro

This occurs when using multiple entry points in which one references code from the other.

### Expected Behavior
In the bundled output the second entry point imports content from the first to avoid duplicating code. This is what rollup normally does when this plugin isn't present.
```js
import { one } from './one.js';

function two() {
    one();
    console.log('two');
}

export { two };
```

### Actual Behavior
In the bundled output the second entry point inlines content from the first, duplicating code (since the first entry point also is generated and still has this code).
```js
function one() {
    console.log('one');
}

function two() {
    one();
    console.log('two');
}

export { two };
```

### Additional Information
Removing @rollup/plugin-typescript from the plugins array eliminates the repro.

Making a quick local edit to wrap the return value from the plugin's `resolveId` method in a call to `path.normalize` appears to fix the issue, suggesting rollup is expecting an exact string match on the ID. Since the plugin is returning a POSIX normalized path even on Windows, this would cause the string comparison to fail if rollup is expecting an OS normalized path.:
https://github.com/rollup/plugins/blob/b12acfa910905d2cf2ec486be9f596de4de151bf/packages/typescript/src/index.ts#L101

Apart from fixing this plugin, it may be worth making rollup itself more flexible to handle paths with different types of separators than the host OS (so other plugins don't suffer from the same issue). Curious to hear your thoughts. Thanks!

antross:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: #643

### Description

Fix #643 by normalizing module ids to use host os separators before returning to rollup.

I initially opened this PR with just a failing test case to confirm the failure in CI before adding the fix.

### DIFF
--- a/packages/typescript/src/index.ts
+++ b/packages/typescript/src/index.ts
@@ -98,7 +98,7 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
 
       if (resolved) {
         if (resolved.extension === '.d.ts') return null;
-        return resolved.resolvedFileName;
+        return path.normalize(resolved.resolvedFileName);
       }
 
       return null;
--- a/packages/typescript/test/fixtures/normalize-ids/one.js
+++ b/packages/typescript/test/fixtures/normalize-ids/one.js
@@ -1,0 +1,4 @@
+/* eslint-disable */
+export function one() {
+    console.log('one');
+}
--- a/packages/typescript/test/fixtures/normalize-ids/tsconfig.json
+++ b/packages/typescript/test/fixtures/normalize-ids/tsconfig.json
@@ -1,0 +1,7 @@
+{
+  "compilerOptions": {
+    "allowJs": true,
+    "checkJs": true,
+  },
+  "include": ["*.js", "**/*.js"]
+}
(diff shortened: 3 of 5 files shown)


antross:
Example failing test on Windows here: https://github.com/rollup/plugins/pull/653/checks?check_run_id=1414099436

shellscape (approved):
Looks good to me. I'll let @NotWoods weigh in before merging.

NotWoods (approved):
LGTM as well

shellscape:
thanks!

Issue #282, opened 2020-03-25
Expected fileName to be present in command line


madhudskumar:
- Rollup Plugin Name: @rollup/plugin-typescript
- Rollup Plugin Version: 4.0.0
- Rollup Version: 1.23.1
- Operating System (or Browser): mac
- Node Version: 10.16.3

### How Do We Reproduce?
use the same plugin to bundle any .ts/.tsx file it throws an error
```
components/index.ts → dist/components/index.js, dist/components/index.es.js...
[!] (plugin typescript) Error: Could not load /Users/madhuds/proj/uie/components/index.ts: Debug Failure. False expression: Expected fileName to be present in command line
Error: Could not load /Users/madhuds/proj/uie/components/index.ts: Debug Failure. False expression: Expected fileName to be present in command line
    at Object.getOutputFileNames (/Users/madhuds/proj/uie/node_modules/typescript/lib/typescript.js:92435:18)
    at findTypescriptOutput (/Users/madhuds/proj/uie/node_modules/@rollup/plugin-typescript/dist/index.js:389:33)
    at Object.load (/Users/madhuds/proj/uie/node_modules/@rollup/plugin-typescript/dist/index.js:508:28)
    at Promise.resolve.then (/Users/madhuds/proj/uie/node_modules/rollup/dist/shared/node-entry.js:13117:25)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

### Expected Behavior
should ideally find the file

### Actual Behavior
unable to find the file

shellscape:
Thanks for opening an issue. Citing the issue template:

>   Issues without minimal reproductions will be closed! Please provide one by:
>   1. Using the REPL.it plugin reproduction template at https://repl.it/@rollup/rollup-plugin-repro
>   2. Provide a minimal repository link (Read https://git.io/fNzHA for instructions).
>      These may take more time to triage than the other options.
>   3. Using the REPL at https://rollupjs.org/repl/

Please add a reproduction and we'll be happy to triage further.

etylsarin:
hi @madhudskumar 
could you find a fix after all? I'm having the very same issue.

fregante:
In my case this was because I didn't add `@rollup/plugin-node-resolve`

The full error was:

```
$ npx rollup index.ts -c

index.ts → index.js, module.js...
(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
select-dom (imported by index.ts, utils.ts)
[...]

PR #657, opened 2020-11-22
fix(typescript): pick up new files in watch mode

koenkivits:
See several comments in #282

## Rollup Plugin Name: `{TypeScript}`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

* fixes part of #282

### Description

The TypeScript plugin doesn't pick up new files when in watch mode. It crashes with an error like:

```
Expected fileName to be present in command line
```

The problem has been reported by some users in #282, but the error can have multiple causes and not all comments in that issue are related to this specific problem.

This error happens here because the plugin determines all file names when it is set up, and it doesn't update the list afterwards. This PR makes the plugin detect new TypeScript files and adds them to the list of file names.

The fix itself is only 3 lines of code; the majority of this PR is the test I added. I noticed there were no tests yet using watch mode so I didn't have a template to go by, so please let me know if I can improve anything here. :)


koenkivits:
Writing the fix wasn't too hard, fixing it to work on my Windows machine was also not a lot of work. But getting `rollup.watch` to work in Windows CI seems tricky. 😅 If anyone has any ideas I'd love to hear them, but otherwise I'll make another attempt later.

packages/typescript/test/test.js
```diff
@@ -1129,3 +1130,68 @@ function fakeTypescript(custom) {
...
+  watcher.close();
+
+  const code = fs.readFileSync('fixtures/watch/dist/main.js');
+  const usage = code.includes('Is it me');
+  t.true(usage, 'should contain usage');
+});
+
+function waitForWatcherEvent(watcher, eventCode) {
+  return new Promise((resolve, reject) => {
+    watcher.on('event', function handleEvent(event) {
```
NotWoods, line 1187: You can remove the `off` calls with `once`:
```suggestion
    watcher.once('event', function handleEvent(event) {
```

NotWoods (approved):
Looks good to me. I think we should mark this as a breaking change just in case, since the loaded files will be different.

packages/typescript/test/test.js
koenkivits: Not in this case, unfortunately. :slightly_smiling_face: 

`once` removes the listener after the first event of a type is triggered, but this function doesn't listen for the first event of a type. Rollup issues all watcher events as events of type `event` -- if I used `once` here then it would remove the listener directly after the initial (`BUNDLE` / `BUNDLE_START`) events.

koenkivits:
I managed to fix the tests for Windows CI by updating the watched files instead of overwriting them. All green now!

@NotWoods thanks for reviewing this! I'm not sure if this is a breaking change (TS would always error if the file name wasn't in the parsed config list of file names), but this is the first time I'm contributing to a project like Rollup so it's hard for me to imagine all the scenarios that could go wrong. I'd just be happy if this got merged eventually. :smile:

PR #661, opened 2020-11-26
Add a comparison between TypeScript plugins

benmccann:

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [X] documentation
- [ ] other

Are tests included?

- [X] no

Breaking Changes?

- [X] no

### Description

Adds a comparison between Rollup TypeScript plugins. Having multiple options can make it difficult for users to know which to choose. Providing a bit of context makes it easier to choose the appropriate solution


shellscape:
Thanks for putting this together. This is something that I would rather see on https://github.com/rollup/awesome or perhaps Rollup's website in a new section. We get a lot of "this one sucks, use this other one it's awesome!" with regard to the typescript plugin, and I desperately want to avoid anything in the docs here that could be misconstrued as a comparison, good or bad. It was our hope that the "official" nature of the plugin being scoped to `@rollup` would lend more credence and direction to which users should choose by default, and then seek alternatives should that not work out. Because of the afore mention, I'm going to pass on this one. Though other maintainers may chime in an overrule me :smile:

PR #673, opened 2020-12-01
chore(typescript): fix TypeScript warnings

benmccann:

This PR contains:

- [ ] bugfix
- [ ] feature
- [X] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [X] no

Breaking Changes?

- [X] no

### Description

This fixes the red squiggles that show up in VS Code when working on the TypeScript plugin


packages/typescript/src/index.ts
```diff
@@ -51,7 +51,7 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
           formatHost,
           resolveModule,
           parsedOptions,
-          writeFile(fileName, data) {
+          writeFile(fileName: string, data: string) {
```
NotWoods, line 54: These type annotations shouldn't be necessary as they're provided by the `createWatchProgram` function. Same goes for `parsedOptions.fileNames.forEach`.

NotWoods (changes requested):
I'm happy to include `@types/node` but I don't think the function annotations are needed. It might be a sign of a missing dependency in your local copy.

packages/typescript/src/index.ts
benmccann: You're right. It was because of the way the TypeScript types were being imported. I'm not quite sure why it was being done the way it was. I've gone ahead and updated it, but let me know if there's a reason to leave it as it was and I can revert it

packages/typescript/README.md
```diff
@@ -242,8 +242,8 @@ export default {
   input: './main.ts',
   plugins: [
     typescript({ module: 'CommonJS' }),
-    commonjs({ extensions: ['.js', '.ts'] }), // the ".ts" extension is required
-  ],
+    commonjs({ extensions: ['.js', '.ts'] }) // the ".ts" extension is required
```
NotWoods, line 245: It looks like the README got formatted unintentionally

packages/typescript/src/diagnostics/emit.ts
```diff
@@ -13,7 +14,7 @@ export function emitDiagnostic(
   ts: typeof import('typescript'),
   context: PluginContext,
   host: DiagnosticsHost,
-  diagnostic: import('typescript').Diagnostic
+  diagnostic: Diagnostic
```
NotWoods, line 17: This was written before `import type` was a thing, so thanks for moving it over :)

packages/typescript/README.md
benmccann: Is it necessary to change it back? I suppose I could add the README to `.prettierignore`, but these seem like reasonable changes and I'm not sure we'd want to permanently avoid formatting the README?

NotWoods: Yes this needs to be reverted, because it's inconsistent with the formatting used by the rest of the project.

NotWoods (approved):

packages/typescript/README.md
benmccann: Hmm. It's just inconsistent because the other docs haven't been reformated yet. The others would get updated as those projects are formatted again with the latest version of prettier and eslint config

I've reverted the changes to the README, but it required turning off linting of the README. That doesn't really seem like the right solution to me, but I'm not quite sure how else I would avoid it being reformatted.

NotWoods: The linting in the docs doesn't match what's used in our code. For example, single-argument arrow functions have no parenthesis in your linted version. That indicates that your local checkout might not be set up correctly - for example your editor is running formatting separate from our prettier config.

NotWoods: Taking a closer look it seems like we're use a different config for the README? @shellscape can you weigh in on what format we should be using?

benmccann: I was able to drop the change to the README from this PR. I opened https://github.com/rollup/plugins/pull/676 with more details about this issue

packages/typescript/test/test.js
```diff
@@ -1085,50 +1085,48 @@ test('supports custom transformers', async (t) => {
...
+      };
+    },
+
+    getOutputFileNames(_, id) {
+      return [id.replace(/\.tsx?/, '.js')];
+    },
+
+    // eslint-disable-next-line no-undefined
+    getTsBuildInfoEmitOutputFilePath: () => undefined,
+    ...custom
```
shellscape, line 1128: Do we have tests that cover `custom`? There are quite a few differences with `Object.assign` and `custom` so I want to make sure we're covered.

benmccann: I've reverted the changes to this file. I think the linting changes would better be handled as part of the discussion in https://github.com/rollup/plugins/issues/678

benmccann:
I've rebased this PR so it should be good to merge now

benmccann:
@shellscape any chance I could get a new release of the TypeScript plugin after you merge this? This one should be good to go, btw

shellscape:
I typically batch releases. I'll keep you posted.

Issue #287, opened 2020-03-28
Typescript plugin 4.0.0 Error: Unexpected token (Note that you need plugins to import files that are not JavaScript)


hasangenc0:
- Rollup Plugin Name: typescript
- Rollup Plugin Version: 4.0.0
- Rollup Version: 2.2.0
- Operating System (or Browser): macOsx
- Node Version: 10.16.3

### How Do We Reproduce?
```js
import tsPlugin from '@rollup/plugin-typescript';
import pkg from './package.json';

export default {
	input: 'src/index.tsx',
	output: [
		{
			file: pkg.main,
			format: 'cjs',
			sourcemap: true,
		},
		{
			file: pkg.module,
			format: 'es',
			sourcemap: true,
		},
	],
	plugins: [
		tsPlugin(),
	],
};

```

### Expected Behavior
````
src/index.tsx → dist/index.js, dist/index.es.js...
created dist/index.js, dist/index.es.js in 1.1s
````

### Actual Behavior
````
[!] Error: Unexpected token (Note that you need plugins to import files that are not JavaScript)
src/index.tsx (1:25)
1: export function hasan(num: number): number {
                            ^
````

I hope I'm doing something wrong because it is so bad to cant use ts plugin.
Ts plugin version 3.0.0 works perfect by the way.

shellscape:
Thanks for opening an issue. Citing the issue template:

>   Issues without minimal reproductions will be closed! Please provide one by:
>   1. Using the REPL.it plugin reproduction template at https://repl.it/@rollup/rollup-plugin-repro
>   2. Provide a minimal repository link (Read https://git.io/fNzHA for instructions).
>      These may take more time to triage than the other options.
>   3. Using the REPL at https://rollupjs.org/repl/

Please add a reproduction and we'll be happy to triage further.

JohanSchobben:
I have the same error and made a small project to reproduce the error [(see this project)](https://github.com/JohanSchobben/ts-rollup-reproduction)

JohanSchobben:
I switched back to @rollup/plugin-typescript version 3.1.1. In this version I don't have the problems. I use tsc to make the declarations.

shellscape:
Thanks for the reproduction @JohanSchobben

sssylvan:
try to create a .ts file in the root ...  
I resolved it by this ...

sssylvan:
Quick fix for now is to manually set your rootDir compiler option to "src".

talentlessguy:
[...]

PR #674, opened 2020-12-01
Build TypeScript plugin with TypeScript 4

benmccann:
## Rollup Plugin Name: all

This PR contains:

- [ ] bugfix
- [ ] feature
- [X] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [X] no

Breaking Changes?

- [X] no

I don't expect this would cause folks issues in upgrading. The only code changes I had to make were slight and internal to the plugins, so I don't think the external types will change

### Description

There are a number of open issues with the TypeScript plugin. I was poking around to see what would be required to fix some of them. Being on an older version of TypeScript makes things harder. E.g. it seems like there might be a bug in one of the TypeScript methods (https://github.com/rollup/plugins/issues/287#issuecomment-674552597), but I can't report issues to the TypeScript team unless using the latest version of TypeScript. I also was trying to use an API that I saw in `tsc` and appeared to be exported, but VS Code is telling me it doesn't exist and I wonder if being on an older version of TypeScript might be why.

### DIFF
--- a/packages/typescript/package.json
+++ b/packages/typescript/package.json
@@ -58,7 +58,8 @@
     "@rollup/plugin-commonjs": "^11.1.0",
     "@rollup/plugin-typescript": "^5.0.2",
     "buble": "^0.20.0",
-    "rollup": "^2.14.0"
+    "rollup": "^2.14.0",
+    "typescript": "^4.0.0"
   },
   "types": "types/index.d.ts",
   "ava": {
--- a/packages/typescript/package.json
+++ b/packages/typescript/package.json
@@ -59,7 +59,7 @@
     "@rollup/plugin-typescript": "^5.0.2",
     "buble": "^0.20.0",
     "rollup": "^2.14.0",
-    "typescript": "^4.0.0"
+    "typescript": "^4.1.2"
   },
   "types": "types/index.d.ts",
   "ava": {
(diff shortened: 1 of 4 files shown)


shellscape:
Appreciate the effort here. For PRs that will affect more than one plugin, the changes should really be discussed ahead of time. Looks like I'm going to need to update the contributing document to cover this. PRs that affect multiple plugins should be rare and require a special commit history and a regular merge commit versus squash commit - this is to facilitate an independent plugin publishing process (versus babel where they publish the world for each new version).

We should hold off on this right now. If you'd like to update just the typescript plugin, roll the other plugins back, and repurpose this branch and PR for only that plugin, then I can get behind that. I leave that choice up to you.

benmccann:
Ok. I will update this PR an open an issue to discuss the process of updating all plugins

shellscape:
@NotWoods this was rebased to only affect the Typescript plugin, so it'll need your approval.

packages/typescript/package.json
```diff
@@ -58,7 +58,8 @@
     "@rollup/plugin-commonjs": "^11.1.0",
     "@rollup/plugin-typescript": "^5.0.2",
     "buble": "^0.20.0",
-    "rollup": "^2.14.0"
+    "rollup": "^2.14.0",
+    "typescript": "^4.0.0"
```
NotWoods, line 62: Let's update to the latest version (4.1.2)

benmccann: done (though the `^` prefix means get the latest 4.x version in either case, so they are equivalent)

NotWoods (approved):

PR #682, opened 2020-12-02
test(typescript): code in src sub-directory

benmccann:

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [X] other

Are tests included?

- [ ] no

Breaking Changes?

- [X] no

### Description

Add a disabled test for https://github.com/rollup/plugins/issues/287


shellscape:
What's the benefit of adding a test that won't be run?

benmccann:
I had a couple goals:
- it makes it easier for someone to add a fix. All they have to do is create the fix and they don't have to write any test code. They'll just need to remove the `skip`
- it's possible the bug described in #287 is in the TypeScript library. It would make it easier for me to report to the TypeScript team to have a unit test that they can test against to see the issue

NotWoods (approved):
This is a common issue so I'm happy to have a test written for it. Thanks for the help!

packages/typescript/test/test.js
```diff
@@ -1181,6 +1181,32 @@ test.serial('picks up on newly included typescript files in watch mode', async (
   t.true(usage, 'should contain usage');
 });
 
+// https://github.com/rollup/plugins/issues/287. this fails with the message:
+// Unexpected token (Note that you need plugins to import files that are not JavaScript)
+test.serial.skip('works when code is in src directory', async (t) => {
```
shellscape, line 1186: Last ask: Please add a note as to why this test is skipped

benmccann: sure! done

Issue #574, opened 2020-09-15
WASM: eval prevents dead code elimination


arvigeus:
- Rollup Plugin Name: WASM
- Rollup Plugin Version: 5.1.0

### Expected Behavior / Situation

If no wasm code is used then the code generated from this plugin should be removed by `terser`

### Actual Behavior / Situation

Terser is unable to remove `_loadWasmModule` due to having `eval` in it

### Modification Proposal

https://github.com/rollup/plugins/blob/master/packages/wasm/src/index.ts#L66
Don't use eval here. Terser cannot remove the function: [source](https://github.com/terser/terser/issues/497#issuecomment-544469100)

shellscape:
Happy to review a PR to resolve this. I'd recommend using `Function` if possible

arvigeus:
Why the requires have to be in eval anyway?

I think the solution should be something like:

```js
var fs = (new Function( 'return require("fs")' )())
var path = (new Function( 'return require("path")' )())
```

I hope the tests will tell me if I am wrong.

arvigeus:
Did not work. Tests failed. I don't understand the code enough to know how to fix it.

arvigeus:
I would do it, but I need help

PR #685, opened 2020-12-03
fix(wasm): Remove `eval` calls and enable dummy sourcemap

featherbear:
## Rollup Plugin Name: `@rollup/wasm`

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [ ] yes
- [x] no

Breaking Changes?

- [x] no

List any relevant issue numbers: #574 

### Description

* Remove calls to `eval` which would raise warnings in rollup
* Add a dummy sourcemap to suppress warnings in rollup (See [@rollup/json/index.js#L23](https://github.com/rollup/plugins/blob/8c3a3a6eaa4dc301a4dcd6ea0b4efcc8243c9e12/packages/json/src/index.js#L23))
* Add eslint ignore statements to wasm test files to pass linting

### DIFF
--- a/packages/wasm/test/test.js
+++ b/packages/wasm/test/test.js
@@ -51,6 +51,7 @@ test('fetching WASM from separate file', async (t) => {
 
   global.result = null;
   global.t = t;
+  // eslint-disable-next-line global-require, import/no-dynamic-require
   require(outputFile);
 
   await global.result;
@@ -97,6 +98,7 @@ test('imports', async (t) => {
 });
 
 try {
+  // eslint-disable-next-line global-require
   const { Worker } = require('worker_threads');
   test('worker', async (t) => {
     t.plan(2);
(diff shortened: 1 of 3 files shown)


packages/wasm/test/test.js
```diff
@@ -51,6 +51,7 @@ test('fetching WASM from separate file', async (t) => {
 
   global.result = null;
   global.t = t;
+  // eslint-disable-next-line global-require, import/no-dynamic-require
```
benmccann, line 54: does it not work to use `import` here instead of `require`?

featherbear: If `import` is used, `t.plan(3)` needs to be changed to `t.plan(2)`.
Not sure if there will be any unexpected consequences as a result of that.

shellscape (approved):
LGTM

PR #688, opened 2020-12-04
fix(html): Template files should only include entry chunks

ianpurvis:
## Rollup Plugin Name: `html`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: #634

### Description

I ran into a regression in #634 where non-entry scripts are injected into the default html template. This can cause unexpected behavior especially if those scripts have side-effects. This PR fixes the file filter and also adds a guard in case rollup emitted any javascript asset files.


packages/html/src/index.ts
```diff
@@ -6,9 +6,7 @@ import { RollupHtmlOptions, RollupHtmlTemplateOptions } from '../types';
 
 const getFiles = (bundle: OutputBundle): RollupHtmlTemplateOptions['files'] => {
   const files = Object.values(bundle).filter(
-    (file) =>
-      file.type === 'chunk' ||
-      (typeof file.type === 'string' ? file.type === 'asset' : file.isAsset)
+    (file: any) => file.isEntry || !file.fileName.endsWith('.js')
```
shellscape, line 9: so this isn't exactly what we're after. we do want all files emitted in the build to be present in the result here.

shellscape:
Thanks for the PR! You're on the right path, but just off of it. I think the idea is good for default behavior as well, and while it is breaking, it'll make default behavior for others easier to understand when the build gets a little bigger. So good call.

Where you want to make the change though, is here: https://github.com/rollup/plugins/blob/a54d2a111458141a8298f8d7fb4505da63950ada/packages/html/src/index.ts#L41-L46

ianpurvis:
@shellscape Sounds good- actually, I looked at that approach too but went the other direction because I thought the behavior change was a mistake in https://github.com/rollup/plugins/commit/a1b2fc3e6df5b1630acd8d5e3dbeea96ceef5800. I'll update this PR 👍

packages/html/src/index.ts
ianpurvis: In that case, I think we can remove the filter altogether and rename `getFiles` to `groupFiles` ? What do you think?

ianpurvis: In the end I went with `getFilesByExtension`... Check out 233fe7c and let me know if that looks good. Thanks!

shellscape:
I appreciate that you're trying to make this easier for the use case on your end, but I don't think you're considering the flexibility needed for the userbase as a whole. I'd like the original `getFiles` to be reinstated, and the block I mentioned here https://github.com/rollup/plugins/pull/688#issuecomment-739445066 to filter for entry files like you had originally intended. That allows any custom template or consumer to get the full stack of build files as originally intended.

packages/html/src/index.ts
shellscape: that's not going to work for me. I commented on the main PR thread as to why

shellscape (changes requested):
I appreciate that you're trying to make this easier for the use case on your end, but I don't think you're considering the flexibility needed for the userbase as a whole. I'd like the original getFiles to be reinstated, and the block I mentioned here https://github.com/rollup/plugins/pull/688#issuecomment-739445066 to filter for entry files like you had originally intended. That allows any custom template or consumer to get the full stack of build files as originally intended.

ianpurvis:
Wait, doesn't removing the filter from `getFiles` allow the template function to get the full stack of files? My intention in https://github.com/rollup/plugins/pull/688/commits/233fe7ccbb7f685c7c4ef7b79f13033e4e8a49a0 was to allow all files to flow to the template function, moving the `isEntry` filter to the default template implementation only.

About the filter itself, I am using some assumptions about Rollup's `OutputAsset|OutputChunk` type. Assuming that `.type` must equal 'chunk' or 'asset', the filter logic from a1b2fc3 already seems to be a no-op:

```js
 (file) =>
      file.type === 'chunk' ||
      (typeof file.type === 'string' ? file.type === 'asset' : file.isAsset)
```

But maybe I am missing something- is that assumption invalid? Let me know, I am ready to tweak the PR as needed, just need more clarification...

shellscape:
@ianpurvis looks like this one has been abandoned. if you'd like to see it merged, please give this a look.

ianpurvis:
@shellscape Go ahead and close as needed... this has been dead a long time. Thanks!

PR #691, opened 2020-12-06
fix(image): use 'var' instead of  'const'

tjenkinson:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] yes
- [ ] no

Breaking Changes?

- [x] no

### Description
This updates the plugin to use `var` instead of `const`, because older browsers do not support `var`.

This is inline with rollup itself which defaults to `var` but has an opt-in with the `output.preferConst` option.

At first I thought we could respect the same option in this plugin, but it's an output option which means we are unable to use that. We could add a `preferConst` option to the plugin itself if wanted?


NotWoods (approved):
Most of our plugins just use `var`, I don't think we need to handle `preferConsts` right now.

PR #693, opened 2020-12-06
feat(node-resolve): support pkg imports and export array

LarsDenBakker:

This PR contains:

- [X] bugfix
- [X] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [X] no

List any relevant issue numbers: https://github.com/rollup/plugins/issues/670

### Description

To fix https://github.com/rollup/plugins/issues/670 I started reviewing the algorithm at https://nodejs.org/dist/latest-v15.x/docs/api/esm.html#esm_resolver_algorithm_specification and figured it would be better to implement that more closely. 

This PR re-implements the whole package exports feature, basing it on the spec. This fixes the reported bug, and also adds support for the package imports feature defined at https://nodejs.org/api/packages.html#packages_subpath_imports

### DIFF
--- a/packages/node-resolve/src/package/resolvePackageImports.js
+++ b/packages/node-resolve/src/package/resolvePackageImports.js
@@ -1,32 +1,7 @@
-/* eslint-disable no-await-in-loop */
-import path from 'path';
-import fs from 'fs';
 import { pathToFileURL } from 'url';
-import { promisify } from 'util';
 
-import { createBaseErrorMsg, InvalidModuleSpecifierError } from './utils';
+import { createBaseErrorMsg, findPackageJson, InvalidModuleSpecifierError } from './utils';
 import resolvePackageImportsExports from './resolvePackageImportsExports';
-
-const fileExists = promisify(fs.exists);
-
-function isModuleDir(current, moduleDirs) {
-  return moduleDirs.some((dir) => current.endsWith(dir));
-}
-
-async function findPackageJson(base, moduleDirs) {
-  const { root } = path.parse(base);
-  let current = base;
-
-  while (current !== root && !isModuleDir(current, moduleDirs)) {
-    const pkgJsonPath = path.join(current, 'package.json');
-    if (await fileExists(pkgJsonPath)) {
-      const pkgJsonString = fs.readFileSync(pkgJsonPath, 'utf-8');
-      return { pkgJson: JSON.parse(pkgJsonString), pkgPath: current, pkgJsonPath };
-    }
-    current = path.resolve(current, '..');
-  }
-  return null;
-}
 
 async function resolvePackageImports({
   importSpecifier,
--- a/packages/node-resolve/src/package/resolvePackageImports.js
+++ b/packages/node-resolve/src/package/resolvePackageImports.js
@@ -1,0 +1,71 @@
+/* eslint-disable no-await-in-loop */
+import path from 'path';
+import fs from 'fs';
+import { pathToFileURL } from 'url';
+import { promisify } from 'util';
+
+import { createBaseErrorMsg, InvalidModuleSpecifierError } from './utils';
+import resolvePackageImportsExports from './resolvePackageImportsExports';
+
+const fileExists = promisify(fs.exists);
+
+function isModuleDir(current, moduleDirs) {
+  return moduleDirs.some((dir) => current.endsWith(dir));
+}
+
+async function findPackageJson(base, moduleDirs) {
+  const { root } = path.parse(base);
+  let current = base;
+
+  while (current !== root && !isModuleDir(current, moduleDirs)) {
+    const pkgJsonPath = path.join(current, 'package.json');
+    if (await fileExists(pkgJsonPath)) {
+      const pkgJsonString = fs.readFileSync(pkgJsonPath, 'utf-8');
+      return { pkgJson: JSON.parse(pkgJsonString), pkgPath: current, pkgJsonPath };
+    }
+    current = path.resolve(current, '..');
+  }
+  return null;
+}
+
+async function resolvePackageImports({
+  importSpecifier,
+  importer,
+  moduleDirs,
+  conditions,
+  resolveId
+}) {
+  const result = await findPackageJson(importer, moduleDirs);
+  if (!result) {
+    throw new Error(createBaseErrorMsg('. Could not find a parent package.json.'));
+  }
+
+  const { pkgPath, pkgJsonPath, pkgJson } = result;
+  const pkgURL = pathToFileURL(`${pkgPath}/`);
+  const context = {
+    importer,
+    importSpecifier,
+    moduleDirs,
+    pkgURL,
+    pkgJsonPath,
+    conditions,
+    resolveId
+  };
+
+  const { imports } = pkgJson;
+  if (!imports) {
+    throw new InvalidModuleSpecifierError(context, true);
+  }
+
+  if (importSpecifier === '#' || importSpecifier.startsWith('#/')) {
+    throw new InvalidModuleSpecifierError(context, 'Invalid import specifier.');
+  }
+
+  return resolvePackageImportsExports(context, {
+    matchKey: importSpecifier,
+    matchObj: imports,
+    internal: true
+  });
+}
+
+export default resolvePackageImports;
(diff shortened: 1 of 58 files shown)


shellscape (approved):
Took me a while to digest all of it, but it looks good to me.

packages/node-resolve/src/package/resolvePackageImports.js
```diff
@@ -0,0 +1,49 @@
...
+import { pathToFileURL } from 'url';
+
+import { createBaseErrorMsg, InvalidModuleSpecifierError } from './utils';
+import resolvePackageImportsExports from './resolvePackageImportsExports';
+
+function findPackageJson(base) {
+  const { root } = path.parse(base);
+  let current = base;
+
+  while (current !== root && !current.endsWith('node_modules')) {
```
tjenkinson, line 12: I think the name of `node_modules` is configurable on an option?

packages/node-resolve/src/package/resolvePackageImports.js
```diff
@@ -0,0 +1,49 @@
...
+    const pkgJsonPath = path.join(current, 'package.json');
+    if (fs.existsSync(pkgJsonPath)) {
```
tjenkinson, line 14: Could we use the async version and make the function async?

packages/node-resolve/src/package/resolvePackageTarget.js
```diff
@@ -0,0 +1,111 @@
+/* eslint-disable no-await-in-loop */
+import { pathToFileURL } from 'url';
+
+import { isUrl, InvalidModuleSpecifierError, InvalidPackageTargetError } from './utils';
+
+function includesInvalidSegments(pathSegments) {
+  return pathSegments
+    .split('/')
+    .slice(1)
+    .some((t) => ['.', '..', 'node_modules'].includes(t));
```
tjenkinson, line 10: Same as other comment re node_modules

tjenkinson (commented):
This wasn’t a thorough review. Just a few things I noticed skimming through.

Looks good though

packages/node-resolve/src/package/resolvePackageImports.js
LarsDenBakker: Good point, made this configurable.

packages/node-resolve/src/package/resolvePackageImports.js
LarsDenBakker: Done

packages/node-resolve/src/resolveImportSpecifiers.js
```diff
@@ -178,7 +25,7 @@ async function resolveId({
   preserveSymlinks,
   useBrowserOverrides,
   baseDir,
-  moduleDirectories
+  moduleDirectories = ['node_modules']
```
tjenkinson, line 28: When would this be undefined? Don’t we set a default in index?

LarsDenBakker: You're right it's set in the index. I removd it.

guybedford (approved):
Incredible work, thanks again @LarsDenBakker.

ctavan:
@LarsDenBakker just out of curiosity: Did you check the efforts in https://github.com/browserify/resolve/pull/224 ? Could that be re-used?

LarsDenBakker:
@ctavan yes potentially. that PR has been open for a long time, and it was created after I started working on it here.

remorses:
Shouldn’t this logic lobe in the resolve package? Will you delete this code after resolve implements the exports logic?

LarsDenBakker:
Like I said above... maybe. I think the `resolve` package intends to implement commonjs-style resolve, not node-js es module style resolve. The rollup plugin also has some custom options, we need to see if these things can work together. But it would be better to have one implementation, also with webpack which also has it's own implementation.

Note that this PR is an upgrade of functionality that is already in the node-resolve plugin, a PR that I started back in august: https://github.com/rollup/plugins/pull/540. At the time I wasn't aware of the work being done in the `resolve` package for this. Now the original PR is merged, and I'm just making sure we implement the spec correctly.

packages/node-resolve/src/resolveImportSpecifiers.js
```diff
@@ -215,8 +62,33 @@ async function resolveId({
 
   let location;
 
-  const pkgName = getPackageName(importPath);
-  if (pkgName) {
+  const pkgName = getPackageName(importSpecifier);
```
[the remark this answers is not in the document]
LarsDenBakker: That's a good catch... but something we can add as a followup PR I think.

packages/node-resolve/src/package/resolvePackageTarget.js
```diff
@@ -0,0 +1,111 @@
...
+    for (const [key, value] of Object.entries(target)) {
+      if (key === 'default' || context.conditions.includes(key)) {
+        const resolved = await resolvePackageTarget(context, {
+          target: value,
+          subpath,
+          pattern,
+          internal
+        });
+
+        if (resolved) {
```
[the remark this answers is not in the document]
LarsDenBakker: I'm not sure what you mean, could you elaborate on this?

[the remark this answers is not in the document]
guybedford: I was mainly referring to this case:

```json
{
  "exports": {
    "./features/*": "./features/*.js",
    "./features/internal/*": null
  }
}
```

But I guess your code here will handle this case just fine.

So it's more about the sort of technical edge cases between conditional fallbacks and resolutions that happen to be `null`, since usually conditions will fall through to the next, while `null` is seen as an actual result value. But I don't think these really correspond to real world use cases so no stress.

ExE-Boss:
@LarsDenBakker **CommonJS**‑style `resolve` includes the package `exports` and `imports` fields, along with the self‑import when `exports` is specified:

- <https://nodejs.org/api/packages.html#packages_subpath_exports>
- <https://nodejs.org/api/packages.html#packages_subpath_imports>
- <https://nodejs.org/api/packages.html#packages_self_referencing_a_package_using_its_name>

LarsDenBakker:
@ExE-Boss that's correct, but esm imports have different resolve logic such as no magic `index.js` handling and requiring file extensions.

packages/node-resolve/src/package/resolvePackageTarget.js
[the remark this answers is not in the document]
LarsDenBakker: Ah I see, I hadn't considered the case when someone would put `null` in there. I've handled this case now, for compatibility :)

packages/node-resolve/src/resolveImportSpecifiers.js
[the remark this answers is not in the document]
LarsDenBakker: It looked pretty straightforward to add, so added this one in too :)

LarsDenBakker:
@shellscape I see that circle CI is failing on a security warning, I suspect that warning is there on master as well.

ljharb:
@LarsDenBakker `resolve` intends to implement every node resolution algorithm. The first step is the linked PR, which will lead to full "exports" support for CJS; the next step will be to release a new async-only API which provides ESM resolution.

shellscape:
@LarsDenBakker please pull from upstream/master to fix the audit error in the analysis step. _(note that you can create branches directly on this repo now)_

LarsDenBakker:
@shellscape done, thanks! My local repo is still configured to use my remote, will update that.

shellscape:
For those participating; we've had some discussion internally about whether to wait for browserify/resolve to get that PR through, or deliver some value to the userbase immediately and have sided with value for the userbase immediately. We'll revisit these changes when `resolve` updates, but will merge and release for now.

ljharb:
@shellscape please do follow that PR; when it lands and is released hopefully you'll be able to refactor to use resolve's implementation.

lukastaegert:
I think we should do definitely that. While I very welcome delivering this now for Rollup, your/browserify's package has always been an important centerpiece of this plugin and allowed us to have superior performance due to very intelligent directory-lookup caching possibilities.

If we can deliver the same functionality we implemented now using `resolve`, possibly even with better caching, then I think this would be a win for everyone. There was already one complaint with regard to the performance impact of this feature.

LarsDenBakker:
@shellscape @lukastaegert is it possible to cut a release of this?

pi0:
v11.1.0 is out with fix :)

PR #694, opened 2020-12-07
feat(node-resolve): add `ignoreSideEffectsForRoot` option

### LINKED ISSUE #692: [node-resolve]: `sideEffects` in project root package.json unexpectedly affects the bundled result of codes out of node_modules
- Rollup Plugin Name: plugin-node-resolve
- Rollup Plugin Version: 11.0.0
- Rollup Version: 2.34.2
- Operating System (or Browser): Mas OS, Chrome
- Node Version: v10.22.0
- Link to reproduction _(⚠️ read below)_:

https://repl.it/join/xcjdwawr-07akioni

### Expected Behavior
The `sideEffect` field in project root package.json shouldn't affect the bundled code inside the project.

### Actual Behavior
It has effects.

```bash
app-index.js → app-output.js...
created app-output.js in 101ms
var dist = { name: 'object' };

console.log(dist);

app-index.js → app-output-no-node-resolve.js...
created app-output-no-node-resolve.js in 57ms
var dist = { name: 'object' };

dist.sideEffect = true;

console.log(dist);
```

### Additional Information

I'm building a library without side effects. However the demo site (may have some side-effects, for example css or some codes) is both placed in same the folder. Should we make codes out of node_modules follow its `package.json`?

tjenkinson:
## Rollup Plugin Name: `resolve`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] yes
- [ ] no

Breaking Changes?

- [ ] yes
- [x] no

List any relevant issue numbers: #692

### Description

Use the `rootDir` option to determine if the file being resolved is in the root package, and if it is do not respect the package.json `side-effects` option, providing `ignoreSideEffectsForRoot` is `true`.

Note the default is remaining as `false` and we are not treating this as a bug, because `false` matches the default behaviour in webpack.

Refs #692


LarsDenBakker:
I'm not sure if this is a bug, why should side effects setting not apply to the project itself?

07akioni:
> I'm not sure if this is a bug, why should side effects setting not apply to the project itself?

I'm not sure about the behavior too. However I think it should be documented, or **made as an option**.

For example I'm making a non-side-effects library, and the doc site is placed in the project root too. If I use rollup to create a doc site bundle (which vite does), css (for the site ) may be shaken by rollup. When I apply `sideEffects: false` in package.json, I only want the entry point to be shaken.

shellscape:
It might be worth comparing to webpack@4, since v5 is still a bit of a trash can fire.

shellscape:
I think webpack has set good precedent here and following their lead is reasonable. I also think documenting this as suggested by @07akioni is a great idea. @LarsDenBakker and @07akioni could you take another look at this one please?

LarsDenBakker (dismissed):

LarsDenBakker:
Sorry I misunderstood, I thought webpack does what youre proposing in this PR. Lets keep it as is then.

shellscape:
😆  I'll wait from direction from you both on when/if to merge.

shellscape:
I'm good adding that option.

07akioni:
Hope this can be merged soon. 😄

LarsDenBakker (approved):

shellscape (approved):

Issue #679, opened 2020-12-02
Update plugins to be built with TypeScript 4


benmccann:
- Rollup Plugin Name: various
- Rollup Plugin Version: `master`

### Expected Behavior / Situation

I would like the plugins to be built with TypeScript 4 so that we are using the latest version

### Actual Behavior / Situation

`package.json` contains `"typescript": "^3.9.7",`

### Modification Proposal

Update `package.json` to `"typescript": "^4.0.0",`. Here's a commit showing the other required changes: https://github.com/rollup/plugins/pull/674/commits/8d07a42cdc03188ae14107c439f70a976af8d33c

What is the process for this? Do we need to update one plugin at a time?

shellscape:
I'm going to update the `CONTRIBUTING` doc to cover this question. Updating multiple plugins at once isn't something we'd like to encourage;  mostly because it requires a super clean commit history in the feature branch, but also because multiple plugin updates are harder for contributors to review, and we've got a lot of folks that only tend to one or two plugins here. To make matters more murky, it looks like the typescript dependency is at the root, so that makes things nice and uniform but also tricky.

For the sake of expediency for this effort, I'd say tackle one at a time, and add the updated typescript dependency to each plugin being worked on. (We can do a follow up commit later to remove this deps if needs be). I do also want to raise awareness in this issue that we have a lot of plugins that we haven't been able to get around to converting to TS, and that's likely more important than updating the TS version, but it's hard to say for certain because one could argue that it'd be easier to do on an updated version of TS.

benmccann:
There's about 20 different packages that currently use TypeScript. Once https://github.com/rollup/plugins/pull/674 is merged the only change necessary to the others is bumping the TypeScript version. None of the others require code changes and it will not create any backwards incompatibilities. I'm not sure if it's an effective use of time to create 20 different PRs all bumping the version for the same package.

Is this something that would be possible for a maintainer to bump them all in one go? This really seems like something that could be a 1-line change. I would understand if there were code changes to the individual packages, but it doesn't seem like a dependency bump with no code change should need individual reviews from 20 different package maintainers. I don't really see what the risk is.

benmccann:
The individual plugins are using TypeScript 4 now. Should we remove the TypeScript version from the base `package.json` or should we set the base version to version 4 and remove the version from the sub-packages?

shellscape:
I'd update the base. It'll be a while before we have to do another major version bump of that.

shellscape:
[...]

PR #701, opened 2020-12-08
Move dependencies to devDependencies and upgrade devDependencies

benmccann:
## Rollup Plugin Name: common development setup

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [X] other

Are tests included?

- [X] no

Breaking Changes?

- [X] no

### Description

The dependencies listed under `dependencies` are only used in scripts and not in any of the plugins themselves, so I believe they should be listed under `devDependencies` since they're not included in the source, but only used during the development process. I also upgraded most of the dependencies as well. They're just used during tests, etc. so won't affect the plugins themselves


shellscape:
But why? The range changes are superficial, and because it's a monorepo deps vs devDeps doesn't really mean anything. If anything, maybe we talk about what it means in an issue.

I appreciate all of the work you're taking on here, but let's chat about some things before you dive right on into it. We're a team here, and the onslaught (in a goodish way) today is bananas. We're not likely to be able to approve and merge them all in a day (a week?) and because we're a team you've unloaded a lot of burden at once. A steady pace would be more welcome.

benmccann:
It's inconsistent to have some as `dependencies` and some as `devDependencies`, so this brings greater consistency.

There are also several packages with major version bumps.

The majority of the PRs are simply bumping the TypeScript version. I'm not quite sure what is required to review those. I did mention in https://github.com/rollup/plugins/issues/679 that it would create quite a few PRs, but that was the suggestion that was made to me.

shellscape:
I'd recommend going back and rereading what I wrote again, and absorb the message in that reply.

benmccann:
Ok, I thought your earlier suggestion regarding repo-level changes was only referring to changes which would affect the plugin distributions. These changes are only for development tools and don't actually affect the code that end users receive from the individual plugins, so I didn't think the earlier statement applied here

I agree we should try to fix the CI issue if possible. It seems like that would address a lot of the manual review work, etc.

PR #715, opened 2020-12-08
chore(pluginutils): upgrade TypeScript

benmccann:

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [X] other

Are tests included?

- [X] no

Breaking Changes?

- [X] no

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -44,5 +44,13 @@ module.exports = {
         plugins: ['prettier-plugin-package']
       }
     ]
-  }
+  },
+  overrides: [
+    {
+      "files": ["**/*.ts"],
+      "rules": {
+        "no-undef": "off"
+      }
+    }
+  ]
 };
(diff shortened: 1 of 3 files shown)


NotWoods (approved):

benmccann:
`lint` is failing on `master`, so the failure is unrelated to this PR. See https://github.com/rollup/plugins/pull/718 for more details

benmccann:
@NotWoods I'm not quite sure the best way to fix the `lint` error here. Using `import type` as proposed in https://github.com/rollup/plugins/pull/718 still seems the most sensible to me. I filed a bug (https://github.com/alexgorbatchev/eslint-import-resolver-typescript/issues/64) and they suggested the rule should not be enabled on TypeScript files - though it seems complicated in our case because that would also turn it off for the JS files.

So three paths forward:
- Change dynamic imports to `import type`
- Turn off the rule for `.ts` and `.js` files
- `estlint-disable-line no-undef` comments on each line

Do you have a preference on the solution here?

NotWoods:
Can we use `overrides` in the ESLint config to only turn off the rule in TypeScript files?

benmccann:
Thanks for the suggestion. That worked and everything is passing now

PR #708, opened 2020-12-08
chore(alias): upgrade TypeScript

benmccann:
## Rollup Plugin Name: `alias`

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [X] other

Are tests included?

- [ ] yes (_bugfixes and features will not be merged without tests_)
- [X] no

Breaking Changes?

- [ ] yes (_breaking changes will not be merged unless absolutely necessary_)
- [X] no

If yes, then include "BREAKING CHANGES:" in the first commit message body, followed by a description of what is breaking. 

List any relevant issue numbers:

PR #717, opened 2020-12-08
chore(alias): fix lint errors

benmccann:
`lint` is failing on `master` for this project. This fixes it

### DIFF
--- a/packages/alias/src/index.ts
+++ b/packages/alias/src/index.ts
@@ -25,9 +25,7 @@ function matches(pattern: string | RegExp, importee: string) {
   return importeeStartsWithKey && importeeHasSlashAfterKey;
 }
 
-function normalizeId(id: string): string;
-function normalizeId(id: string | undefined): string | undefined;
-function normalizeId(id: string | undefined) {
+function normalizeId<T extends string | undefined>(id: T) {
   if (typeof id === 'string' && (IS_WINDOWS || VOLUME.test(id))) {
     return slash(id.replace(VOLUME, ''));
   }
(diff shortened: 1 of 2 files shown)


packages/alias/src/index.ts
```diff
@@ -25,9 +25,7 @@ function matches(pattern: string | RegExp, importee: string) {
   return importeeStartsWithKey && importeeHasSlashAfterKey;
 }
 
-function normalizeId(id: string): string;
-function normalizeId(id: string | undefined): string | undefined;
-function normalizeId(id: string | undefined) {
+function normalizeId<T>(id: T) {
```
NotWoods, line 28: This should be restricted to just strings
```suggestion
function normalizeId<T extends string | undefined>(id: T) {
```

NotWoods (changes requested):
What lint errors are being triggered here?

benmccann:
>packages/alias/src/index.ts
  29:10  error  'normalizeId' is already defined  no-redeclare
  30:10  error  'normalizeId' is already defined  no-redeclare

packages/alias/src/index.ts
benmccann: makes sense. done

NotWoods:
This sounds like an ESLint issue as we're using a standard TypeScript feature here called function overloading. I'd prefer to keep the overloads as its slightly faster and easier to read.

benmccann:
Ok, I've reverted the fix for the `no-redeclare` failure and just included the other fix for now. I can report an issue to the typescript eslint project after this is merged for the remaining issue

NotWoods (approved):

shellscape:
Closing in favor of #708. The fix for the linting as related to a typescript upgrade (whether or not that's the cause) should be done there since it's related.

PR #718, opened 2020-12-09
chore(pluginutils): fix lint errors

benmccann:

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [X] other

Are tests included?

- [X] no

Breaking Changes?

- [X] no

### Description

This fixes all but one lint error. I think the last one is a bug in `typescript-eslint` and not a real error. I'll file a bug there to try to get it addressed after this is merged

### DIFF
--- a/packages/pluginutils/src/extractAssignedNames.ts
+++ b/packages/pluginutils/src/extractAssignedNames.ts
@@ -1,3 +1,11 @@
+import type {
+  ArrayPattern,
+  AssignmentPattern,
+  Identifier,
+  ObjectPattern,
+  RestElement
+} from 'estree';
+
 import { ExtractAssignedNames } from '../types';
 
 interface Extractors {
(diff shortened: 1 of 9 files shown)


packages/pluginutils/src/extractAssignedNames.ts
```diff
@@ -1,27 +1,35 @@
...
+
 import { ExtractAssignedNames } from '../types';
 
 interface Extractors {
   [key: string]: (names: string[], param: any) => void;
 }
 
 const extractors: Extractors = {
-  ArrayPattern(names: string[], param: import('estree').ArrayPattern) {
+  ArrayPattern(names: string[], param: ArrayPattern) {
```
NotWoods, line 16: The `import()` style is used here to avoid confusion between extractors and estree. Can we leave this spot alone and file an issue with the linter?

packages/pluginutils/src/normalizePath.ts
```diff
@@ -2,7 +2,7 @@ import { win32, posix } from 'path';
 
 import { NormalizePath } from '../types';
 
-const normalizePath: NormalizePath = function(filename: string) {
+const normalizePath: NormalizePath = function (filename: string) {
```
NotWoods, line 5: Was there a change in prettier here? IIRC the previous style is the one used by prettier.

benmccann: yeah, I didn't do this. it was auto-formatted this way

packages/pluginutils/src/extractAssignedNames.ts
benmccann: ok, I'll close this PR then

PR #725, opened 2020-12-09
Added CSP hash injection functionality to html plugin

ohmree:
## Rollup Plugin Name: `html`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

When building apps with languages that compile to WASM (like Rust) I've faced the issue of webkit and blink-based browsers refusing to load the WASM unless I specifically allow the loader scripts using CSP (the same also applies to styles).  
It's even more of an issue with my specific use-case - in addition to regular serving I also package my web app into a single HTML file and embed it inside a Rust binary, which means everything is inline - so I had to find a solution for this problem.
Since allowing all unsafe evals and/or inline scripts/styles is bad (more so in a regular web app but I'd prefer not to allow them even in my native wrapper), and since webpack has a [similar plugin](https://github.com/ohmree/csp-html-webpack-plugin), I decided to try and hack together something that works for me (tm).  
I know this won't get accepted without tests and I'd be open to learning how to write them, but I've never really written unit tests before and am not sure where to begin.
Alternatively, or if this is beyond the scope of the HTML plugin, is there a way to implement something like this as a separate plugin without forking or essentially duplicating this one? After becoming a little bit familiar with this plugin's structure I'd guess it isn't possible, but I'm not 100% sure.

### DIFF
--- a/packages/html/package.json
+++ b/packages/html/package.json
@@ -44,9 +44,6 @@
   "peerDependencies": {
     "rollup": "^1.20.0||^2.0.0"
   },
-  "dependencies": {
-    "hasha": "^5.2.2"
-  },
   "devDependencies": {
     "@rollup/plugin-typescript": "^6.1.0",
     "rollup": "^2.23.0",
--- a/packages/html/package.json
+++ b/packages/html/package.json
@@ -44,6 +44,9 @@
   "peerDependencies": {
     "rollup": "^1.20.0||^2.0.0"
   },
+  "dependencies": {
+    "hasha": "^5.2.2"
+  },
   "devDependencies": {
     "@rollup/plugin-typescript": "^6.1.0",
     "rollup": "^2.23.0",
--- a/packages/html/src/index.ts
+++ b/packages/html/src/index.ts
@@ -1,6 +1,15 @@
 import { extname } from 'path';
 
-import { Plugin, NormalizedOutputOptions, OutputBundle, EmittedAsset } from 'rollup';
+import hasha from 'hasha';
+
+import {
+  Plugin,
+  NormalizedOutputOptions,
+  OutputBundle,
+  EmittedAsset,
+  OutputChunk,
+  OutputAsset
+} from 'rollup';
 
 import { RollupHtmlOptions, RollupHtmlTemplateOptions } from '../types';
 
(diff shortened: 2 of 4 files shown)


ohmree:
I just realized my code pulls in the `hasha`dependency unconditionally. 
Perhaps it'd be better to make it an optional dependency, dynamically import it and only enable CSP-related functionality if the import succeeded.  
Either that, or I can port the code to [the node crypto api](https://nodejs.org/api/crypto.html#crypto_class_hash) but it'd involve a bit of learning.

ohmree:
I've removed the `hasha` dependency, the code now uses `crypto.Hash`.  
That does mean no async API (`stream/promises`) if we want to support node < 15 but I'm not sure how meaningful that really is.

shellscape:
@ohmree thanks for putting this together - it looks like a lot of effort went into it. I always recommend to folks wanting to submit a large feature like this to open a new feature issue before putting the time in on it. That said, I think this feature is out of scope for the plugin. We intended the plugin to be very svelte and to empower users to do more with it via the options provided  - that's why the `template` option allows for overriding or reusing everything.

It sounds like you had an immediate need and may be using a fork in the meantime anyhow, so I don't think it was wasted effort. I'd also suggest a complimentary utility package (which we could mention in the README) to apply these changes using the `template` option of the plugin as-is. Of course, if you'd rather use your fork we'd be happy to point folks to that as well in the README.

I'll keep this open to allow other maintainers the opportunity to comment, or disagree with me, but it's my position at the moment that it's not in scope and would benefit too few users for the added complexity and maintenance. Thanks once more for the effort you put in on this.

ohmree:
Hi @shellscape , sorry for taking a long time to answer.

This was really a hack I put together to suit my particular needs (and you were right about me using a fork in my project).

Since making this I've stopped working on the project I needed this for and haven't done any wasm or node work, but IIRC the reason I needed a fork was because the plugin didn't expose enough data for me to put in `template`, I can't remember what exactly I needed but I can try to look into it.

If I do pinpoint what I needed exposed and write up a PR to implement it, would you be open to accept it?

PR #726, opened 2020-12-10
fix(typescript) replace getOutputFileNames

### LINKED ISSUE #287: Typescript plugin 4.0.0 Error: Unexpected token (Note that you need plugins to import files that are not JavaScript)

### LINKED ISSUE #608: [typescript] Broken vue3 + rollup + typescript
- Rollup Plugin Name: typescript
- Rollup Plugin Version: 6.0.0 (current latest)
- Rollup Version: 2.29.0
- Operating System (or Browser): linux if neccesary, reproduces at bundle build time
- Node Version: 14.13.1
- Link to reproduction _(⚠️ read below)_: https://github.com/iassasin/vue3-rollup-ts-bug

### Expected Behavior

Working bundle.

### Actual Behavior

`[!] (plugin typescript) Error: Could not load .../MainView.vue?vue&type=script&lang.ts (imported by MainView.vue): Debug Failure. False expression: Expected fileName to be present in command line`

### Additional Information

Also reported to vue-rollup-plugin (https://github.com/vuejs/rollup-plugin-vue/issues/400), but after some debug I suggess that bug is in plugin-typescript, because with rollup-plugin-typescript2 bundle built successfuly. Difference to plugin-typescript2 is official plugin-typescript uses function `ts.getOutputFileNames`, that produces failed assert.

Samantha-uk:
## Rollup Plugin Name: `{@rollup/plugin-typescript}`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no _(I don't believe so)_

An issue I experienced _(details below)_ and rather than logging it, I set out to fix it.
Potentially resolves #287 
Potentially resolves #608 
### Description

The problem I experienced was that when I used the Rollup API, if I set `tsconfig:false` and passed in all of my compilerOptions to the typescript plugin, the system would fail ... with a very unhelpful message.

Turns out is was a failing assertion deep inside  typescript as part of the `getOutputFileNames` call chain. _(Little did I realise at the time that it had been causing grief for others in a different fashion)_.

So I set out to see if there was another approach...

It turns out that TS knows _(obviously)_ and  _already tells the  typescript plugin where it has generated files and what the corresponding source files are, this comes via the BuildProgram `writeFile` callback.  By utilising the extra information this PR obviates the need for `getOutputFileNames` an external call we no longer need to make. 🥳 

I've added a new test _(that failed pre these changes)_ and I've _slightly_ tweaked a couple of others as the order of files reported to the test harness has changed and the FakeTypescript test needed to also pass in the newly used source file information.

I also changed the way that the `generateBundle` figures out the baseDir path to use ... basing it on the approach MS document on their TS site _(URL in the source)_ ... however, whilst it works and does not break tests ... I feel that there is perhaps a better way to calculate this ... perhaps just copying the hierarchy that the TS program generates? 🤔 

There is a possibility that this may also address issue #287 however I was unsuccessful in getting the test that @benmccann added ... but I think it is close ... 🤞

### DIFF
--- a/packages/typescript/package.json
+++ b/packages/typescript/package.json
@@ -51,6 +51,7 @@
   },
   "dependencies": {
     "@rollup/pluginutils": "^3.1.0",
+    "common-path-prefix": "^3.0.0",
     "resolve": "^1.17.0"
   },
   "devDependencies": {
--- a/packages/typescript/src/index.ts
+++ b/packages/typescript/src/index.ts
@@ -1,7 +1,8 @@
 import * as path from 'path';
 
-import { Plugin, SourceDescription } from 'rollup';
-import type { Watch } from 'typescript';
+import { NormalizedOutputOptions, Plugin, SourceDescription } from 'rollup';
+import type { SourceFile, Watch } from 'typescript';
+import CommonPathPrefix from 'common-path-prefix';
 
 import { RollupTypescriptOptions } from '../types';
 
@@ -10,7 +11,7 @@ import createModuleResolver from './moduleResolution';
 import getPluginOptions from './options/plugin';
 import { emitParsedOptionsErrors, parseTypescriptConfig } from './options/tsconfig';
 import { validatePaths, validateSourceMap } from './options/validate';
-import findTypescriptOutput, { getEmittedFile } from './outputFile';
+import getTSGeneratedOutput, { getEmittedFile, TSGeneratedFile } from './outputFile';
 import createWatchProgram, { WatchProgramHelper } from './watchProgram';
 import TSCache from './tscache';
 
@@ -25,7 +26,9 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
     typescript: ts
   } = getPluginOptions(options);
   const tsCache = new TSCache(cacheDir);
-  const emittedFiles = new Map<string, string>();
+
+  const tsGeneratedFiles = new Map<string, TSGeneratedFile>();
+  let commonPathPrefix = ``;
   const watchProgramHelper = new WatchProgramHelper();
 
   const parsedOptions = parseTypescriptConfig(ts, tsconfig, compilerOptions);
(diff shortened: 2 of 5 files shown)


Samantha-uk:
Oh dear ... how embarrassing that the tests are failing.
I'm running Debian buster / node 14.15.1 and they all pass.
image
Any ideas of what I may need to change?

benmccann:
Not sure about the failing tests, but I wonder if the issue you were facing would have been fixed by https://github.com/microsoft/TypeScript/pull/41811. This seems like potentially extra code in the plugin that might not really be necessary after upgrading TypeScript

Samantha-uk:
I saw that and It may be fixed in that but not soon.  The PR changes are working nicely in my build environment so I’m thankfully unblocked.

PR #733, opened 2020-12-15
Turn off lint rule that's causing attempted reformatting

benmccann:
## Rollup Plugin Name: `all`

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [X] other

Are tests included?

- [X] no

Breaking Changes?

- [X] no

### Description

This project runs `eslint --fix` as part of `pnpm -r lint`. There are a lot of files that do not abide by this rule, so it tries to replace every instance of `Object.assign` with a spread operator. There are a pretty good number of places where the code does not conform to these expectations. This would let the code remain how it is instead of it being automatically updated

There's one other major issue to address and I've raised a separate issue for it: https://github.com/rollup/plugins/issues/734

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -35,6 +35,7 @@ module.exports = {
     'import/no-named-export': 'off',
     'no-redeclare': 'off',
     'no-unused-vars': 'off',
+    'prefer-object-spread': 'off',
     'spaced-comment': 'off',
     'prettier/prettier': [
       'error',


shellscape:
I'd like to know the _why_ of this before we turn off another rule.

benmccann:
I'm guessing that this rule was never enforced (either because it's a new rule or the default as to whether to enforce it changed), but when `eslint` was upgraded it started running this check. I think the bigger question is do we want to turn off this rule or do we want to replace all instances of `Object.assign` with a spread operator?

We should at some point also try to enforce it such that the code can't get out-of-sync again. I tried to see if we could have `lint` run on all projects on the CI to prevent this in the future, but there were too many `lint` failures to get it running. This addresses a lot of the `lint` failures. If we can get them all fixed then we can experiment with enforcing the `lint` checks

shellscape:
Thanks for the explanation. That's good enough for me. I don't think we should replace all instances, and I think there are situations that call for Object.assign over spread as well.

shellscape (approved):

PR #750, opened 2020-12-30
fix(node-resolve): only log last resolve error

LarsDenBakker:

This PR contains:

- [X] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [X] no

### Description

When attempting to resolve a list of specifiers we are logging warnings for failed attempts. This change only logs the last failed attempt.

This is visible for example when resolving `.js` imports in a `.ts` file.

### DIFF
--- a/packages/node-resolve/src/resolveImportSpecifiers.js
+++ b/packages/node-resolve/src/resolveImportSpecifiers.js
@@ -132,11 +132,10 @@ async function resolveId({
         );
         location = fileURLToPath(resolvedPackageExport);
       } catch (error) {
-        if (!(error instanceof ResolveError)) {
-          throw error;
+        if (error instanceof ResolveError) {
+          return error;
         }
-        warn(error);
-        return null;
+        throw error;
       }
     }
   }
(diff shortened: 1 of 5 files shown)


shellscape:
For posterity, could you post the before and after of what the console output looks like?

LarsDenBakker:
Let's say I have a file called `my-file.ts`, with these contents:

```js
import 'some-package/foo.js';
```

This package has an export map:

```js
{
  "exports": {
    "./foo.js": "./foo.js"
  }
}
```

and node resolve is configured with extensions `['.ts', '.js']`

Because of the special handling for .js imports in TS files, node-resolve first tries to resolve `some-package/foo.ts`, triggering a warning from the new export map functionality:

```
Could not resolve import "some-package/foo.ts" in my-file.ts using exports defined in /node_moduls/some-package/package.json. 
```

It then tries to resolve using `some-package/foo.js` which succeeds. So the warning is quite confusing to the user.

With this change the first warning is never logged. 

In the event that the import is actually incorrect, before this change the warning would be logged twice, once for each extension. With this change it's logged once with the import the user made.

shellscape (approved):

PR #767, opened 2021-01-14
feat(babel): add filter option

chengcyber:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

~Since `createFilter` has a third options, which can pass `resolutionBase`. It might be better the option can be controlled by `rollup-plugin-babel`~

According to the following discussion, adding filter option to @rollup/plugin-babel.

### DIFF
--- a/packages/babel/README.md
+++ b/packages/babel/README.md
@@ -92,6 +92,21 @@ Type: `String | RegExp | Array[...String|RegExp]`<br>
 
 A [minimatch pattern](https://github.com/isaacs/minimatch), or array of patterns, which specifies the files in the build the plugin should operate on. When relying on Babel configuration files you cannot include files already excluded there.
 
+### `filter`
+
+Type: (id: unknown) => boolean<br>
+
+Custom [filter function](https://github.com/rollup/plugins/tree/master/packages/pluginutils#createfilter) can be used to determine whether or not certain modules should be operated upon.
+
+Usage:
+
+```js
+import { createFilter } from '@rollup/pluginutils';
+const include = 'include/**.js';
+const exclude = 'exclude/**.js';
+const filter = createFilter(include, exclude, {});
+```
+
 ### `extensions`
 
 Type: `Array[...String]`<br>
--- a/packages/babel/README.md
+++ b/packages/babel/README.md
@@ -94,7 +94,7 @@ A [minimatch pattern](https://github.com/isaacs/minimatch), or array of patterns
 
 ### `filter`
 
-Type: (id: unknown) => boolean<br>
+Type: (id: string) => boolean<br>
 
 Custom [filter function](https://github.com/rollup/plugins/tree/master/packages/pluginutils#createfilter) can be used to determine whether or not certain modules should be operated upon.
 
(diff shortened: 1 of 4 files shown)


shellscape:
Thanks for the PR. Unfortunately we cannot consider this PR until tests are added for the change. Please add tests or we'll have to close.

chengcyber:
Hi @shellscape ,

As per this issue, https://github.com/rollup/rollup/issues/3013#issuecomment-517133601

Do you think it's a better solution to add `cwd` option to rollup self, and change `createFilter` internally respect the rollup's `cwd` options as default resolutionBase?

I found it's kind of weird to add test for `resolutionBase` option to `@rollup/plugin-babel` 😅

Andarist:
It seems that maybe instead of adding more `createFilter`-related options it would be better to just accept a single `filter` option and have people pass the `createFilter` result to the babel plugin instead? @shellscape WDYT?

shellscape:
@Andarist if it makes sense for the plugin, then I'm in agreement. we'll definitely want the docs to be clear about why this plugin behaves differently than the other ones, and why it accepts different filtering options.

chengcyber:
refactor the code and add filter option to @rollup/plugin-babel instead.

packages/babel/README.md
```diff
@@ -75,6 +75,21 @@ Type: `String | RegExp | Array[...String|RegExp]`<br>
 
 A [minimatch pattern](https://github.com/isaacs/minimatch), or array of patterns, which specifies the files in the build the plugin should operate on. When relying on Babel configuration files you cannot include files already excluded there.
 
+### `filter`
+
+Type: (id: unknown) => boolean<br>
```
Andarist, line 80: doesn't this always receive a `string`?

chengcyber: Originally, the type should be ReturnType<typeof createFilter> right?. As per https://github.com/rollup/plugins/blob/master/packages/pluginutils/types/index.d.ts#L52

packages/babel/src/index.js
```diff
@@ -122,21 +122,29 @@ function createBabelInputPluginFactory(customCallback = returnObject) {
...
           skipPreflightCheck,
           ...babelOptions
         } = unpackInputPluginOptions(pluginOptionsWithOverrides, this.meta.rollupVersion));
 
         const extensionRegExp = new RegExp(
           `(${extensions.map(escapeRegExpCharacters).join('|')})$`
         );
-        const includeExcludeFilter = createFilter(include, exclude);
-        filter = (id) => extensionRegExp.test(stripQuery(id).bareId) && includeExcludeFilter(id);
+        const userDefinedFilter = (id) => {
```
Andarist, line 140: Sorry for not mentioning that before (I would sweat that I've done that but apparently I didn't) - I would not try to compose those things here. If a custom `filter` gets passed in I would just throw if `exclude` or `include` would be passed in as well. 

Something along those lines:
```js
if (customFilter && (include || exclude)) {
  throw new Error('...')
}
const configFilter = customFiler || createFilter(include, exclude);
filter = (id) => extensionRegExp.test(stripQuery(id).bareId) && configFilter(id);
```

chengcyber: Change the code logic, switch to your design. and add test as well.

Andarist (approved):

PR #771, opened 2021-01-18
fix(typescript): only emit tsbuildinfo file when there is something to emit

### LINKED ISSUE #681: Rollup fails with `The "path" argument must be of type string. Received undefined`
- Rollup Plugin Version: 8.0.0
- Rollup Version: 2.34.0
- Operating System (or Browser): macOS
- Node Version: v12.18.3
- Link to reproduction _(⚠️ read below)_: https://repl.it/join/yyhrqzyx-idubrov

### Expected Behavior

Rollup rolls up everything without any errors.

### Actual Behavior

Rollup fails with the following error:
```
[!] TypeError: The "path" argument must be of type string. Received undefined
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
```

Run `npm build` in the REPL.it above to see the issue.

### Additional Information

Line where rollup fails (plus/minus the version): https://github.com/rollup/rollup/blob/1fa937468fd6033f8d8ec10e800d671fc1d01f7e/src/rollup/rollup.ts#L251

The cause of the failure is chunk created for the ".tsbuildinfo" file, which is emitted at this line: https://github.com/rollup/plugins/blob/a41600fac25475608983466119d2e93a1f7d2f3c/packages/typescript/src/index.ts#L140-L144

When TypeScript is running in incremental mode & there is nothing to rebuild, it does not produce any output for `.tsbuildinfo` (`emittedFiles.get(tsBuildInfoPath)` in the link above is `undefined`). It seems like emitting file with `undefined` `source` is what causes `rollup` to fail.

Necessary conditions:

1. TypeScript is running with incremental mode (in our case, via `composite` set to `true` in `tsconfig.json`). See the reproduction scenario.
2. Only second build fails. The first one seems to succeed because TS does all the work & everything is fine. On the second build it seems like TS decides not to emit anything (since there is nothing to do).

cherryblossom000:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no
~~I'm still trying to figure out why [this test](https://github.com/rollup/plugins/blob/e4a4cec3ef0419f0a9a6ef6a02ef4034c7919451/packages/typescript/test/test.js#L853-L878) didn't fail before this fix.~~
It was because the bundle output wasn't written to disk, so the second run would be identical to the first as TypeScript wouldn't find the existing `.tsbuildinfo`.

Breaking Changes?

- [x] no

List any relevant issue numbers: #681

### Description

Fixes #681

The plugin now checks if TypeScript has emitted a tsbuildinfo file. If and only if it has, the plugin emits the tsbuildinfo file.

This prevents the TypeError from occurring when the plugin used to emit the file even when there was no source for it (`emittedFiles.get(tsBuildInfoPath)` was `undefined`).

### DIFF
--- a/packages/typescript/src/index.ts
+++ b/packages/typescript/src/index.ts
@@ -142,11 +142,15 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
 
       const tsBuildInfoPath = ts.getTsBuildInfoEmitOutputFilePath(parsedOptions.options);
       if (tsBuildInfoPath) {
-        this.emitFile({
-          type: 'asset',
-          fileName: normalizePath(path.relative(outputOptions.dir!, tsBuildInfoPath)),
-          source: emittedFiles.get(tsBuildInfoPath)
-        });
+        const tsBuildInfoSource = emittedFiles.get(tsBuildInfoPath);
+        // https://github.com/rollup/plugins/issues/681
+        if (tsBuildInfoSource) {
+          this.emitFile({
+            type: 'asset',
+            fileName: normalizePath(path.relative(outputOptions.dir!, tsBuildInfoPath)),
+            source: tsBuildInfoSource
+          });
+        }
       }
     }
   };
--- a/packages/typescript/test/fixtures/incremental-output-cache/main.ts
+++ b/packages/typescript/test/fixtures/incremental-output-cache/main.ts
@@ -1,0 +1,6 @@
+type AnswerToQuestion = string | undefined;
+
+const answer: AnswerToQuestion = '42';
+
+// eslint-disable-next-line no-console
+console.log(`the answer is ${answer}`);
--- a/packages/typescript/test/fixtures/incremental-output-cache/tsconfig.json
+++ b/packages/typescript/test/fixtures/incremental-output-cache/tsconfig.json
@@ -1,0 +1,7 @@
+{
+  "compilerOptions": {
+    "incremental": true,
+    "outDir": "./dist",
+    "tsBuildInfoFile": "./dist/.tsbuildinfo"
+  }
+}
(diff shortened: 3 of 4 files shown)


packages/typescript/src/index.ts
```diff
@@ -142,11 +142,15 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
...
       const tsBuildInfoPath = ts.getTsBuildInfoEmitOutputFilePath(parsedOptions.options);
       if (tsBuildInfoPath) {
-        this.emitFile({
-          type: 'asset',
-          fileName: normalizePath(path.relative(outputOptions.dir!, tsBuildInfoPath)),
-          source: emittedFiles.get(tsBuildInfoPath)
-        });
+        const tsBuildInfoSource = emittedFiles.get(tsBuildInfoPath);
+        // https://github.com/rollup/plugins/issues/681
+        if (tsBuildInfoSource) {
```
shellscape, line 147: > The plugin now checks if TypeScript has emitted a tsbuildinfo file. If it hasn't, the plugin doesn't emit the file, which prevents the TypeError from occurring.

This `if` block here seems to be opposite of what you're describing in the PR description. Which one is the correct behavior?

cherryblossom000: If and only if TypeScript has emitted a tsbuildinfo file (`if (tsBuildInfoSource)`), the plugin emits the tsbuildinfo file.	This is the same behaviour described in the PR description, but I'll update it to be a bit clearer. Sorry for the confusion.

shellscape: Thanks for the further explanation, and no worries.

shellscape (approved):

shellscape:
thanks!

PR #780, opened 2021-01-29
Check types in typescript packages in CI

### LINKED ISSUE #779: Check types in CI for all plugins written in ts
- Rollup Plugin Name: all
- Rollup Plugin Version: any

### Feature Use Case
Have early feedback in every PR that typescript types are not compromised.

### Feature Proposal
Add task `ci:check-types` to every plugin
Run that task on circle ci config

fernandopasik:
## Rollup Plugin Name: `alias, auto-install, data-url, eslint, html, pluginutils, typescript, virtual, wasm`

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no (adds tasks to projects and ci)

Breaking Changes?

- [x] no

fixes #779 

### Description

Have early feedback in every PR that typescript types are not compromised.
Add task test:ts to every plugin written in typescript
Run  test:ts on circle ci config

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,6 +70,9 @@ jobs:
       - run:
           name: Run linting.
           command: pnpm run ci:lint
+      - run:
+          name: Run check TS types.
+          command: pnpm run test:ts
       - save_cache:
           key: dependency-cache-{{ checksum "pnpm-lock.yaml" }}
           paths:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,7 +71,7 @@ jobs:
           name: Run linting.
           command: pnpm run ci:lint
       - run:
-          name: Run check TS types.
+          name: Check TypeScript types.
           command: pnpm run test:ts
       - save_cache:
           key: dependency-cache-{{ checksum "pnpm-lock.yaml" }}
--- a/package.json
+++ b/package.json
@@ -15,7 +15,8 @@
     "pub": "node scripts/pub.js",
     "publish": "node scripts/publish.js",
     "security": "pnpm audit --audit-level=moderate",
-    "test": "node scripts/run-changed.js test"
+    "test": "node scripts/run-changed.js test",
+    "test:ts": "node scripts/run-changed.js test:ts"
   },
   "dependencies": {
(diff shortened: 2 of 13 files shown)


.circleci/config.yml
```diff
@@ -70,6 +70,9 @@ jobs:
       - run:
           name: Run linting.
           command: pnpm run ci:lint
+      - run:
+          name: Run check TS types.
```
NotWoods, line 74:
```suggestion
          name: Check TypeScript types.
```

NotWoods (changes requested):
I think "check" or "typecheck" would be a better script name

fernandopasik:
Hi @NotWoods, after a bit more thought I asked some question about this approach here https://github.com/rollup/plugins/issues/779#issuecomment-769987045

> I think I'm missing context on why is it necessary to commit the types/index.d.ts when the source is already typescript. I was expecting that file to be generated along side the js files before publishing to npm only. @shellscape Do you have more information on this?

I don't think the `test:ts` task is checking well the types. I checked this reverting locally the changes in https://github.com/rollup/plugins/pull/647 and it doesn't catch the error.
It did catch it if I directly run `tsc --noEmit`, because that runs on all ts files of the plugin.

shellscape:
So I know that historically we've had the the `index.d.ts` files hand-written before plugins were moved to TypeScript. That made an additional test of the types necessary. Some of those have stuck around. We even have some TypeScript-sourced plugins (maybe they're all like this still?) that have a separate, hand-written `index.d.ts`, which again is important to check against the code.

In the case of plugins which are already switched over to TypeScript sources, we should probably be generating the definitions at build time, and then removing the additional type check tests (as you point out, it'd be unnecessary). If we do that, for any plugin which uses generated definitions rather than the hand-written definition, we'll have to do a separate PR for each plugin.

fernandopasik:
Ok, sounds good @shellscape. But for the scope of this PR I managed to deliver the type checking by checking all the ts files in each package, not just the `index.d.ts` like before.

@NotWoods I've merged your proposed change.

NotWoods (approved):

Issue #772, opened 2021-01-22
[node-resolve11] resolve builtin module as external module breaks rollup-plugin-node-polyfills


chengcyber:
- Rollup Plugin Name: `node-resolve`
- Rollup Plugin Version: `^11`
- Rollup Version: `2.37.1`
- Operating System (or Browser): MacOS
- Node Version: 12
- Link to reproduction _(⚠️ read below)_: https://repl.it/@chengcyber/rollup-plugin-repro-1

### Expected Behavior

builtin module can be polyfill by `rollup-plugin-node-polyfills`

### Actual Behavior

builtin module like `buffer` can not be polyfilled by other plugin.

### Additional Information

As per https://github.com/rollup/plugins/pull/627/files#diff-dbc8c6c4e7383a37fbae0c12d8e6d8901c93631bb7774f4bd1a2a88bbe1980a0L262 `node-resolve` force treats `buffer` as external,which returns false instead of null when `resolveId('buffer')`. This prevents `rollup-plugin-node-polyfills`'s resolveId logic, since `resolve`, `commonjs` should be placed before `rollup-plugin-node-polyfills`

`@rollup/plugin-node-resolve` version `10` works.

tjenkinson:
Hi @chengcyber,

You need to tell the plugin that you don't want it to prefer builtins. That way `node-resolve` won't mark anything as `external`, and it also means if `rollup-plugin-node-polyfills` is unable to polyfill something you should end up with a warning from rollup stating there are unresolved dependencies.

e.g

```js
nodeResolve({ preferBuiltins: false })
```

Does this work for you? I'll open a PR to update the docs.

chengcyber:
> Hi @chengcyber,
> 
> You need to tell the plugin that you don't want it to prefer builtins. That way `node-resolve` won't mark anything as `external`, and it also means if `rollup-plugin-node-polyfills` is unable to polyfill something you should end up with a warning from rollup stating there are unresolved dependencies.
> 
> e.g
> 
> ```js
> nodeResolve({ preferBuiltins: false })
> ```
> 
> Does this work for you? I'll open a PR to update the docs.

I try to set `{ preferBuiltins: false }` in the repl.it, and it works. Thanks for this information!

PR #782, opened 2021-01-30
docs(node-resolve): clearer "Resolving Built-Ins" doc

tjenkinson:

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

List any relevant issue numbers: #772

### Description
With a recent update the plugin now marks builtins as external automatically when `resolveBuiltins` is not explicitly `false`.

Therefore I think the documentation to use `builtins` to achieve the same behaviour is not needed, given doing it is essentially a no-op.

Also when using `rollup-plugin-node-polyfills` you should set `preferBuiltins ` to `false`, and this was not super clear the way it was worded previously.

### DIFF
--- a/packages/node-resolve/README.md
+++ b/packages/node-resolve/README.md
@@ -187,14 +187,17 @@ By default this plugin will prefer built-ins over local modules, marking them as
 
 See [`preferBuiltins`](#preferbuiltins).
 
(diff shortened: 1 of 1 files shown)


shellscape (approved):
LGTM

Issue #774, opened 2021-01-24
[babel] Possibility to pass combinedSourcemap into plugins


KingSora:
- Rollup Plugin Name: `@rollup/plugin-babel`
- Rollup Plugin Version: `5.2.2`

### Feature Use Case

My rollup pipeline looks like this:
1. typescript
2. other plugins like `node-resolve` and `commonjs`
3. babel

I'm using the babel plugin [babel-plugin-istanbul](https://github.com/istanbuljs/babel-plugin-istanbul) to make it possible to generate coverage for my browser tests.

The resulting coverage reports aren't 100% accurate, because the `babel-plugin-istanbul` plugin relies on inline source maps.
My `typescript` build step is doing transformation and because rollup is doing its own sourcemaps management I need to pass the `combinedSourcemap` from rollup into the  `babel-plugin-istanbul` plugin so it knows how to handle the input code.

### Feature Proposal

Make it possible to pass the combinedSourcemap to babel plugins, so that plugins like `babel-plugin-istanbul` can use this information.
I guess it would be enough if the `createBabelInputPluginFactory` would get the rollup context so that you could access rollup internal stuff like the `getCombinedSourcemap()` function.

Or maybe there is a different possibility to handle my case..?

shellscape:
Thanks for the suggestion. This sounds like a reasonable feature to add. Would you be interested in contributing a PR for this feature?

KingSora:
@shellscape great! You can find the PR here: #784 😃

PR #784, opened 2021-01-31
feat(babel): pass rollup context as this context into override config function

KingSora:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

Implements my suggestion from https://github.com/rollup/plugins/issues/774
Basically its now possible that you can pass information from the rollup context (like `getCombinedSourcemap()`) into any babel plugin with the `createBabelInputPluginFactory`, because the rollup context is now passed as `this` context into the `override config` function. (like its done in the `override result` function).

Also updated `index.d.ts` for typescript users.


shellscape (approved):

shellscape:
@Andarist any thoughts on this one?

packages/babel/test/as-input-plugin.js
```diff
@@ -377,6 +377,7 @@ test('supports customizing the loader', async (t) => {
   const customBabelPlugin = createBabelInputPluginFactory(() => {
     return {
       config(cfg) {
+        t.true(typeof this === 'object', 'override config this context is rollup context');
```
Andarist, line 380: could u test smth more distinct? it's rather easy to regress based on this simple check

packages/babel/types/index.d.ts
```diff
@@ -53,10 +53,12 @@ export type RollupBabelCustomOutputPluginOptions = (
   pluginOptions: RollupBabelOutputPluginOptions;
 };
 export type RollupBabelCustomPluginConfig = (
+  this: PluginContext | TransformPluginContext,
```
Andarist, line 56: I don't think this should be a union - this only gets called in a single place in the code and thus the type should be always the same. Since it's called from within a `transform` hook I would suspect that this can be simplified to:
```suggestion
  this: TransformPluginContext,
```

Andarist:
@shellscape looks good overall, I've left just some minor comments - I would actually treat this as a bug fix (this was always the intention, it has regressed during some refactors).

packages/babel/types/index.d.ts
KingSora: I did this because the `createBabelOutputPluginFactory` calls the `transformCode` function in the `renderChunk` function where rollup passes only the `PluginContext`. 
Only the `createBabelInputPluginFactory` calls `transformCode` in the `transform` function where rollup passes the `TransformPluginContext`.

packages/babel/test/as-input-plugin.js
KingSora: @Andarist do you have an idea how to do that? Unfortunately I'm no expert in rollup, so I don't know wether I can do something like `instanceof` or similar.

Andarist: just maybe try to check some example property/methods that exist on this object?

packages/babel/types/index.d.ts
Andarist: Good catch. It seems though that we should not reuse the `RollupBabelCustomPluginResult` and `RollupBabelCustomPluginConfig` types for both `RollupBabelCustomInputPlugin` and `RollupBabelCustomOutputPlugin`. I think we should split those so the `this` could be typed correctly for both cases

packages/babel/test/as-input-plugin.js
KingSora: @Andarist alright, done.. Wanted to do this initially, but then I thought such checks could also be a bit shaky.
3b0e78b

packages/babel/types/index.d.ts
KingSora: @Andarist Changed things up a bit in: 3b0e78b Please take a look

Andarist (approved):

shellscape:
@Andarist I kept this marked as a feat because from a high level view it really looks like more of a feature that contains a bugfix (it happens). On the version graph, there's not much difference there so I think we're safe. If you disagree with that, I can always force the commit back to a `fix`

PR #785, opened 2021-01-31
docs(replace): clarify replacement. fixes #737

### LINKED ISSUE #737: [plugin-replace] The docs are confusing
- Rollup Plugin Name: plugin-replace
- Rollup Plugin Version: ^2.3.4

Documentation Is:

- [ ] Missing
- [ ] Needed
- [x] Confusing
- [x] Not Sure?

### Please Explain in Detail...

The docs of `plugin-replace` state that it's possible to directly use primitives as values. This gives the false impression that primitives values will be injected as is which is wrong.

This is the example given in the docs:

```
    replace({
      __buildEnv__: 'production',
      __buildDate__: () => new Date(),
      __buildVersion: 15
    })
```

In this case `__buildEnv__` will not be replaced by the string `production` in the bundled code as the docs seem to imply. What will actually happen is that the bundled code will be looking for a `production` variable.

To inject primitive strings either `'"production"'` (with two pairs of quotes) or `JSON.stringify('production')` should be used.

After that the docs state that `For complex values, use JSON.stringify.` which is also confusing as non-complex values do need to be *stringified* too.

### Your Proposal for Changes

Not sure if this is the result of a change in the plugin itself or the docs have been always like this.

I can do a PR with changes to the docs to make this abundantly clear and prevent future confusion.

shellscape:
## Rollup Plugin Name: `replace}`

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

This clarifies the function of the replace plugin, based on the concern in #737

### DIFF
--- a/packages/replace/README.md
+++ b/packages/replace/README.md
@@ -9,7 +9,7 @@
 
 # @rollup/plugin-replace
 
-🍣 A Rollup plugin which replaces strings in files while bundling.
+🍣 A Rollup plugin which replaces targeted strings in files while bundling.
 
 ## Requirements
 
@@ -34,43 +34,25 @@ export default {
   input: 'src/index.js',
   output: {
     dir: 'output',
-    format: 'cjs'
+    format: 'cjs',
   },
   plugins: [
     replace({
-      __buildEnv__: 'production',
-      __buildDate__: () => new Date(),
-      __buildVersion: 15
-    })
-  ]
+      'process.env.NODE_ENV': JSON.stringify('production'),
+      __buildDate__: () => JSON.stringify(new Date()),
+      __buildVersion: 15,
+    }),
+  ],
 };
 ```
 
 Then call `rollup` either via the [CLI](https://www.rollupjs.org/guide/en/#command-line-reference) or the [API](https://www.rollupjs.org/guide/en/#javascript-api).
 
-The configuration above will replace every instance of `__buildEnv__` with `'production'` and `__buildDate__` with the result of the given function in any file included in the build. _Note: Values have to be either primitives or functions that return a string. For complex values, use `JSON.stringify`._
+The configuration above will replace every instance of `process.env.NODE_ENV` with `"production"` and `__buildDate__` with the result of the given function in any file included in the build.
+
+_Note: Values must be either primitives (e.g. string, number) or `function` that returns a string. For complex values, use `JSON.stringify`. To replace a target with a value that will be evaluated as a string, set the value to a quoted string (e.g. `"test"`) or use `JSON.stringify` to preprocess the target string safely._
 
 Typically, `@rollup/plugin-replace` should be placed in `plugins` _before_ other plugins so that they may apply optimizations, such as dead code removal.
-
-The most popular case is replacing process.env.NODE_ENV with development or production environment.
-
-```js
-import replace from '@rollup/plugin-replace';
-
-export default {
-  input: 'src/index.js',
-  output: {
-    dir: 'output',
-    format: 'cjs'
-  },
-  plugins: [
-    replace({
-      // alternatively, one could pass process.env.NODE_ENV or 'development` to stringify
-      'process.env.NODE_ENV': JSON.stringify('production')
-    })
-  ]
-};
-```
 
 ## Options
 


TrySound (approved):

packages/replace/README.md
```diff
@@ -34,43 +34,25 @@ export default {
...
-The configuration above will replace every instance of `__buildEnv__` with `'production'` and `__buildDate__` with the result of the given function in any file included in the build. _Note: Values have to be either primitives or functions that return a string. For complex values, use `JSON.stringify`._
-
-Typically, `@rollup/plugin-replace` should be placed in `plugins` _before_ other plugins so that they may apply optimizations, such as dead code removal.
-
-The most popular case is replacing process.env.NODE_ENV with development or production environment.
+The configuration above will replace every instance of `process.env.NODE_ENV` with `"production"` and `__buildDate__` with the result of the given function in any file included in the build.
 
-```js
-import replace from '@rollup/plugin-replace';
+_Note: Values must be either primitives (e.g. string, number) or `function` that returns a string. For complex values, use `JSON.stringify`. To replace a target with a value that will be evaluated as a string, set the value to a quoted string (e.g. `"test"`) or use `JSON.stringify` to preprocess the target string safely._
```
tjenkinson, line 53: Looks like internally it always creates a function so the function also could return primitives just like using a value

packages/replace/README.md
```diff
@@ -34,43 +34,25 @@ export default {
...
   output: {
     dir: 'output',
-    format: 'cjs'
+    format: 'cjs',
   },
   plugins: [
     replace({
-      __buildEnv__: 'production',
+      'process.env.NODE_ENV': JSON.stringify('production'),
       __buildDate__: () => new Date(),
```
tjenkinson, line 42: Should we json stringify this too?

tjenkinson (approved):
LGTM! Just a few comments

packages/replace/README.md
shellscape: eh yeah, but iirc we phrase it this way to encourage people not to. it's that as an "incase they do" thing

packages/replace/README.md
shellscape: negative. it's important to show that it's possible. there's an older issue where this was added. I can't remember if it was the old repo or this one, but someone asked for this explicitly.

shellscape: strike that - I think the old ask was assuming that the `__buildDate__` would be in some sort of resource file, but that's not obvious at all. 
```suggestion
      __buildDate__: () => JSON.stringify(new Date()),
```

packages/replace/README.md
```diff
@@ -34,43 +34,25 @@ export default {
...
   },
   plugins: [
     replace({
-      __buildEnv__: 'production',
-      __buildDate__: () => new Date(),
-      __buildVersion: 15
-    })
-  ]
+      'process.env.NODE_ENV': JSON.stringify('production'),
+      __buildDate__: () => new JSON.stringify(Date()),
```
tjenkinson, line 42: `new` in the wrong place

shellscape: shhhh :)

danielgindi:
Actually I think this requires much more clarification. As I got into trouble because of this plugin.
It replaces everywhere. Like for instance inside of strings.
So I did the classic example of replacing `process.env.NODE_ENV` => `JSON.stringify('production').
Now one of the modules had a string somewhere, that was kind of a comment that talked about `process.env.NODE_ENV`.
So the result was `var blah = "blah blah about "production" and more blah";`.
This broke the code and rollup could not continue parsing the code.

So it should be made very clear, or better yet- be transformed into a module that actually parses the code, and allows replacement inside strings vs non-string code fragments, or some other kind of resolution. 
Perhaps an option to replace specifically variables or any kind of "get" operation with something else, or to replace only function calls.

shellscape:
I'm all for another PR that expands upon this for more clarity. but let's not lose focus. this PR is explicitly for the issue mentioned and to address that user's concerns. that said, please do open another PR with the improvements you think should be in there

Issue #472, opened 2020-06-23
[node-resolve]: Support "browser" field external mappings


guybedford:
- Rollup Plugin Name: node-resolve
- Rollup Plugin Version: latest

### Feature Use Case

The package.json `"browser"` field supports external mappings of the form:

```js
{
  "browser" {
    "external": "newexternal",
    "./local.js": "external/subpath"
  }
}
```

For example see https://github.com/sodium-friends/sodium-universal/blob/master/package.json#L14

### Feature Proposal

By default the node-resolve plugin is currently handling these as local package paths instead of external resolutions as they should be.

volkanunsal:
This is very much needed.

shellscape:
@volkanunsal we'd happily review a pull request that fulfilled this request

guybedford:
I'm going to reopen this one as it seems like an important omission in the default resolution behaviour unless I've missed that it's been addressed already.

volkanunsal:
I looked at the source code to potentially provide a patch for it, but the resolution algorithm is so convoluted and written in an idiosyncratic way. I can't make head or tail of it.

PR #786, opened 2021-02-01
fix(node-resolve): fix "browser" field mappings.

DorianBlues:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: #472, #540

### Description

When `browser` option of node-resolve is `true`, the package.json browser filed should override the export map result. 
But I found that node-resolve not handled it correctly when use commonjs to require some pacakage. (e.g. nanoid)

### DIFF
--- a/packages/node-resolve/rollup.config.js
+++ b/packages/node-resolve/rollup.config.js
@@ -12,7 +12,7 @@ export default {
     throw new Error(warning);
   },
   output: [
-    { file: pkg.main, format: 'cjs', exports: 'named' },
+    { file: pkg.main, format: 'cjs', exports: 'named', sourcemap: true },
     { file: pkg.module, format: 'es', plugins: [emitModulePackageFile()] }
   ]
 };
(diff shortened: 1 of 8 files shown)


guybedford (changes requested):
I think we should modify this so that the "browser" exports should take precedence over the base-level "browser" field, as otherwise it will be too confusing when each will apply.

DorianBlues:
I found that the "browser" exports of package `nanoid` has optimization conditions: `development` and `production` which now are supported by webpack. Should we change the type of "browser" option of node-resolve to string so that we can support it.

guybedford:
@DorianBlues right, you can enable these via `"conditions": ["development"]` etc also being passed to the plugin.

It might also make sense for RollupJS to default to a development or production condition, for cases where these are treated as mutually exclusive by packages like nanoid. Note these are now all endorsed by Node.js fwiw (https://nodejs.org/dist/latest-v15.x/docs/api/packages.html#packages_conditions_definitions). //cc @LarsDenBakker

shellscape:
@DorianBlues please give this PR some attention. When PRs have no activity for 30 days, we typically close them as abandoned unless the author responds.

Issue #734, opened 2020-12-15
lint: inconsistent spacing before function parens


benmccann:
- Rollup Plugin Name: all
- Rollup Plugin Version: `master`

### Expected Behavior / Situation

Spacing is consistent throughout the codebase and `lint` does not change the code because the code all complies with the linting rules

### Actual Behavior / Situation

Spacing is inconsistent throughout the codebase and `lint` tries to add space before function parens such that we have `function ()`

### Modification Proposal

We can set spacing on each of three different configurations:

- anonymous for anonymous function expressions (e.g. function() {}).
- named for named function expressions (e.g. function foo() {}).
- asyncArrow for async arrow function expressions (e.g. async () => {}).

I would propose to use a space only for async arrow and not for `function`, but I'm happy to change this in anyway decided. I would hope that this can be done in one PR since it's only formatting and that I don't have to send a PR for each package resulting in about ~20 PRs.

shellscape:
This would be the place to do that: https://github.com/rollup/eslint-config-rollup. I'm for it 👍

tjenkinson:
Thoughts on prettier? Works great if you're happy with its choices it's nice to not have to worry about all the rules etc.

shellscape:
We use prettier as an ESLint plugin. https://github.com/rollup/eslint-config-rollup/blob/c4d81fb97d59fc1b16f209d5925001a75262babd/.eslintrc.json#L31-L38

tjenkinson:
Ah interesting. We might be able to simplify things a little bit and enforce prettier is happy with everything on CI. Might have a go at some point

tjenkinson:
"some point" being today :) https://github.com/rollup/plugins/pull/787

PR #787, opened 2021-02-01
chore(repo): Ensure prettier has run on every file it supports

tjenkinson:
## Rollup Plugin Name: `all`

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

List any relevant issue numbers: https://github.com/rollup/plugins/issues/734#issuecomment-770447967

### Description
This runs `prettier --check .` on CI, ensuring that every file that prettier supports has been run through prettier.

It also adds `eslint-config-prettier`, which is an eslint plugin which just disables any eslint rule that conflicts with prettier.

Finally it changes the repo so that the same prettier rules are used for the README's as the code. This makes things a bit simpler. More work is needed if we want to keep separate configs. Also I was wondering if we should remove the `trailingComma` override. I added some comments inline.

Let me know what you think. It might be that you don't want everything run through prettier, in which case this might not be a simple solution and I'm fine closing this.


.prettierrc.json
```diff
@@ -0,0 +1,6 @@
+{
+  "arrowParens": "always",
+  "printWidth": 100,
+  "singleQuote": true,
+  "trailingComma": "none"
```
tjenkinson, line 5: For readmes previously had this set to 'always'. IMO the default `es5` (almost 'always') is quite nice because it reduces git diffs when adding new lines and helps with conflicts. Wonder if we could change it?

.prettierignore
```diff
@@ -1,3 +1,3 @@
-.github/ISSUE_TEMPLATE
-.github/ISSUE_TEMPLATE.md
-.github/PULL_REQUEST_TEMPLATE.md
```
tjenkinson, line 3: We can undo this if there is a reason prettier shouldn't touch these files :)

packages/alias/package.json
```diff
@@ -22,9 +22,9 @@
     "ci:lint:commits": "commitlint --from=${CIRCLE_BRANCH} --to=${CIRCLE_SHA1}",
     "ci:test": "pnpm run test -- --verbose",
     "lint": "pnpm run lint:js && pnpm run lint:docs && pnpm run lint:package",
-    "lint:docs": "prettier --single-quote --arrow-parens avoid --trailing-comma none --write README.md",
+    "lint:docs": "prettier --write README.md",
     "lint:js": "eslint --fix --cache src test types --ext .js,.ts",
-    "lint:package": "prettier --write package.json --plugin=prettier-plugin-package",
```
tjenkinson, line 27: the plugin should already be loaded automatically given it starts with `prettier-plugin-`

packages/alias/package.json
```diff
@@ -22,9 +22,9 @@
     "ci:lint:commits": "commitlint --from=${CIRCLE_BRANCH} --to=${CIRCLE_SHA1}",
     "ci:test": "pnpm run test -- --verbose",
     "lint": "pnpm run lint:js && pnpm run lint:docs && pnpm run lint:package",
-    "lint:docs": "prettier --single-quote --arrow-parens avoid --trailing-comma none --write README.md",
```
tjenkinson, line 25: Having different rules for readmes and the code made things more complicated, so I made them the same. If we want to keep them different more work is needed

.eslintrc.js
```diff
@@ -37,22 +37,13 @@ module.exports = {
     'no-unused-vars': 'off',
     'prefer-object-spread': 'off',
     'spaced-comment': 'off',
-    'prettier/prettier': [
```
shellscape, line 40: Do prettier's defaults line up with the rules that were removed below?

tjenkinson: I moved these rules into the new .prettierrc.json file (except for `plugins`, because that option doesn't exist and the docs say it will be auto loaded)

shellscape: Ah, I much prefer them in the ESLint config. It'd be sweet if you could move those back. Also, since we're already extending eslint-config-rollup, you're going to get some conflicts with https://github.com/rollup/eslint-config-rollup/blob/c4d81fb97d59fc1b16f209d5925001a75262babd/.eslintrc.json#L31-L38

If anything, we should be making changes there, and cleaning up the extension of that config here.

tjenkinson: >Also, since we're already extending eslint-config-rollup, you're going to get some conflicts

Ah yep.

I moved it to the separate file because otherwise the direct `prettier` runs (like the `prettier --check .`) won't see it 🤔

shellscape: hm, good point. maybe prettier has extendability now, I'll take a look

shellscape: https://prettier.io/docs/en/configuration.html#sharing-configurations

so what we'll want to do is add the prettier config to `eslint-config-rollup` and export that. and then use `.prettierrc.js` to consume that locally in the repo here. then I think we'll be in a manageable spot without code/config dupes.

.github/ISSUE_TEMPLATE/BUG.md
```diff
@@ -38,13 +37,10 @@ about: Something went awry and you'd like to tell us about it.
   ⚠️ ZIP Files are unsafe and maintainers will NOT download them.
 -->
 
-
 ### Expected Behavior
 
-
```
shellscape, line 44: these extra lines were intentional (41, 44, 47)

.github/ISSUE_TEMPLATE/DOCS.md
```diff
@@ -32,5 +31,4 @@ Documentation Is:
 
 ### Please Explain in Detail...
 
-
```
shellscape, line 35: intentional extra line

.github/ISSUE_TEMPLATE/FEATURE.md
```diff
@@ -23,5 +22,4 @@ about: Suggest an idea for this project
 
 ### Feature Use Case
 
-
```
shellscape, line 26: intentional extra line

.github/ISSUE_TEMPLATE/MODIFICATION.md
```diff
@@ -23,8 +22,6 @@ about: Would you like something work differently? Have an alternative approach?
 
 ### Expected Behavior / Situation
 
-
```
shellscape, line 26: intentional extra lines (26, 29)

.github/workflows/node-windows.yml
```diff
@@ -4,28 +4,27 @@ on: [push, pull_request]
...
-        node-version: ${{ matrix.node }}
-    - name: checkout master
-      run: git branch -f master origin/master
-    - name: install pnpm
-      run: npm install pnpm -g
-    - name: pnpm install
-      run: pnpm install --ignore-scripts
-    - name: run tests
-      run: pnpm run ci:test
+      - name: Configure git line-breaks
```
shellscape, line 15: are we sure that the indentation added here won't mess up the workflow?

.prettierignore
shellscape: probably the intentional extra lines that I added, which prettier removed

.vscode/launch.json
```diff
@@ -1,22 +1,16 @@
...
-            ],
-            "console": "integratedTerminal"
-        }
-    ]
-}
\ No newline at end of file
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
```
shellscape, line 5: we should remove this file from the repo imo. the config here isn't valid (since we use ava)

lint-staged.config.js
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable import/no-extraneous-dependencies */
+const micromatch = require('micromatch');
```
shellscape, line 2: why do we need this file? I'm generally not in favor of adding more boilerplate code to have to maintain when the configuration in package.json has been sufficient thus far

.eslintrc.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-  extends: ['rollup', 'plugin:import/typescript'],
+  extends: ['rollup', 'plugin:import/typescript', 'prettier'],
```
shellscape, line 2: given, https://github.com/rollup/plugins/pull/787/files#r568671377, let's remove the `prettier` preset. I'd rather have us set our rules up rather than battling defaults from another package.

.eslintrc.js
```diff
@@ -37,22 +37,13 @@ module.exports = {
...
     'prefer-object-spread': 'off',
     'spaced-comment': 'off',
-    'prettier/prettier': [
-      'error',
-      {
-        arrowParens: 'always',
-        printWidth: 100,
-        singleQuote: true,
-        trailingComma: 'none',
-        plugins: ['prettier-plugin-package']
```
shellscape, line 47: I see that `prettier-plugin-package` has been removed everywhere, but it hasn't been reimplemented anywhere. this is a hard block for me.

.github/CODE_OF_CONDUCT.md
```diff
@@ -8,19 +8,19 @@ In the interest of fostering an open and welcoming environment, we as contributo
 
 Examples of behavior that contributes to creating a positive environment include:
 
-* Using welcoming and inclusive language
```
shellscape, line 11: this is a good opportunity for us to centralize the CoC. let's nuke this doc and refer folks to Rollup's main project CoC

```md
This repository has adopted and follows the primary Rollup Code of Conduct https://github.com/rollup/rollup/blob/master/CODE_OF_CONDUCT.md
```

packages/alias/package.json
shellscape: ah didn't see this comment. in my tests with the plugins system I didn't find that to be the case. that's why we had the flag set all over the place.

.prettierignore
tjenkinson: cool will being these back 👍

packages/alias/package.json
tjenkinson: I'll double check the plugin is actually loaded

.github/workflows/node-windows.yml
tjenkinson: I'd hope prettier wouldn't reformat something in a way that changes its meaning. Looking at the changes now I can't see why there'd be a problem

.vscode/launch.json
tjenkinson: ok will delete it 👍

lint-staged.config.js
tjenkinson: Yeh this is the only way to have prettier run on all the format it supports (`prettier
  .getSupportInfo()`) without hardcoding a static list of extensions.

.eslintrc.js
tjenkinson: see the other comment. I'll check it's actually being loaded, and if not will bring this back where needed

.eslintrc.js
tjenkinson: Sure . Or counter proposal, we remove prettier from the eslint configs completely and just use eslint for code quality, and then prettier can just do it's thing as part of the commit hook?

https://prettier.io/docs/en/comparison.html

Thoughts?

lint-staged.config.js
shellscape: I would feel a lot better about a static list of extensions.

.github/workflows/node-windows.yml
shellscape: cool. I'm not very knowledgable about YAML syntax so I wanted to make sure.

.eslintrc.js
shellscape: prettier and ESLint fight each other quite a bit and I very much dislike running the two tools separately. I've found the integration of the two to be a far superior solution, so I'd like to keep it as is.

tjenkinson: @shellscape I opened a pr to expose the prettier config: https://github.com/rollup/eslint-config-rollup/pull/4

packages/alias/package.json
thorn0: Autoload doesn't work with pnpm. You can try specifying the plugin name in `.prettierrc`: `"plugins": "prettier-plugin-package"`.

packages/alias/package.json
thorn0: See https://prettier.io/docs/en/configuration.html#configuration-overrides

shellscape:
This is superseded by the combined work on https://github.com/rollup/eslint-config-rollup/pull/4 and some subsequent commits here. Huge thanks to @tjenkinson for doing the bulk of the work and discovery on this aspect. Apologies once again for how long this took to get over the finish line.

PR #793, opened 2021-02-05
feat(typescript): better error when tslib is not installed

### LINKED ISSUE #598: Clearer error if tslib is not installed
- Rollup Plugin Name: 
- Rollup Plugin Version:

### Feature Use Case
The error message if TypeScript tries to find tslib is a little unclear, and has [caused some user confusion](https://github.com/rollup/plugins/issues/243#issuecomment-670149948).

We explicitly import the `tslib.es6.js` subfile to avoid a dependency on the node resolve plugin, so this makes the error more confusing. 

### Feature Proposal
Change error message to something like `Cannot find module 'tslib', is it installed?`

shellscape:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: Fixes #598

### Description

This PR adds a meaningful error for users that addresses #598. Of note, this switches the use of the async `resolve` method to `resolve.sync`. It was just too problematic dealing with promises and getting the error out of the promise chain and back out to the user properly blocking the build. That may actually be an issue with plugin architecture, but we have an easy workaround for it now. Testing tslib being completely missing was a challenge, and I ultimately resorted to a janky envar for simulating that (notes in code comments) after trying a dozen or so methods for removing it (or simulating it being gone) from local node_modules.


packages/typescript/src/options/plugin.ts
```diff
@@ -14,7 +14,7 @@ import getTsLibPath from '../tslib';
  * - `typescript`: Instance of Typescript library (possibly custom).
  * - `tslib`: ESM code from the tslib helper library (possibly custom).
  */
-export default function getPluginOptions(options: RollupTypescriptOptions) {
+export const getPluginOptions = (options: RollupTypescriptOptions) => {
```
NotWoods, line 17: Did we change linter rules to ban default exports?

tsconfig.base.json
```diff
@@ -9,6 +9,7 @@
     "noEmitOnError": false,
     "noUnusedLocals": true,
     "noUnusedParameters": true,
+    "resolveJsonModule": true,
```
NotWoods, line 12: I don't think we need to change this in the base config, most plugins don't use JSON imports

packages/typescript/types/index.d.ts
```diff
@@ -38,40 +70,21 @@ export interface RollupTypescriptPluginOptions {
...
+export interface FlexibleCompilerOptions extends CompilerOptions {
+  [option: string]: CompilerOptionsValue | TsConfigSourceFile | undefined | any;
 }
 
 /** Properties of `CompilerOptions` that are normally enums */
 export type EnumCompilerOptions = 'module' | 'moduleResolution' | 'newLine' | 'jsx' | 'target';
 
 /** JSON representation of Typescript compiler options */
-export type JsonCompilerOptions = Omit<CompilerOptions, EnumCompilerOptions> &
+export type JsonCompilerOptions = Omit<FlexibleCompilerOptions, EnumCompilerOptions> &
```
NotWoods, line 81: The type changes here appear unnecessary, and throws out a lot of type-checking for typos. I'd rather leave it as is.

packages/typescript/src/tslib.ts
```diff
@@ -1,13 +1,27 @@
...
-export default function getTsLibPath() {
-  return resolveIdAsync('tslib/tslib.es6.js', { basedir: __dirname });
-}
+export const getTsLibPath = () => {
+  // Note: This isn't preferable, but we've no other way to test this bit. Removing the tslib devDep
+  //       during the test run doesn't work due to the nature of the pnpm flat node_modules, and
+  //       other workspace dependencies that depenend upon tslib.
+  try {
+    // eslint-disable-next-line no-underscore-dangle
+    return resolveId(process.env.__TSLIB_TEST_PATH__ || 'tslib/tslib.es6.js', {
```
NotWoods, line 21: I wonder if we should pass a flag into `getTsLibPath` instead?

NotWoods (changes requested):

packages/typescript/src/options/plugin.ts
shellscape: Just changed the eslint config so that it didn't demand default exports. I hate em.

tsconfig.base.json
shellscape: doesn't hurt anything, but it is left over from a previous attempt at removing tslib for the test.

packages/typescript/types/index.d.ts
shellscape: Unfortunately it necessary for everything to play nicely together. The tests weren't written in TypeScript, and as soon as  I started passing `tslib` as an option in a TypeScript test, it started throwing typing errors. Going to have to overrule you on this one, unless you can get the types playing nice in a more visually pleasing manner.

packages/typescript/src/tslib.ts
shellscape: Considered it, but because the call to `getTsLibPath` is buried in the call chain, it was more trouble than it was worth. For as ugly as it is, it's effective and doesn't modify the public API.

packages/typescript/src/options/plugin.ts
NotWoods: Cool, lots of the default exports here were only to appease the linter :)

packages/typescript/types/index.d.ts
NotWoods: OK, I can always try taking a look at this again in a future PR

NotWoods (approved):

packages/typescript/types/index.d.ts
shellscape: Sorry for laying that out tersely, super tired over here. The blame is not yours, so didn't mean to imply that. Just how the types ended up, we got that funky collision. We'd probably have to use an alternate option name like `tslibPath` or something.

PR #798, opened 2021-02-08
feat(replace): prevent accidental replacement within assignment

danielroe:
## Rollup Plugin Name: `@rollup/replace`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

It's possible when replacing `process.env` values that packages may be writing to these values. This PR adds a negative lookahead test to the generated RegExp to prevent some of these situations.


shellscape:
Thanks for this PR. Could you go into some more detail as to how this change works and the problem that it solve? Examples would be wonderful. (This will be important information for context for folks that may be looking at this PR via blame)

danielroe:
@shellscape This PR will prevent replacing strings where they are followed by a single equals sign. For example, where the plugin is called as follows:
```js
replace({
  values: {
    'process.env.DEBUG': 'false'
  }
})
```

Observe the following code:
```js
// Input
process.env.DEBUG = false
if (process.env.DEBUG == true) {
  //
}
// Before this PR
false = false // this throws an error because false cannot be assigned to
if (false == true) {
  //
}
// After this PR
process.env.DEBUG = false
if (false == true) {
  //
}
```

shellscape:
Thanks for the explanation. That's some pretty specific behavior, and not one that I'm sure we want to enable by default. That's one of those "ecosystem gotchas" that I'm sure you're aware of: users rely on a specific behavior and structure code accordingly, only to have that behavior "fixed" in a patch version, causing chaos. I see the value in this change, and it was also raised earlier today in #803. This is a pretty big change that could break some behaviors that folks are relying on, and I wouldn't want to hamstring those folks with no way back, so to speak.

That said, I think it'd be reasonable to make this an option of some sort, defaulting to the original behavior. I would even be willing to output a warning if the option was `false` or original behavior, indicating that would change in the next major. I wouldn't remove the option in the next major, but just flip the bit.

danielroe:
@shellscape That seems very reasonable. I've added a `preventAssignment` option and warning, and happy to rename option or edit warning as you like; just let me know.

packages/replace/README.md
```diff
@@ -65,6 +65,13 @@ Default: `['\b', '\b']`
 
 Specifies the boundaries around which strings will be replaced. By default, delimiters are [word boundaries](https://www.regular-expressions.info/wordboundaries.html). See [Word Boundaries](#word-boundaries) below for more information.
 
+### `preventAssignment`
+
+Type: `Boolean`<br>
+Default: `false`
+
+Prevents replacing strings where they are followed by a single equals sign (e.g. `process.env.DEBUG = false`).
```
shellscape, line 73: let's insert the same example from the PR here for quick reference for users. when we flip the bit on this feature, a good example in the README will be important.

packages/replace/src/index.js
```diff
@@ -37,14 +37,22 @@ function mapToFunctions(object) {
...
+  const { delimiters, preventAssignment } = options;
   const functionValues = mapToFunctions(getReplacements(options));
-  const keys = Object.keys(functionValues)
-    .sort(longest)
-    .map(escape);
+  const keys = Object.keys(functionValues).sort(longest).map(escape);
+  if (![true, false].includes(preventAssignment)) {
+    // eslint-disable-next-line no-console
+    console.warn(
+      "replace: 'preventAssignment' currently defaults to false. It is recommended to configure this option explicitly, as it may default to true in a future major version."
```
shellscape, line 46:
```suggestion
      "@rollup/plugin0-replace: 'preventAssignment' currently defaults to false. It is recommended to set this option to `true`, as the next major version will default this option to `true`."
```

packages/replace/src/index.js
```diff
@@ -37,14 +37,22 @@ function mapToFunctions(object) {
...
-  const { delimiters } = options;
+  const { delimiters, preventAssignment } = options;
   const functionValues = mapToFunctions(getReplacements(options));
-  const keys = Object.keys(functionValues)
-    .sort(longest)
-    .map(escape);
+  const keys = Object.keys(functionValues).sort(longest).map(escape);
+  if (![true, false].includes(preventAssignment)) {
+    // eslint-disable-next-line no-console
+    console.warn(
```
shellscape, line 45:
```suggestion
    this.warn(
```

we'll use the plugin system's warning mechanism

packages/replace/src/index.js
```diff
@@ -37,14 +37,22 @@ function mapToFunctions(object) {
...
 export default function replace(options = {}) {
   const filter = createFilter(options.include, options.exclude);
-  const { delimiters } = options;
+  const { delimiters, preventAssignment } = options;
   const functionValues = mapToFunctions(getReplacements(options));
-  const keys = Object.keys(functionValues)
-    .sort(longest)
-    .map(escape);
+  const keys = Object.keys(functionValues).sort(longest).map(escape);
+  if (![true, false].includes(preventAssignment)) {
```
shellscape, line 43: I think we'd rather warn if `preventAssignment` is `false|undefined`, bc the user needs to know that it'll be changing. Even if it is a bit noisy.

danielroe: Personally, I would prefer allowing the user to set their preference and not overriding it even if the default does change. As you say, it will be very noisy and there is no way to opt out of the warnings, even though (if the user has explicitly set the behaviour to false) there will be no breaking changes.

shellscape: Fair enough

shellscape: Given that, let's add a test to assert the warning is being thrown if not explicitly set. This is one of the plugins that are next up for TypeScript conversion, and I'd like to guard against regression there. e.g. https://github.com/rollup/plugins/blob/990ad2f0664460078ccca100be93daea2fbbbae4/packages/html/test/test.js#L63-L73

shellscape:
thanks!

shellscape (approved):

acmiyaguchi:
What is the process of getting an npm package deployed with new feature changes like this one? This feature is relevant to some issues I'm facing while trying to replace `process.browser` in certain packages.

shellscape:
@acmiyaguchi patience. you exercise patience. please refrain from similar replies in the future, it just creates noise.

bdkjones:
I like this change a lot, but I do have a suggestion:

Because users will now receive a warning if they do not explicitly configure this option, the examples in README should be updated to show explicit configuration of this option in all cases. Someone following the current examples will be surprised at the warning.

shellscape:
@bdkjones good suggestion. would you be willing to open a PR?

bdkjones:
@shellscape Sure. I would use `true` in the examples even though the default is `false` because I think true is the better value and false is the default simply to avoid breaking changes. Thoughts?

PR #818, opened 2021-02-24
chore(typescript): bump TypeScript version

benmccann:

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [X] other

Are tests included?

- [ ] no

Breaking Changes?

- [X] no

List any relevant issue numbers: https://github.com/rollup/plugins/issues/287

### Description

Closes https://github.com/rollup/plugins/issues/287 now that TypeScript 4.2 has come out


benmccann:
Hmm. It seems that some other test is failing when we upgrade TypeScript. I'm not sure which one it is though because the error message from Ava is terrible.

benmccann:
I've updated this PR to revert https://github.com/rollup/plugins/pull/794. It was making all tests fail with TypeScript 4.2. It also seems unnecessary because the PR description says "The issue is that without a directory from which to read the tsconfig.json file, and without those root options, TS doesn't know where to start things off." However, that's exactly the issue that TypeScript 4.2 is fixing.

shellscape (approved):

shellscape:
@benmccann some comments in #287 have alluded to 4.2 not fixing that particular issue, so we'll need to keep an eye out for use cases which aren't covered by this. I'm good with reverting the preflight check for now.

benmccann:
Thanks for the invite. Unfortunately I didn't see it in time and the link is only valid for seven days

shellscape:
@benmccann I'll send another. please do keep an eye out.

benmccann:
thanks. got it. though I'm not sure I'll have much more time to contribute than I have historically (plus Rich ripped out TypeScript from SvelteKit so not really using this plugin anymore)

dstaley:
Just to verify: I need to wait for a release of this plugin > 8.2.0 before #287 might be resolved, correct? If it isn't, should I file a new issue since #287 was closed/locked?

beetaa:
> Just to verify: I need to wait for a release of this plugin > 8.2.0 before #287 might be resolved, correct? If it isn't, should I file a new issue since #287 was closed/locked?

Looking forward to see the result of #287 , too.

beetaa:
> > Just to verify: I need to wait for a release of this plugin > 8.2.0 before #287 might be resolved, correct? If it isn't, should I file a new issue since #287 was closed/locked?
> 
> Looking forward to see the result of #287 , too. It would be nice if the next version can be published to npm as soon as possible. Thanks so much.

@shellscape @benmccann waiting for publishing to npm, thanks so much for your jobs.

Issue #769, opened 2021-01-17
Feature request for @rollup/plugin-run: Allow manual run via hotkey


mnpenner:
- Rollup Plugin Name: @rollup/plugin-run
- Rollup Plugin Version: 2.0.2

### Feature Use Case

`@rollup/plugin-run` works really great for node servers to restart them after editing some source files, but it doesn't work quite as well for CLI scripts. Particularly scripts that take a long time to run or read/write a lot files, I don't want them to automatically re-run after every little edit, but I *do* want a way to quickly re-run them when I'm ready.

### Feature Proposal

Instead of re-running as soon as rollup finishes recompiling, let the user choose when to run by waiting for a keypress, like <kbd>r</kbd>. Maybe make it an option like `{ runOnKeyPress: 'r' }`

shellscape:
Thanks for the suggestion. We haven't seen a call for a feature like that before, so the demand isn't very high. The code for the run plugin is pretty simple, clocking in at a measly 58 lines and I think your best bet is to copy the source, modify it to your needs, and use it as a local plugin for your builds. I'm going to mark this as `wontfix` for now, but we'll leave it open for a few more weeks to let any interested folks weigh in.

nicholas-ochoa:
@shellscape Would you accept a pull request that implements this functionality? This is similar to functionality that I've missed since switching from using webpack / nodemon to rollup and this plugin for development.

For example, testing configuration file changes that aren't watched (they're not part of the source code) thus far has mean CTRL+C and then re-running the whole build/run process.

shellscape:
@nicholas-ochoa we'd probably have to see how much overhead and complexity that would add. it looks like you might be a windows user, and this will have to be cross-platform compatible.

nicholas-ochoa:
I've created a quick proof of concept that mimics the functionality of nodemon (typing "rs" and pressing enter kills / restarts the bundle). It looks like the functionality that @mnpenner describes might not be easily doable with the existing plugin hooks however. Adding this functionality only adds about a dozen lines of code, although I still need to consider how to place this functionality behind a configuration flag.

While I do some development in Windows, most of the software I develop targets Linux so cross platform support is always in mind.

I'll continue working on it as I have time and submit a pull request for review to see if this is something worth adding.

mnpenner:
I have a WIP here btw: https://github.com/mnpenner/rollup-plugins/tree/default/packages/rollup-plugin-run It

[...]

PR #820, opened 2021-02-24
feat(run): add `allowRestarts` option

nicholas-ochoa:
## Rollup Plugin Name: `run`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

#769 

### Description
This PR adds functionality to enable optional manual restarts of the running bundle by entering `rs`, `restart` or pressing <kbd>Ctrl</kbd>+<kbd>K</kbd> on the terminal followed by <kbd>Enter</kbd>. This mimics the behavior of [nodemon](https://github.com/remy/nodemon#manual-restarting). Additionally, the terminal screen can be cleared by entering `cls`, `clear` or pressing <kbd>Ctrl</kbd>+<kbd>L</kbd> followed by <kbd>Enter</kbd>.

The use case for this functionality is developing applications that reference external configuration files, resources or databases that may change while the application source remains unchanged. When these are updated often times the application will need to be restarted. In my case, this is currently accomplished by terminating the current instance using <kbd>Ctrl</kbd>+<kbd>C</kbd> and then restarting it by running `pnpm run dev`.

I've tested this on Windows (under `cmd.exe` as well as under Bash using Windows Terminal) and Linux (bash, over SSH using Putty, as well as using `ssh` under Windows Terminal) and it works as expected.

I did add a single test case but I admit it is pretty simplistic and could use improvement. I was unable to get `sinon` to properly mock the `process.stdin.resume` function call to determine whether or not it was called.

### DIFF
--- a/packages/run/README.md
+++ b/packages/run/README.md
@@ -32,9 +32,9 @@ export default {
   input: 'src/index.js',
   output: {
     file: 'dist/index.js',
-    format: 'cjs'
+    format: 'cjs',
   },
-  plugins: [run()]
+  plugins: [run()],
 };
 ```
 
@@ -71,6 +71,19 @@ export default {
   ]
 };
 ```
+
+### `allowRestarts`
+
+Type: `Boolean`<br>
+Default: `false`
+
+If `true`, instructs the plugin to listen to `stdin` for the sequences listed below followed by enter (carriage return).
+
+`restart` - Kills the currently running bundle and starts it again. _Note: This does not create a new bundle, the bundle is run again "as-is". This can be used to test configuration changes or other changes that are made without modifying your source_
+Also allowed: `rs`, `CTRL+K`
+
+`clear` - Clears the screen of all text
+Also allowed: `cls`, `CTRL+L`
 
 ## Practical Example
 
--- a/packages/run/README.md
+++ b/packages/run/README.md
@@ -79,11 +79,15 @@ Default: `false`
 
 If `true`, instructs the plugin to listen to `stdin` for the sequences listed below followed by enter (carriage return).
 
-`restart` - Kills the currently running bundle and starts it again. _Note: This does not create a new bundle, the bundle is run again "as-is". This can be used to test configuration changes or other changes that are made without modifying your source_
-Also allowed: `rs`, `CTRL+K`
+#### `stdin` Input Actions
 
-`clear` - Clears the screen of all text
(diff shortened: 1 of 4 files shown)


packages/run/README.md
```diff
@@ -72,6 +72,19 @@ export default {
...
 ```
 
+### `allowRestarts`
+
+Type: `Boolean`<br>
+Default: `false`
+
+If `true`, instructs the plugin to listen to `stdin` for the sequences listed below followed by enter (carriage return).
+
+`restart` - Kills the currently running bundle and starts it again. _Note: This does not create a new bundle, the bundle is run again "as-is". This can be used to test configuration changes or other changes that are made without modifying your source_
```
shellscape, line 82: Im not sure what this and the `clear` bit below are describing. The type of this option is stated to be a `Boolean` so where does this come into play?

packages/run/src/index.ts
```diff
@@ -41,15 +43,34 @@ export default function run(opts: RollupRunOptions = {}): Plugin {
     },
 
     writeBundle(outputOptions, bundle) {
+      function forkBundle(dir: string, entryFileName: string) {
```
shellscape, line 46: please use a `const` arrow function

nicholas-ochoa: done

packages/run/README.md
shellscape: OK, thanks. It'll be worth delineating that a bit more:

```suggestion
#### `stdin` Input Actions

When this option is enabled, `stdin` will listen for the following input and perform the associated action:

- `restart` → Kills the currently running bundle and starts it again. _Note: This does not create a new bundle, the bundle is run again "as-is". This can be used to test configuration changes or other changes that are made without modifying your source_
```

(And please copy the `- key →` format to the `clear` bit below

shellscape (approved):
thanks!

packages/run/README.md
nicholas-ochoa: In hindsight, I wrote that with my own assumptions of how this functionality works in mind without considering how someone unfamiliar with it might interpret it. I've made the changes you suggested.

shellscape:
We generally like to get more than one review before merging. Just waiting for some others to have the time to check.

shellscape:
thanks!

PR #822, opened 2021-02-25
feat(html): assets inline injection support

yurks:
## Rollup Plugin Name: `html`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

Add `injectAssets` option for injecting CSS and JS assets as inline code into result HTML.


shellscape:
Thanks for creating this PR. For features, it's always a good idea to open a new feature issue to discuss if the project would like to accept a new feature for a plugin. Really appreciate your time and effort to put this together.

Unfortunately, I don't feel that this feature would be widely used, as it's not been mentioned here before iirc. This kind of modification is exactly why the `template` option is allowed to be so robust. If this is a pattern you're using frequently, I would recommend creating a utility package, or forking this plugin for your use case.

Since this has already been open for two weeks without comment, we'll leave this open for another few weeks for comments from other maintainers.

yurks:
Well, several years ago I thought rollup wouldn't be widely used as it's not been mentioned in my environment ;)
Appreciate you for your review time, it's not a problem at all to cancel this PR, of course I can handle this functionality in a way I prefer.
In my vision this is in line with Rollup concept for smaller and faster bundles: proposed `injectAssets` option is a possibility to build "truly single page" app with all stuff included in single html, and avoid to duplicate a huge `template` logic. Anyway, there is not so high costs to have this feature out-of-the-box)

shellscape:
Thanks again for putting this PR together and for your time. Unfortunately this didn't get any support from the other maintainers. If you'd like to leave your fork on your account, other users might see this closed PR and lend support to it later on, in which case we'd consider reopening the PR. Totally your call. In the meantime, we'll close this PR.

andreavaccari:
We'd use this feature if it were available. Would it make sense to add it to [`recipes`](https://github.com/rollup/plugins/tree/master/packages/html/recipes)?

shellscape:
@andreavaccari would definitely welcome additions to recipes.

PR #825, opened 2021-03-03
[dynamic-import-vars] Give default promise rejection timing parity with cases

bearfriend:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

The immediately rejected promise in the `default` clause of the generated `switch` does not have timing parity with the other clauses, making it impossible to register a `catch` on the promise before it is rejected, which causes debuggers with "pause on exceptions" (but not "pause on caught exceptions") enabled to always pause in the `default` clause, even if you have registered a `catch` on the returned promise. This queues the promise rejection as a microtask to remedy that situation.


shellscape:
@LarsDenBakker it looks like this has some community support. Please have a look over this when you have the opportunity.

LarsDenBakker:
This changes the natural behavior of dynamic imports, I'm not sure if it's a good idea. 

For example, given:

```js
import(`./pages/${page}.js`);
```

If the requested file does not exist, the browser will throw an error before bundling with rollup. Why should it not throw an error after bundling with rollup? 

If your code is intended to ignore/handle 404s, you can do this:

```js
try {
  import(`./pages/${page}.js`);
} catch (error) {
  // handle the error
}
```

That way the code behaves the same both before and after bundling.

LarsDenBakker:
You're right I'm missing an `await` in my example. 

Improving the way we handle the rejected promise is fine by me, I wasn't aware we didn't achieve this already with the current code.

I'm personally still not convinced of the use case you describe. If there is high demand for it, we can add it. But as far as I can see it's a few lines of code to handle the catch yourself and it makes the code more explicit and work without bundling as well...

bearfriend:
I think that's fair. If we queue the promise rejection as a microtask, making the default clause optional becomes less necessary, and it can be handled in a `catch` much more effectively. I probably should've split the methods into separate PRs to begin with.

Should I go ahead and revert the `defaultClause` option?

LarsDenBakker:
Sure, that would be fine!

LarsDenBakker (approved):

packages/dynamic-import-vars/README.md
```diff
@@ -134,7 +134,11 @@ function __variableDynamicImportRuntime__(path) {
     case './locales/nl-NL.js':
       return import('./locales/nl-NL.js');
     default:
-      throw new Error('Unknown variable dynamic import: ' + path);
+      return new Promise((resolve, reject) => {
+        queueMicrotask(() => {
+          reject(new Error("Unknown variable dynamic import: " + path))
```
tjenkinson, line 139: Sorry not reviewed this in detail, but wondering if `queueMicrotask` is supported in most places or if we should use something a bit older instead to achieve this?

Would a standard `return Promise.reject()` not work?

bearfriend: It would not. The actual code falls back to `setTimeout`.

tjenkinson: Looked in a bit more detail.

What wouldn’t work with `return Promise.reject(new Error("Unknown variable dynamic import: " + path))` work?

The return is always a promise then which would resolve/reject async. I understand why the sync throw  before was not great

tjenkinson (changes requested):
Makes sense to me, but I think we could just return a `Promise.reject(...)` and solve the problem with less code, so that would be preferable IMO

packages/dynamic-import-vars/README.md
bearfriend: The code actually already is using `Promise.reject`. The documentation just doesn't match. Please run this with your dev tools open and `Pause on exception` enabled to see what this fixes.

https://jsfiddle.net/g6u4femL/

Note that even though both promises are caught, it will still break on `promiseA`.

tjenkinson: Ah ok thanks for explaining.

Seems strange there’s not a simpler way but 👍

tjenkinson (approved):

packages/dynamic-import-vars/README.md
tjenkinson: Feels like this is something dev tools should handle? If you’re pausing on uncaught exceptions I wonder why it doesn’t wait a tick internally until the point when it actually rejects 🤔

packages/dynamic-import-vars/src/index.js
```diff
@@ -52,10 +52,13 @@ function dynamicImportVariables({ include, exclude, warnOnError } = {}) {
             // will turn these into chunks automatically
             ms.prepend(
               `function __variableDynamicImportRuntime${dynamicImportIndex}__(path) {
-   switch (path) {
-  ${paths.map((p) => `   case '${p}': return import('${p}');`).join('\n  ')}
-     default: return Promise.reject(new Error("Unknown variable dynamic import: " + path));
-   }
+  switch (path) {
+${paths.map((p) => `    case '${p}': return import('${p}');`).join('\n')}
+${`    default: return new Promise((resolve, reject) => {
```
tjenkinson, line 57: Should we use a standard function in here vs the arrow one?

tjenkinson (commented):
Sorry just spotted something else

packages/dynamic-import-vars/src/index.js
bearfriend: I suppose an arrow function isn't strictly necessary. Do you see a downside to it in this case? There's no scope it's ignoring.

tjenkinson: It depends what browsers this is targeting. Promises are polyfillable but the arrow syntax isn’t, so it would be safer to stick with the older function format I think.

bearfriend: Ah ok. Didn't realize we were going back that far with browser support, but yeah I guess it's the safest thing to do and no real downside.

shellscape:
@tjenkinson put your green checkmark on it and I'll merge it post haste

tjenkinson (approved):

PR #826, opened 2021-03-04
chore(repo): Add "directory" to package.json repository fields.

developit:
## Rollup Plugin Name: (all plugins)

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [ ] yes
- [x] no (non-functional change)

Breaking Changes?

- [ ] yes
- [x] no

### Description

This is a trivial thing, but every time I search for a Rollup plugin, the top results are npm package listing pages. That's great, but of the two sidebar links, one points to the plugins repo itself, whereas the other points to the package's README. This uses npm's [repository.directory package.json field](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#:~:text=you%20can%20specify%20the%20directory%20in%20which%20it%20lives) to make them both point to the correct location (see [npm cli source](https://github.com/npm/cli/blob/8806015fdd025f73ccf4001472370eafd3c5a856/lib/repo.js#L35-L48) that confirms the logic).

### DIFF
--- a/packages/alias/package.json
+++ b/packages/alias/package.json
@@ -6,7 +6,10 @@
   },
   "description": "Define and resolve aliases for bundle dependencies",
   "license": "MIT",
-  "repository": "rollup/plugins",
+  "repository": {
+    "url": "rollup/plugins",
+    "directory": "packages/alias"
+  },
   "author": "Johannes Stein",
   "homepage": "https://github.com/rollup/plugins/tree/master/packages/alias#readme",
   "bugs": "https://github.com/rollup/plugins/issues",
--- a/packages/auto-install/package.json
+++ b/packages/auto-install/package.json
@@ -6,7 +6,10 @@
   },
   "description": "Automatically install dependencies that are imported by a bundle",
   "license": "MIT",
-  "repository": "rollup/plugins",
+  "repository": {
+    "url": "rollup/plugins",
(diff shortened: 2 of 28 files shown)


shellscape:
@developit good to see you outside of twitter :)

The way this PR is structured, our publishing scripts won't pick up anything for each plugin. It needs an upgrade to a more robust recognition of changes, and to take advantage of a new pnpm feature since it was written, but like all things I lack the time at the moment. With the one commit, none of the plugins will pick this up for publish until a commit targeting each plugin has been merged.

Regardless, this _can_ play well with what we have, but the commits in this branch/PR need to be setup differently. We don't get these commits often (maybe 3 or 4 since the repo was conceived) so it hasn't been documented. Essentially, each plugin needs it's own commit. e.g. `chore(auto-install): add directory to package.json`. We'll then use a plain old merge commit for the PR and magically each plugin will see a change that'll allow us to run the publish scripts as they are. It's silly busy work, and I don't expect you to take the time to jump through that hoop. Without the hoop, you may not see your change for a bit in each plugin.

The choice is yours on which way to go with this, and we'll gladly accept either.

shellscape:
Welp, we'll merge this, but please be aware that it won't affect any plugins until they're published again. Which means most won't get this update published.

PR #834, opened 2021-03-12
fix(dynamic-import-vars): allow ./${var}.suffix.js

lennyburdette:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

Allow dynamically importing files in the same directory with an extra suffix before the extension.

My motivating example is trying to import a library generated with [Stencil](https://stenciljs.com/docs/introduction). It generates a dynamic import that looks like this:
```
return import(
    /* webpackInclude: /\.entry\.js$/ */
    /* webpackExclude: /\.system\.entry\.js$/ */
    /* webpackMode: "lazy" */
    `./${bundleId}.entry.js${ ''}`)
```

My interpretation of the [README](https://github.com/rollup/plugins/blob/master/packages/dynamic-import-vars/README.md) suggests this should be possible. I changed the assertion for ["Imports to your own directory must specify a filename pattern"](https://github.com/rollup/plugins/blob/master/packages/dynamic-import-vars/README.md#imports-to-your-own-directory-must-specify-a-filename-pattern) to a stricter regex. I'm happy to change the implementation if regexes aren't desired.

Thanks!

### DIFF
--- a/packages/dynamic-import-vars/src/dynamic-import-to-glob.js
+++ b/packages/dynamic-import-vars/src/dynamic-import-to-glob.js
@@ -86,10 +86,14 @@ export function dynamicImportToGlob(node, sourceString) {
     );
   }
 
-  if (glob.startsWith('./*.')) {
+  // Disallow ./*.ext
+  const ownDirectoryStarExtension = /^\.\/\*\.[\w]+$/;
+  if (ownDirectoryStarExtension.test(glob)) {
     throw new VariableDynamicImportError(
-      `${`invalid import "${sourceString}". Variable imports cannot import their own directory, ` +
-        'place imports in a separate directory or make the import filename more specific. '}${example}`
+      `${
+        `invalid import "${sourceString}". Variable imports cannot import their own directory, ` +
+        'place imports in a separate directory or make the import filename more specific. '
+      }${example}`
     );
   }
 
--- a/packages/dynamic-import-vars/test/src/dynamic-import-to-glob.test.js
+++ b/packages/dynamic-import-vars/test/src/dynamic-import-to-glob.test.js
@@ -17,6 +17,15 @@ test('template literal with variable filename', (t) => {
   t.is(glob, './foo/*.js');
 });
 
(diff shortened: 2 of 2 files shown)


LarsDenBakker:
This effectively disables the check. I think it's better to add an option which allows these kinds of imports, although they're quite dangerous because the glob will match your current file as well.

lennyburdette:
My assumption was that if `./mod-${id}.js` was allowed, then `./${id}.mod.js` should also be allowed — both technically allow matching the current file if it's named "mod-current.js" or "current.mod.js".

But! For my specific use-case I discovered that Stencil provides a [tree-shakable, non-lazy-loading build](https://stenciljs.com/docs/custom-elements), so I'm unblocked. Feel free to close or amend this however you like. Thanks for the review!

LarsDenBakker:
Ah I see, do we allow `./mod-${id}.js` currently? I guess that makes sense then.

lennyburdette:
[yep!](https://github.com/rollup/plugins/blob/master/packages/dynamic-import-vars/README.md#imports-to-your-own-directory-must-specify-a-filename-pattern)

```js
// not allowed
import(`./${foo}.js`);
// allowed
import(`./module-${foo}.js`);
```

what's tripping me up is that the Stencil file output is is `component-name.entry.js` — using a period makes the current `startsWith` think that a file extension is next, not a module name suffix.

shellscape:
@LarsDenBakker @danielgindi @tjenkinson gentle ping for review

shellscape:
@LarsDenBakker @danielgindi @tjenkinson another gentle ping. I'm inclined to move this one forward.

danielgindi:
Looks good to me and makes sense.

danielgindi (approved):

LarsDenBakker (approved):
LGTM as well. sorry haven't been spending a lot of time on open source lately

lennyburdette:
Thank you!

PR #835, opened 2021-03-12
Fix bug where fs.exists is incorrectly promisified

Hexxeh:
## Rollup Plugin Name: node-resolve

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

Using the node-resolve plugin in a browser where the fs module is mocked fails to find resolve any paths. Upon debugging this issue, it appears to be because `fs.exists` is passed to `util.promisify`. The Node [docs](https://nodejs.org/api/fs.html#fs_fs_exists_path_callback) explicitly call out this function as being non-standard and not following the pattern for `util.promisify` to work correctly.

As far as I can tell, this works on Node because `fs.exists` actually internally already provides a promisified version via: https://nodejs.org/api/util.html#util_custom_promisified_functions

Using stat, which doesn't have this non-standard-ness, works around this issue.

I haven't included tests because this fixes a bug when running in an admittedly quite strange setup that's currently not replicated in tests at all as far as I know.

### DIFF
--- a/packages/node-resolve/src/fs.js
+++ b/packages/node-resolve/src/fs.js
@@ -2,8 +2,16 @@ import fs from 'fs';
 
 import { promisify } from 'util';
 
-export const exists = promisify(fs.exists);
+export const access = promisify(fs.access);
 export const readFile = promisify(fs.readFile);
 export const realpath = promisify(fs.realpath);
 export { realpathSync } from 'fs';
 export const stat = promisify(fs.stat);
+export async function exists(filePath) {
+  try {
+    await access(filePath);
+    return true;
+  } catch {
+    return false;
+  }
+}


packages/node-resolve/src/fs.js
```diff
@@ -2,8 +2,15 @@ import fs from 'fs';
 
 import { promisify } from 'util';
 
-export const exists = promisify(fs.exists);
 export const readFile = promisify(fs.readFile);
 export const realpath = promisify(fs.realpath);
 export { realpathSync } from 'fs';
 export const stat = promisify(fs.stat);
+export async function exists(filePath) {
```
merceyz, line 9: Since `fs.exists` is just a wrapper around `fs.access` you should probably use that to make sure the behaviour is the same and to not unnecessarily create the stat object just to throw it away

https://github.com/nodejs/node/blob/5ff2e582cd867feba71bb13d9d8bed9953721145/lib/fs.js#L240-L252

packages/node-resolve/src/fs.js
```diff
@@ -2,8 +2,15 @@ import fs from 'fs';
...
+  try {
+    await stat(filePath);
```
TrySound, line 11: There is "access" function to check file existence.

TrySound:
Also looks like already promisified version are available in node 10
https://nodejs.org/api/fs.html#fs_fspromises_access_path_mode

danielgindi (approved):
Using stat (or even better, `access`) makes more sense anyway, as `exists` is kind of deprecated in node. It's recommended to use a more informative method that tells what the thing is (file/folder/other), if it's accessible to you, etc.

Hexxeh:
Swapped `stat` for `access` per comments. Thanks for the reviews!

lukastaegert (approved):

shellscape:
thanks!

PR #843, opened 2021-03-24
fix(node-resolve): fix "browser" and "exports.browser" field mappings

brizental:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

Hey folks, I encountered this error and noticed there was an open PR (#786) related to it that was getting stale. I built on top of that PR and attended to the review comments.

I did not attend to the comments related to the `production` and `development` conditions (https://github.com/rollup/plugins/pull/786#issuecomment-771294449, https://github.com/rollup/plugins/pull/786#issuecomment-771462019) as I believe they are unrelated and I really just want the `browser` fix right now. I'd be happy to take that on as a follow up, but I would love for this or the other PR to be merged sooner, if possible.

Cheers!

### DIFF
--- a/packages/node-resolve/test/browser.js
+++ b/packages/node-resolve/test/browser.js
@@ -208,4 +208,24 @@ test('pkg.browser can override the export map result', async (t) => {
   const { module } = await testBundle(t, bundle);
 
   t.is(module.exports, 'browser');
-})
+});
+
+test('exports.browser takes precedence over export map result, when browser:true', async (t) => {
+  const bundle = await rollup({
+    input: 'browser-exports-browser.js',
+    plugins: [nodeResolve({ browser: true }), commonjs()]
+  });
+  const { module } = await testBundle(t, bundle);
+
+  t.is(module.exports, 'browser');
+});
+
+test('exports.browser does not take precedence over export map result, when browser:false', async (t) => {
+  const bundle = await rollup({
+    input: 'browser-exports-browser.js',
+    plugins: [nodeResolve(), commonjs()]
+  });
+  const { module } = await testBundle(t, bundle);
+
+  t.is(module.exports, 'require');
+});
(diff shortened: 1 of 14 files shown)


packages/node-resolve/test/browser.js
```diff
@@ -199,3 +199,33 @@ test('pkg.browser with mapping to prevent bundle by specifying a value of false'
...
+});
+
+test('exports.browser does not take precedence over export map result, when browser:false', async (t) => {
+  const bundle = await rollup({
+    input: 'browser-exports-browser.js',
+    plugins: [nodeResolve(), commonjs()]
+  });
+  const { module } = await testBundle(t, bundle);
+
+  t.is(module.exports, 'require');
```
guybedford, line 230: Somewhat suprised to see this is returning `require` by default, I wasn't aware this was the case, it might be a separate issue if not introduced by this PR (in which case the previous comment should also be `require` I suppose). We should surely be defaulting to ESM for top-level imports.

guybedford (commented):
Thank you @brizental for the PR. I've left some feedback below.

packages/node-resolve/test/browser.js
brizental: Maybe it's because I included the `commonjs` plugin as well? That is copy-pasta by the way, I'll just try removing it :)

guybedford: I'm not sure, perhaps @LarsDenBakker can clarify what the expectation is for this case.

brizental: FWIW, removing the commonjs plugin made no difference.

guybedford: That makes sense, it's about the top-level resolver conditions, not the transforms.

guybedford:
Superceded by https://github.com/rollup/plugins/pull/866.

Issue #698, opened 2020-12-08
Incremental build not working


cdalexndr:
- Rollup Plugin Name: @rollup/plugin-typescript
- Rollup Plugin Version: 8.0.0
- Rollup Version: 2.34.2
- Operating System (or Browser): Win10
- Node Version: v10.16.0
- Link to reproduction _(⚠️ read below)_: https://github.com/cdalexndr/rollup-plugin-issue-698

### Expected Behavior
Compilation successfull.

### Actual Behavior
`Error: @rollup/plugin-typescript: Rollup 'dir' option must be used when Typescript compiler options 'tsBuildInfoFile' or 'incremental' are specified.`

### Additional Information
I have a working build configuration that outputs bundles.

When I add `incremental` to my `tsconfig.json`, the build fails with:
`Error: @rollup/plugin-typescript: Rollup 'dir' option must be used when Typescript compiler options 'tsBuildInfoFile' or 'incremental' are specified.`

If I add `dir: ".rollup.dir"` to my `rollup.config.js`, then build fails with another error:
`Error: You must set either "output.file" for a single-file build or "output.dir" when generating multiple chunks.`

Isn't incremental build supported? I expected this feature to be plug and play.

shellscape:
Thanks for opening an issue. Citing the issue template:

>   🚨 Issues WITHOUT a valid reproduction WILL BE CLOSED!
> 
>   Please provide one by:
>   1. Using the REPL.it plugin reproduction template at https://repl.it/@rollup/rollup-plugin-repro
>   2. Provide a minimal repository link (Read https://git.io/fNzHA for instructions).
>      Please use NPM for installing dependencies!
>      These may take more time to triage than the other options.
>   3. Using the Rollup REPL at https://rollupjs.org/repl/
> 
>   ⚠️ ZIP Files are unsafe and maintainers will NOT download them.

We cannot make this any clearer. Please add a reproduction and we'll be happy to triage further.

cdalexndr:
Added test repo in original post.

shellscape:
Thanks.

@NotWoods @benmccann please have a peek. https://github.com/rollup/plugins/pull/535 added support for this, but I'm not really able to give the user direction based on the tests that were added for incremental builds.

benmccann:
It looks like `incremental` is only supported with `dir` and not `file`

[...]

PR #859, opened 2021-04-15
Fix incremental typescript cache

bornskilled200:
Typescript

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] no

https://github.com/rollup/plugins/issues/698

### Description
After removing the validation, I get an error that rollup should not be emitting files outside of the `outDir` directory.

This seems strange as the typescript incremental cache file is not a user facing file. I removed the usage of rollup's `emitFile` and replaced it with how `TsCache` writes their files, which should be the closest equivalent.

### DIFF
--- a/packages/typescript/src/index.ts
+++ b/packages/typescript/src/index.ts
@@ -150,7 +150,9 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
         const tsBuildInfoSource = emittedFiles.get(tsBuildInfoPath);
         // https://github.com/rollup/plugins/issues/681
         if (tsBuildInfoSource) {
-          fs.writeFileSync(normalizePath(tsBuildInfoPath), tsBuildInfoSource);
+          const normalizedTsBuildInfoPath = normalizePath(tsBuildInfoPath);
+          fs.mkdirSync(path.dirname(normalizedTsBuildInfoPath), { recursive: true });
+          fs.writeFileSync(normalizedTsBuildInfoPath, tsBuildInfoSource);
         }
       }
     }
--- a/packages/typescript/src/index.ts
+++ b/packages/typescript/src/index.ts
@@ -1,5 +1,7 @@
 import * as path from 'path';
 
+import fs from 'fs';
+
 import { Plugin, RollupOptions, SourceDescription } from 'rollup';
 import type { Watch } from 'typescript';
 
@@ -148,11 +150,7 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
         const tsBuildInfoSource = emittedFiles.get(tsBuildInfoPath);
         // https://github.com/rollup/plugins/issues/681
         if (tsBuildInfoSource) {
-          this.emitFile({
-            type: 'asset',
-            fileName: normalizePath(path.relative(outputOptions.dir!, tsBuildInfoPath)),
-            source: tsBuildInfoSource
-          });
+          fs.writeFileSync(normalizePath(tsBuildInfoPath), tsBuildInfoSource);
         }
       }
     }
--- a/packages/typescript/src/index.ts
+++ b/packages/typescript/src/index.ts
@@ -87,7 +87,7 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
 
     renderStart(outputOptions) {
       validateSourceMap(this, parsedOptions.options, outputOptions, parsedOptions.autoSetSourceMap);
-      validatePaths(ts, this, parsedOptions.options, outputOptions);
+      validatePaths(this, parsedOptions.options, outputOptions);
     },
 
     resolveId(importee, importer) {
--- a/packages/typescript/src/index.ts
+++ b/packages/typescript/src/index.ts
@@ -25,7 +25,8 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
     transformers,
     tsconfig,
     tslib,
-    typescript: ts
+    typescript: ts,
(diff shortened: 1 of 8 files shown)


shellscape:
Thanks for the PR. We've just updated `master`. Please update your fork from `upstream/master`.

shellscape:
@NotWoods please have a peek at this one

packages/typescript/src/index.ts
```diff
@@ -148,11 +150,9 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
         const tsBuildInfoSource = emittedFiles.get(tsBuildInfoPath);
         // https://github.com/rollup/plugins/issues/681
         if (tsBuildInfoSource) {
-          this.emitFile({
-            type: 'asset',
-            fileName: normalizePath(path.relative(outputOptions.dir!, tsBuildInfoPath)),
-            source: tsBuildInfoSource
-          });
+          const normalizedTsBuildInfoPath = normalizePath(tsBuildInfoPath);
+          fs.mkdirSync(path.dirname(normalizedTsBuildInfoPath), { recursive: true });
```
NotWoods, line 154: Ideally the plugin should only avoid Rollup's asset system if required. I'd like to see this use `emitFile` if the path falls within the output directory.

- This should use the async versions of the functions and await
- This code should be extracted to a helper function so other parts of the plugin can use it to write outside the dist directory

packages/typescript/test/test.js
```diff
@@ -1057,3 +1097,11 @@ function waitForWatcherEvent(watcher, eventCode) {
     });
   });
 }
+
+function forceRmSync(filePath) {
+  try {
+    fs.unlinkSync(filePath);
```
NotWoods, line 1103: Should use async version

packages/typescript/src/index.ts
```diff
@@ -1,5 +1,7 @@
 import * as path from 'path';
 
+import fs from 'fs';
```
NotWoods, line 3:
```suggestion
import * as fs from 'fs';
```

NotWoods (changes requested):
I've been hesitant to write to the filesystem directly instead of using Rollup's asset system. Doing so makes it impossible to keep Rollup's output in memory. 

I think filesystem writing should be kept behind a flag, such as `outputToFilesystem`:
- By default, allow writing to the file system but show a warning
- If set to false, keep current plugin behaviour (do not allow files to be written if Rollup doesn't allow it)
- If set to true, don't show the warning

bornskilled200:
If you are worried about have files not emitted by rollup, I am confused as to `TsCache` is okay then?
https://github.com/rollup/plugins/blob/master/packages/typescript/src/tscache.ts

Generally I just followed the same style as `TsCache` as the implementation is details are similar.

Just wondering but will be working on the requested changes

packages/typescript/src/index.ts
bornskilled200: I was working on extracting the code into a helper function outside of `index.ts`, but the implementation feels a bit awkward.

I would need access to `outputOptions`, `parsedOptions`, and `this.emitFile`. It might make more sense to have the helper function inside index.js, but I am unsure if this is okay because all the other functions currently are part of `Plugin`

bornskilled200: After looking around the repo a bit more, looks like this is quite normal. Working on it

NotWoods:
That's fair, but I think there is some distinction since the cache files wouldn't need to be used by someone reading compiled files from the Rollup API. I don't remember why the cache is syncronous, I may convert it in the future.

bornskilled200:
still working on the tests

cdalexndr:
Is there a command that I can use to temporary use this fix branch to enable fast dev incremental builds until it is pushed into main and released?
Tried `npm install git+https://github.com/bornskilled200/plugins.git#fix-incremental-typescript-cache -D` but failed with error:
```
3453 error command failed
3454 error command C:\WINDOWS\system32\cmd.exe /d /s /c node scripts/disallow-npm.js
3455 verbose exit 1
```

shellscape:
@cdalexndr that looks like you're trying to add a dependency to this monorepo, which won't work.

vasilev-alex:
hi @bornskilled200! Just to confirm, are you still working on the PR?

shellscape:
@vasilev-alex if you'd like to pick up work on this, please do. @bornskilled200 hasn't checked in on the PR in a month which usually means it's been abandoned.

bornskilled200:
@cdalexndr I used patch-package to make a poc on my personal project. Probably not okay in production

Last time I worked on this, I think I finished the tests and stuff should be okay for a review

I think the way i log the warning messages will probably not pass review

I am also not happy with this line of code, but it works 🤷
https://github.com/rollup/plugins/pull/859/files#diff-120678bfd12e91144ba66253506e2d2e4314e4d287fa75b216b217417cd31d90R91

I will probably not work on this further anytime soon cause of personal life getting in the way

shellscape:
@bornskilled200 thanks for your contributions to this point. I hope things on your end personally are all well.

shellscape:
I'm going to pull this into a local feature branch so we can work on getting it over the finish line.

shellscape:
This PR has been merged into the local branch `rollup:bornskilled200/fix-incremental-typescript-cache` and still requires work. For anyone wishing to pick up work on this, please fork and work from that branch.

PR #866, opened 2021-04-22
Follow up to #843, refining exports and browser field interaction

Jarreddebeer:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] no

List any relevant issue numbers: #843, #786 

### Description

This is a follow up to #843  and #786 based on refining the interaction between the exports and browser field. Specifically of the cases tested in #843. It updates the resolution of the exports field to apply browser internal mappings from the browser field only. **Update:** It appears upstream changes to node-resolve/src/index.js handle the browser internal mappings, so I have removed the files changed relating to that from the PR.

To help aid usability of exports and browser interaction this PR also made it so that setting {browser: true} in the options will also automatically apply the browser exports conditions. This will avoid common pitfalls by having a single configuration for browser mapping.

Tests are adapted from #843 and the output are aligned with webpack output for the web target.

Thanks for your review in advance.


shellscape:
It looks like we have some issues with pnpm and the latest major, and some dependency issues. I'll get those cleaned up soon.

shellscape:
We've just updated `master`. Please update your fork from `upstream/master`.

packages/node-resolve/test/fixtures/node_modules/exports-browser/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "main": "index.cjs",
+  "module": "index.js",
+  "browser": {
+    "./require.js": "./ignored-browser.js"
```
guybedford, line 5: Since there is already a chaining test, how about making this a direct browser override test instead:

```json
{
  "browser": "./ignored-browser.js"
}
```

Where we can then verify that `"exports"` takes precedence over `"browser"` here.

guybedford (commented):
Thanks for the PR @Jarreddebeer. I like the idea of having the `browser: true` option also enable the `browser` condition, so a +1 from me here.

packages/node-resolve/test/fixtures/node_modules/exports-browser/package.json
Jarreddebeer: Thanks, a commit has been added to align the 'exports-browser' test to your suggestions

packages/node-resolve/README.md
```diff
@@ -64,7 +64,7 @@ Setting this option will add extra conditions on top of the default conditions.
 Type: `Boolean`<br>
 Default: `false`
 
-If `true`, instructs the plugin to use the `"browser"` property in `package.json` files to specify alternative files to load for bundling. This is useful when bundling for a browser environment. Alternatively, a value of `'browser'` can be added to the `mainFields` option. If `false`, any `"browser"` properties in package files will be ignored. This option takes precedence over `mainFields`.
+If `true`, instructs the plugin to use the browser module resolutions in `package.json` and adds `'browser'` to `exportConditions` if it is not present so browser conditionals in `exports` are applied. If `false`, any browser properties in package files will be ignored. Alternatively, a value of `'browser'` can be added to the `mainFields` option, however this option takes precedence over `mainFields`.
```
guybedford, line 67: Perhaps note that both adding mainFields and exportConditions is the alternative:

```suggestion
If `true`, instructs the plugin to use the browser module resolutions in `package.json` and adds `'browser'` to `exportConditions` if it is not present so browser conditionals in `exports` are applied. If `false`, any browser properties in package files will be ignored. Alternatively, a value of `'browser'` can be added to both the `mainFields` and `exportConditions` options, however this option takes precedence over `mainFields`.
```

packages/node-resolve/test/browser.js
```diff
@@ -199,3 +199,23 @@ test('pkg.browser with mapping to prevent bundle by specifying a value of false'
...
+  const bundle = await rollup({
+    input: 'browser-exports-browser-browser.js',
+    plugins: [nodeResolve({ browser: true }), commonjs()]
+  });
+  const { module } = await testBundle(t, bundle);
+
+  t.is(module.exports, 'browser');
+});
+
+test('exports.browser does not take precedence over export map result, when browser:false', async (t) => {
```
guybedford, line 213:
```suggestion
test('browser field does not take precedence over export map result', async (t) => {
```

guybedford (approved):
LGTM, I like the unified `browser: true` switch here.

PR #871, opened 2021-05-03
fix(commonjs): idempotence issue

### LINKED ISSUE #869: commonjs: `requireReturnsDefault: "preferred"` yields unstable results
- Rollup Plugin Version: 18.0.0
- Rollup Version: 2.45.2
- Operating System (or Browser): Linux Ubuntu 20, Windows 10 x64
- Node Version: 16.0.0
- Link to reproduction: https://github.com/thorn0/rollup-commonjs-issue

### Expected Behavior

Calling Rollup multiple times with the same settings should lead to the same output.

### Actual Behavior

Only the first time the output is correct and as expected. Subsequent outputs are incorrect and different from the first one.

### Additional Information

```js
for (let i = 0; i <= 3; i++) {
  const result = await rollup({
    input: "input.js",
    plugins: [commonjs({ requireReturnsDefault: "preferred" })],
  });

  await result.write({ file: `output/result${i}.mjs` });
}
```

**First output:**
```js
var numbers = [1, 2, 3, 4, 5, 6];

console.log(numbers);

var input = {

};

export default input; 
```

**Subsequent output:**
```js
var constants = [1, 2, 3, 4, 5, 6];

var constants$1 = /*#__PURE__*/Object.freeze({
	__proto__: null,
	'default': constants
});

console.log(constants$1);

var input = {

};

export default input; 
```

**This works as expected:**
```js
const plugins = [commonjs({ requireReturnsDefault: "preferred" })];

for (let i = 0; i <= 3; i++) {
  const result = await rollup({
    input: "input.js",
    plugins,
  });

  await result.write({ file: `output/result${i}.mjs` });
}
```

thorn0:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: fixes #869

### Description

State was erroneously shared between instances.

### DIFF
--- a/packages/commonjs/src/index.js
+++ b/packages/commonjs/src/index.js
@@ -72,6 +72,7 @@ export default function commonjs(options = {}) {
 
   const esModulesWithDefaultExport = new Set();
   const esModulesWithNamedExports = new Set();
+  const isCjsPromises = new Map();
 
   const ignoreRequire =
     typeof options.ignore === 'function'
@@ -208,7 +209,8 @@ export default function commonjs(options = {}) {
           actualId,
           getRequireReturnsDefault(actualId),
           esModulesWithDefaultExport,
-          esModulesWithNamedExports
+          esModulesWithNamedExports,
+          isCjsPromises
         );
       }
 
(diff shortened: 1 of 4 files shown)


thorn0:
I tried to keep the diff small, but IMHO it'd make sense to refactor this promise map into a class:
https://github.com/thorn0/plugins/commit/9a2a84b8d6218b7b6232a7f3116b86cf3f235368
Should I include that commit to this PR?

guybedford:
Thank you for taking the time to put the minimal PR together here. If you are open to more of a refactoring that could be worthwhile as well here while you are on it - if creating a class for this it might also be worth having that class designed to hold all other cached state to make for an easier clearing / management process as well. Does that seem like it would neaten things up or add more complexity?

thorn0:
I'd rather have this merged as is now as it blocks the release of Prettier.

thorn0:
@guybedford Thanks for the prompt release!

PR #877, opened 2021-05-08
Add ability to disable NodeJS support for Wasm plugin

Ameobea:
## Rollup Plugin Name: `wasm`

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

The Wasm plugin by default emits code that handles loading WebAssembly in a NodeJS environment.  That code contains calls to `require('fs')` and `require('path')`.

When importing libraries that were built using the Wasm plugin from Webpack v5 in projects targeting the browser, Webpack throws build-time errors due to those not being available in the browser.

This change adds an optional flag to the Wasm plugin options which disables that Node.JS support code from being emitted at all.  It prevents the Webpack error, and all browser-targeted Wasm loading support is retained.

### DIFF
--- a/packages/wasm/src/helper.ts
+++ b/packages/wasm/src/helper.ts
@@ -1,6 +1,6 @@
 export const HELPERS_ID = '\0wasmHelpers.js';
 
-export const getHelpersModule = () => `
+export const getHelpersModule = (disableNodeSupport?: boolean) => `
 function _loadWasmModule (sync, filepath, src, imports) {
   function _instantiateOrCompile(source, imports, stream) {
     var instantiateFunc = stream ? WebAssembly.instantiateStreaming : WebAssembly.instantiate;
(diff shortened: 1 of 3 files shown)


shellscape:
Thanks for the PR. While this hasn't gained any traction with the wasm folks that check in from time to time, I'm not keen on adding additional footprint to the plugin's code when there's an alternative that fits your use case: Webpack v4 supports node builtins, and Webpack v5 removed that in core. The happy path for you is to follow the webpack documentation and community examples for bundling code that requires the node polyfills/bulltins https://stackoverflow.com/questions/64557638/how-to-polyfill-node-core-modules-in-webpack-5. Appreciate the effort here, but bundling the plugin output in another bundler for an unexpected environment isn't something that makes a strong case for surface change and added complexity.

filip-dahlberg:
This PR would be of great value. It seems like I need to switch from rollup to webpack for bundling the wasm and run it in browser env. Since rollup does not support bundling wasm for browser.

PR #887, opened 2021-05-22
Add plugins table's columns headers

quantumsheep:
## Rollup Plugins

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ This repository houses plugins that Rollup considers critical to every day use o
 
 ## Plugins Found Here
 
-|                                                     |                                                                                           |
+| Name                                                | Description                                                                               |
 | --------------------------------------------------- | ----------------------------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


shellscape:
Thanks for the PR, but I don't think we really need columns here - it's pretty evident what each of the two are, especially given the preceding heading. I'll wait for other maintainers to weigh in before closing.

quantumsheep:
> Thanks for the PR, but I don't think we really need columns here - it's pretty evident what each of the two are, especially given the preceding heading. I'll wait for other maintainers to weigh in before closing.

I think that even though we think those columns are evident, headings could improve readability.

shellscape:
Closing as we're at 60 days+ without additional comment. Thanks for the suggestion and the PR.

PR #890, opened 2021-05-31
Add resolve options to typescript plugin

chiefidiot:

This PR contains:

- [x] bugfix
- [x] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

#287 

### Description
When targeting files outside of the current working directory of the process, the plugin fails because if filters out all files not inside `process.cwd()`.  This adds the default value to look at the resolved TS config `rootDir` option while adding an additional option to control it `resolveRoot` which is passed to `resolve` in the `createFilter`.  This fixes (at least my issue) seeing `Error: Unexpected token (Note that you need plugins to import files that are not JavaScript)` in a monorepo.

### DIFF
--- a/packages/typescript/README.md
+++ b/packages/typescript/README.md
@@ -79,6 +79,15 @@ Default: `null`
 
 A [minimatch pattern](https://github.com/isaacs/minimatch), or array of patterns, which specifies the files in the build the plugin should operate on. By default all `.ts` and `.tsx` files are targeted.
 
+### `resolveRoot`
+
+Type: `String` | `Boolean`<br>
+Default: `true`
+
+Sets the `resolve` parameter for [createFilter](https://github.com/rollup/plugins/tree/master/packages/pluginutils#createfilter). By default, it will resolve against the `rootDir` set in the TS config file however you can set this to change which files are filtered out by the plugin.
+
+This can fix `Error: Unexpected token (Note that you need plugins to import files that are not JavaScript)` TypeScript error when targeting files outside the current working directory (`process.cwd()`).
+
 ### `tsconfig`
 
 Type: `String` | `Boolean`<br>
--- a/packages/typescript/src/index.ts
+++ b/packages/typescript/src/index.ts
@@ -1,5 +1,7 @@
 import * as path from 'path';
 
+import { createFilter } from '@rollup/pluginutils';
+
 import { Plugin, RollupOptions, SourceDescription } from 'rollup';
 import type { Watch } from 'typescript';
 
@@ -19,10 +21,12 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
   const {
     cacheDir,
     compilerOptions,
-    filter,
     transformers,
     tsconfig,
     tslib,
+    include,
+    exclude,
+    resolveRoot,
     typescript: ts
   } = getPluginOptions(options);
   const tsCache = new TSCache(cacheDir);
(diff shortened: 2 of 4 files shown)


packages/typescript/src/index.ts
```diff
@@ -19,17 +21,22 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
...
+    filterRoot,
     typescript: ts
   } = getPluginOptions(options);
   const tsCache = new TSCache(cacheDir);
   const emittedFiles = new Map<string, string>();
   const watchProgramHelper = new WatchProgramHelper();
 
   const parsedOptions = parseTypescriptConfig(ts, tsconfig, compilerOptions);
+  const filter = createFilter(include || ['*.ts+(|x)', '**/*.ts+(|x)'], exclude, {
+    resolve: filterRoot ?? compilerOptions.rootDir
```
NotWoods, line 38: Can this be handled inside `getPluginOptions`? It looks like the needed variables (`include`, `exclude`, `filterRoot`, `compilerOptions`) are available in that scope too.

packages/typescript/README.md
```diff
@@ -79,6 +79,15 @@ Default: `null`
 
 A [minimatch pattern](https://github.com/isaacs/minimatch), or array of patterns, which specifies the files in the build the plugin should operate on. By default all `.ts` and `.tsx` files are targeted.
 
+### `filterRoot`
+
+Type: `String` | `Boolean`<br>
+Default: `true`
```
NotWoods, line 85: I think "`rootDir` in `tsconfig.json`" or "`process.cwd`" makes more sense as the default value written in the docs.

packages/typescript/README.md
```diff
@@ -79,6 +79,15 @@ Default: `null`
...
+
+Sets the `resolve` parameter for [createFilter](https://github.com/rollup/plugins/tree/master/packages/pluginutils#createfilter). By default, it will resolve against the `rootDir` set in the TS config file however you can set this to change which files are filtered out by the plugin.
```
NotWoods, line 87: These docs are a little difficult to read if you're not familiar with Rollup's internals. I'd prefer something based on `createFilter`'s docs, talking about how this option sets the base directory for `include` and `exclude`.

i.e.:

>Optionally resolves the `include` and `exclude` patterns against a directory other than `process.cwd()`. If a `String` is specified, then the value will be used as the base directory. Relative paths will be resolved against `process.cwd()` first. If `false`, then the patterns will not be resolved against any directory. 
>
> By default, patterns resolve agaisnt the `rootDir` set in your TS config file.
>
>This can fix plugin errors when parsing files outside the current working directory (`process.cwd()`)

NotWoods (changes requested):
Thanks for finding a good system around the issue. I like the general approach but I feel that the documentation is difficult to understand if you don't already know about the issue.

packages/typescript/src/index.ts
chiefidiot: Forgot to change to `parsedOptions` in the oss repo, which isn't available in `getPluginOptions`. It could be refactored to pull the parse config logic into `getPluginOptions`, but that function is only used once with no defined return interface and seems to make sense to keep that single purpose.  Trying to keep changes minimal, it seems appropriate to keep separate the options normalization and the utility instantiations.

packages/typescript/README.md
chiefidiot: Fixed

packages/typescript/README.md
chiefidiot: Updated

NotWoods (changes requested):
Can you add some tests for the new feature? Looks good otherwise :)

shellscape:
@chiefidiot ping re: tests

chiefidiot:
Tests added

NotWoods (approved):

janpe:
@chiefidiot would you be able to check the merge conflicts? I am probably facing the same issue. Is there another workaround for this for the time being?

janpe:
This can be closed now, right?

shellscape:
Yes, superceded by #1050

PR #896, opened 2021-06-03
fix(alias): paths for aliases across multiple Windows drives

Robinfr:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

We ran into an issue where we had aliases to paths on another Windows drive which was not working. After some investigation, I found out that this is because the drive letter is being removed in the alias plugin. Removing that piece of code made the alias work properly across multiple Windows drives.

Questions:
- I also removed usage of the `slash` library as it's not actually needed to work properly. Do you want me to remove the library entirely?
- The normalize functions are doing nothing more than just returning the `id` and `path` straight away, so they could technically be removed entirely. Do you want me to take care of that?

### DIFF
--- a/packages/alias/src/index.ts
+++ b/packages/alias/src/index.ts
@@ -1,12 +1,6 @@
-import { platform } from 'os';
-
 import { PartialResolvedId, Plugin } from 'rollup';
-import slash from 'slash';
 
 import { Alias, ResolverFunction, RollupAliasOptions } from '../types';
-
-const VOLUME = /^([A-Z]:)/i;
-const IS_WINDOWS = platform() === 'win32';
 
 const noop = () => null;
 
@@ -28,9 +22,6 @@ function matches(pattern: string | RegExp, importee: string) {
 function normalizeId(id: string): string;
 function normalizeId(id: string | undefined): string | undefined;
 function normalizeId(id: string | undefined) {
-  if (typeof id === 'string' && (IS_WINDOWS || VOLUME.test(id))) {
-    return slash(id.replace(VOLUME, ''));
-  }
   return id;
 }
 
--- a/packages/alias/test/test.js
+++ b/packages/alias/test/test.js
@@ -2,14 +2,13 @@ import path, { posix } from 'path';
 
 import test from 'ava';
 import { rollup } from 'rollup';
-import slash from 'slash';
 
 // eslint-disable-next-line import/no-unresolved, import/extensions
 import nodeResolvePlugin from '@rollup/plugin-node-resolve';
 
 import alias from '../dist';
 
-const normalizePath = (pathToNormalize) => slash(pathToNormalize.replace(/^([A-Z]:)/, ''));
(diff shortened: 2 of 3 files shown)


Robinfr:
Can I poke someone for this? We're already using this fix ourselves using `patch-package` but would be nice to get this also i the official repo.

shellscape:
Could you share some more information about the issue you ran into? As much detail as possible would be helpful. This is a rather impactful change so we need to fully understand what was happening.

Robinfr:
Sure, here's the situation:

We have some common dependencies installed on drive `C:/`, but the project where Rollup is being used in on drive `D:/`. We are using the `alias` plugin to alias some of these common dependencies, e.g.:

`"react" -> "C:/common/node_modules/react"`

However, when it tries to resolve the import in a file on the `D:/` project that looks like this:

```jsx
import React from "react";
```

The resolve fails because the id is incorrect as it removes the volume from the path, e.g. it becomes `/common/node_modules/react` which is incorrect. Rollup however does not remove the volume when generating the bundle, so there is a mismatch of ids when trying to resolve the id (I believe).

Removing the normalization as it is currently implemented fixes this issue.

shellscape:
putting dependencies in a separate drive is rather bizarre in the node realm. I'm not sure what the ramifications will be for not normalizing things, given that specific work was done around making that happen. for your use case, why not use the Windows equivalent of a symlink?

Robinfr:
I understand it's rather bizarre, but this is unfortunately a requirement we cannot work around as our product is shipped with specific node modules that can be used from a different drive. 

The code that is here seems very specific to normalizing for Windows drives, yet it does not seem to work properly since it does not work across drives, and most likely is not needed when it's on the same drive. The test for it is unfortunately not really doing much since it's simply testing against the same algorithm that's being used for normalizing (I understand it's not actually possible to test for this scenario using normal CI).

What was the specific issue that was trying to be solved by adding this piece of code?

shellscape:
> What was the specific issue that was trying to be solved by adding this piece of code?

Screen Shot 2021-06-08 at 9 34 25 AM

"View git blame"

shellscape:
Added some folks as reviewers to review the conversation thus far. If this receives no further comment for 30 days, we'll unfortunately have to decline the PR.

danielgindi:
Viewing git blame, and issues history, I still don't see a justification for the original change that caused the currently discussed bug.
Furthermore, the fact that it's currently deviating from how the whole rollup ecosystem describes and formats `id`s- is worrying to me.

If anyone sees an actual issue that wS resolved by the original change, then this issue should be clearly described and a test for it should be added.

Otherwise, I vote for merging this PR.

shellscape:
> Otherwise, I vote for merging this PR.

That's all the justification I think we need. Thanks for taking a look @danielgindi.

shellscape (approved):

PR #897, opened 2021-06-03
fix(inject): escape metacharacters in module name string

lightmare:
## Rollup Plugin Name: `inject`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

I don't know how to answer this one. Technically, it might be considered a breaking change. If someone relies on backslashes in module names not being escaped, or on a module with apostrophes + JS code in its name as some sort of JS-injection, then this PR would ruin their day.

n/a

### Description

The inject plugin doesn't escape metacharacters in module name that it pastes as a string. Therefore if you try to inject a module named `d'oh`, you'll end up with `import * from 'd'oh';` — not valid JavaScript.

I have a question about CI. At first I only pushed the first commit with a test, expecting it to fail. But it didn't, it succeeded with [no tests executed](https://github.com/lightmare/plugins/runs/2737328447?check_suite_focus=true#step:10:4). Same with the subsequent commit containing the fix, it didn't execute [any tests](https://github.com/lightmare/plugins/runs/2737400617?check_suite_focus=true#step:10:4).

Is it supposed to work like this? When I run that command locally it does execute tests for `plugin-inject`.


shellscape:
Looks good to me. I'm just super curious about the use case, as I haven't seen an `import` target that had those kinds of characters in them.

lightmare:
Well it's a silly long story, but since you asked... :newspaper_roll: 

TL;DR :next_track_button: found a bug in Babel, fixed it and then tried to break it more.

I purposely set my `$HOME` to have some non-ASCII characters. I do this to avoid repeating mistakes such as assuming `$HOME` is the same as `/home/$USER`, or assuming that a filename argument is always ASCII.

One day I tried to build Babel, which failed because it was emitting absolute paths to modules, but URL-encoded everything non-ASCII. That was fixed in Babel 7.14.5.

Then I had a bunch of failing tests, because those too were assuming ASCII paths, but here it was more complicated. They need to compare actual output containing paths against expected output containing `<CWD>` placeholders; so they normalize the actual output before the comparison, and I noticed this normalization involved un-escaping backslashes, but only on Windows. Which to me didn't make any sense. And I'm not talking about them still using backslashes in paths; that is baffling but irrelevant. If you want to un-escape paths inside JS string literals, it doesn't matter whether you're on Windows.

Naturally, I cloned Babel into directories with various silly characters in the name to see what happens. For some of them Babel didn't even fail gracefully, just spat out nonsense :boom: After finding out it swallowed the original error from ill-formed module name string, I landed here.

shellscape:
That is a wild journey! Thanks for sharing that though.

We did a lot of maintenance on the repo today. If you have a moment, please run the following commands to update your fork, and we'll be happy to merge:

```
git remote add upstream git@github.com:rollup/plugins.git
git fetch upstream
git merge upstream/master
git push
```

shellscape (approved):

shellscape:
@lightmare looks like you've got some failing tests on the Windows tests. please have a look at those.

lightmare:
Maybe I should remove the failing test with backslash in module name — it's too silly, and non-portable.

shellscape:
@lightmare checking in on this

lightmare:
I've disabled the "backslash-in-filename" test on Windows, where that's impossible.

PR #898, opened 2021-06-03
feat(dsv,dynamic-import-vars,image,legacy,multi-entry,strip,sucrase,url,yaml): add missing typings

Luke-zhang-04:
## Rollup Plugin Name: dsv, dynamic-import-vars, image, legacy, multi-entry, strip, sucrase, url, yaml

This PR contains:

- [ ] bugfix
- [x] "feature"
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

Adds typings for plugins which were missing them, including dsv, dynamic-import-vars, image, legacy, multi-entry, strip, sucrase, url, and yaml.


marcopixel (changes requested):
I would recommend making the options parameter optional because this will result in an TS error if no own configuration or an empty object is set inside the plugin function.

```
--- ERROR
plugins: [url()]

--- PASS
plugins: [url({})]
```

Since a few of the official plugins are using no option parameters in their examples
([url](https://github.com/rollup/plugins/tree/master/packages/url#usage) or [multi](https://github.com/rollup/plugins/tree/master/packages/url#usage) as example) & having unnecessary empty objects just to "fix" typing errors, i think it's better to let the user decide if they want to declare additional options.

Luke-zhang-04:
Of course, why didn't I think of that

shellscape:
Thanks for putting in this work. While I haven't had a chance to look it over (yet. I will) I did want to pop in and mention that this will force me to finally get around to improving our publish-release pipeline so it can handle more than one plugin per Pull Request. It'll probably be a week or so, so don't be discouraged if your PR sits for a bit.

packages/dsv/package.json
```diff
@@ -41,6 +41,7 @@
     "del-cli": "^3.0.1",
     "rollup": "^2.23.0"
   },
+  "types": "./types/index.d.ts",
```
NotWoods, line 44: The typings file needs to be added to the `"files"` array in package.json

NotWoods, line 44: [Tests](https://github.com/rollup/plugins/blob/master/packages/buble/test/types.ts) should be added for all the new types files, then added to the package.json scripts as [`"test:ts"`](https://github.com/rollup/plugins/blob/master/packages/buble/package.json#L32)

NotWoods (changes requested):

Luke-zhang-04:
Hm, didn't think you could test types, but I'll look into it.

packages/dsv/package.json
```diff
@@ -25,22 +25,34 @@
...
+    "test:ts": "tsc ./test/types.ts ./types/index.d.ts --noEmit"
   },
   "files": [
     "dist",
+    "types",
     "README.md",
     "LICENSE"
   ],
+  "peerDependencies": {
+    "@types/d3-dsv": "~1.2.0"
```
NotWoods, line 38: This should just be a hard `dependencies` instead of peer.

packages/dynamic-import-vars/test/types.ts
```diff
@@ -0,0 +1,20 @@
+import { RollupOptions } from 'rollup';
+
+import dynamicImportVars from '../types';
```
NotWoods, line 3:
```suggestion
import dynamicImportVars from '../';
```

NotWoods (changes requested):

packages/dsv/package.json
Luke-zhang-04: Hm. Types were installed as peer dependencies in the [Babel Plugin](https://github.com/rollup/plugins/blob/master/packages/babel/package.json#L49).

NotWoods: babel/core is a peer dependency in that plugin but d3-dsv is a built in dependency here. However, we should probably update that plugin too anyways.

Luke-zhang-04: :+1:

Luke-zhang-04:
can someone help me out with the linting process? When I run `pnpm run lint` it formats files that I haven't touched. Do I just leave these?

image

NotWoods (approved):
You should be able to leave out the files that only have lint changes. I'll double check and see if CI passes without it.

packages/buble/test/types.ts
```diff
@@ -1,6 +1,6 @@
 import { RollupOptions } from 'rollup';
 
-import buble from '..';
+import buble from '../types';
```
NotWoods, line 3:
```suggestion
import buble from '..';
```
Please switch all of these imports to use the root directory instead of the types directory.

Luke-zhang-04: Oh I read the diff wrong 🤦

wdfinch:
This is awesome!

shellscape:
@Luke-zhang-04 the repo is now prepared for automatic publishing, and handling more than one plugin per commit. however, it required some refactoring, which has caused conflicts with your branch. could you please pull from upstream/master and resolve the conflicts in your branch?

Issue #914, opened 2021-06-28
Specifying entryFileName option in @rollup/plugin-multi-entry does not output with given value 


Dechowmedia:
- Rollup Plugin Name: @rollup/plugin-multi-entry
- Rollup Plugin Version: 4.0.0
- Rollup Version: 2.52.3
- Operating System (or Browser): NodeJS
- Node Version: 14.16.1
- Link to reproduction _(⚠️ read below)_: https://replit.com/@Dechowmedia/rollup-multi-entry-bug#rollup.config.js

### Expected Behavior
option `entryFileName` should render an output (in the output folder) with `custom-file-entry.js`.

### Actual Behavior
renders default value of `multi-entry.js`

### Additional Information
PR with fix but needs additional testing (perhaps) #915

PR #915, opened 2021-06-28
Overwrite else statement in configure

Dechowmedia:
To accept config modifications earlier - as it defaults to DEFAULT_OUTPUT in all cases forward due to the else overwrite.
There are better ways to do this (such as utilising spread e.g. `{...input, ...config}`) but I am in doubt whether the fix should be applied in such a way - maybe that's preferred - reading the code-style/syntax used previously.
That could fix the unneeded _nullish coalescing operator_ use (which may or may not be available depending on your node version).

This still needs test written as well - I am in the process of figuring out a good way to do that.

## Rollup Plugin Name: `@rollup/multi-entry`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no (have to make these)

Breaking Changes?

- [x] no

#914

### DIFF
--- a/packages/multi-entry/src/index.js
+++ b/packages/multi-entry/src/index.js
@@ -25,7 +25,7 @@ export default function multiEntry(conf = {}) {
     } else if (Array.isArray(input)) {
       config.include = input;
     } else {
-      const { include = [], exclude = [], entryFileName = DEFAULT_OUTPUT, exports } = input;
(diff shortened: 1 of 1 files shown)


shellscape:
Since this hasn't seen any activity in 30 days, we're going to close this Draft PR. The intent here isn't quite clear, and without activity there's not much reason to keep an open Draft PR on this repo. Please feel free to continue to work on this on your fork and PR when you feel it's ready.

Issue #919, opened 2021-07-01
[node-resolve] package.json "browser" paths are incorrectly resolved for sub-paths


eemeli:
- Rollup Plugin Name: @rollup/node-resolve
- Rollup Plugin Version: 13.0.0
- Rollup Version: 2.52.6
- Operating System (or Browser): MacOS
- Node Version: 15.11.0
- Link to reproduction: https://replit.com/@eemeli1/rollup-plugin-repro

### Expected Behavior

Build should succeed.

### Actual Behavior

Build fails with a circular dependency:
```
node_modules/yaml/browser/index.js -> node_modules/yaml/browser/index.js
```

This is due to the browser import path for `"./dist/index.js"` being incorrectly applied when used in the `browser/` sub-path of the `yaml` package.

### Additional Information

Originally reported by @jdesrosiers as eemeli/yaml#291, see there for some initial debugging.

I'll work on a PR potentially fixing this.

guybedford:
Fixed in https://github.com/rollup/plugins/pull/920.

PR #920, opened 2021-07-01
fix(node-resolve): handle browser-mapped paths correctly in nested contexts

eemeli:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: #919, eemeli/yaml#291

### Description

For each entry in a package.json `"browser"` object entry, the `browserMapCache` is filled with mapped paths at keys using both the raw source of the path that needs to be remapped, as well as the resolved path. This effectively enables both package name imports (like `"foo"`) as well as relative path imports (like `"./foo"`) to work when `doResolveId` is called. However, this causes a problem when the raw form of a relative import in a sub-path matches a value that's included in the package.json `"browser"` entry.

In other words, if `"browser"` includes `"./foo": "./other.js"`, an import of `"./foo"` in a file `./bar/baz.js` is mapped to `./other.js`, even though it shouldn't; [the spec](https://github.com/defunctzombie/package-browser-field-spec) is pretty clear on this:

> All paths for browser fields are relative to the `package.json` file location (and usually project root as a result).

To fix, only the resolved paths are checked when applying mappings for relative imports.

### DIFF
--- a/packages/node-resolve/src/index.js
+++ b/packages/node-resolve/src/index.js
@@ -88,7 +88,7 @@ export function nodeResolve(opts = {}) {
         return { id: ES6_BROWSER_EMPTY };
       }
       const browserImportee =
-        browser[importee] ||
+        (importee[0] !== '.' && browser[importee]) ||
         browser[resolvedImportee] ||
         browser[`${resolvedImportee}.js`] ||
         browser[`${resolvedImportee}.json`];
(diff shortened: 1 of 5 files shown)


guybedford:
Thanks you for this fix and apologies for the delayed review! This looks great to me.

guybedford (approved):

PR #921, opened 2021-07-01
fix(node-resolve): Fix bug where JS import was converted to a TS import, resulting in an error when using export maps

### LINKED ISSUE #853: [node-resolve] `.js` files imported from a`.ts` file gets resolved to `.ts` extension
Facing an issue with `@web/dev-server` at https://github.com/modernweb-dev/web/issues/1376 where the node resolution breaks for dependencies that are using `node`'s `exports` map, e.g.

```ts
// main.ts
import { highlight } from 'nodemod/dist/lib/prismjs.js'; // nodemod@2.7.3 is using `exports` map in node
```

FYI, nodemod's [package.json](https://unpkg.com/browse/nodemod@2.7.3/package.json) is as follows:

```json
// nodemod's package.json
...
"exports": {
  "./dist/*": "./dist/*",
  ...
}
```

What's happening in `@web/dev-server` is that files are imported with `.ts` extension due to the following code as pointed out by the author of `@web/dev-server`:

https://github.com/rollup/plugins/blob/53776ee3fc59587fe5ff0c756216185f48372507/packages/node-resolve/src/index.js#L178-L184

## Expected behavior

`import { highlight } from 'nodemod/dist/lib/prismjs.js';`  should resolve to `'node_modules/nodemod/dist/lib/prismjs.js';`

## Actual behavior

`import { highlight } from 'nodemod/dist/lib/prismjs.js';`  resolves to `'node_modules/nodemod/dist/lib/prismjs.ts';`

### LINKED ISSUE #911: Cannot bundle a TS file, which imports JS file from a package using NodeJS conditional exports
Hi,

- Rollup Plugin Name: `@rollup/plugin-node-resolve`
- Rollup Plugin Version: `13.0.0`
- Rollup Version: `2.52.2`
- Operating System (or Browser): `macOS Big Sur 11.4`
- Node Version: `14.15.1`
- Link to reproduction _(⚠️ read below)_: https://github.com/bhovhannes/rollup-plugin-node-resolve-issue

#### Steps to reproduce

1. `git clone` the https://github.com/bhovhannes/rollup-plugin-node-resolve-issue repo
1. run `npm ci`
1. run `npm run build`

#### Details

The reproduction link written above is a small Rollup project. Entrypoint - `index.ts`, is in TypeScript. TypeScript does not really matter here, the only thing which matters is the extension of the file - `.ts`.

Inside an entrypoint file we import a utility function using `import sum from '@abcdef/lib-utils/dist/sum.js'` syntax.

The package `@abcdef/lib-utils` has an `exports` field in its `package.json` containing:
```
"exports": {
    "./dist/*": {
        "import": "./dist_esm/*",
        "require": "./dist/*"
    }
}
``` 
That makes possible to let Rollup consume `sum.js` from `dist_esm` folder and if the tool does not support ESM (looking on you Jest) it will use `sum.js` from `dist` folder.

Lines https://github.com/rollup/plugins/blame/master/packages/node-resolve/src/index.js#L148-L155 in `@rollup/plugin-node-resolve` code check for the case when there is a `.js` import inside a `.ts` file and change extension to `.ts`.  I'll be grateful for any explanation why this is needed. 
After that, plugin resolves the package, taking `exports` field into account and ends up with `node_modules/@abcdef/lib-utils/dist_esm/sum.ts` url, which obviously does not exist.

### Expected Behavior

Build should complete without errors.

### Actual Behavior

Build errors out, trying to bundle non-existent TypeScript file. The error:
```
Error: Could not load /Users/hovhannesbabayan/Projects/github/bhovhannes/
rollup-plugin-node-resolve-issue/node_modules/@abcdef/lib-utils/dist_esm/sum.ts (imported by input.ts): 

ENOENT: no such file or directory, open '/Users/hovhannesbabayan/Projects/github/bhovhannes/
rollup-plugin-node-resolve-issue/node_modules/@abcdef/lib-utils/dist_esm/sum.ts'
```

### Additional Information

Removing lines https://github.com/rollup/plugins/blame/master/packages/node-resolve/src/index.js#L148-L155 fixes the issue.
  
Or, perhaps we can check if the resolved file indeed exists in https://github.com/rollup/plugins/blob/master/packages/node-resolve/src/package/resolvePackageTarget.js#L13 ?

I'll be happy to submit a pull request, just I am not sure in which direction to proceed.

Thanks!

tjenkinson:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: fixes #911, fixes #853

### Description
Currently if when using export maps we resolve to a file that does not exist we don't try the remaining options. This fixes that.

We now always make the import the user wrote the first candidate, and we also only try the exports algorithm on the first candidate, and fallback to the standard node resolve algorithm for all the candidates only if the first one determines that exports are not configured.

It's specifically a problem when importing a `.js` file from a `.ts` file when you have `.ts` extensions configured, because for reasons I don't fully understand we first rewrite `*.js` imports to `*.ts` imports. [Typescript itself added a 'feature' to do this](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#module-identifiers-allow-for-js-extension), so we followed them.

_I think it's somehow related to js modules requiring ids to end in '*.js', but not sure why typescript decided from that to allow .js and convert .ts_


guybedford:
Just to clarify I'm understanding the logic here, this PR will:

* If the file pointed to by "exports" does not exist, it "falls back" to applying the normal resolver?
* As a result pkg/x.js can resolve pkg/x.ts?
* But this fallback itself does not apply the exports mappings?

Would help to get a clear sense of it.

guybedford:
> By returning null from resolveId though it means the for loop in resolveImportSpecifiers will continue to run and try the other candidates where previously it would not.

This is the thing that's the issue here - "exports" is designed to be an authoritative resolution system. If the file is not found it should not resolve, _only unless it is a "require" resolution in which case CommonJS extension searching can apply_.

For ES module resolutions exports should include the final file extension always.

> Before this PR this is already the case for the normal resolver, and think it should also be the case for when the export map resolver is used?

The issue is what if the exports mapping is like `"exports": { "./*": "./dist/*.js" }` - we then fallback on the directory indirection which is the problem - where the not found should be correct.

guybedford:
> In which case maybe we should scrap the *.js => *.ts logic when using exports then?

Agreed this should not be happening.

packages/node-resolve/test/package-entry-points.js
```diff
@@ -352,7 +352,8 @@ test('can override a star pattern using null', async (t) => {
 
 test('can self-import a package when using exports field', async (t) => {
   const bundle = await rollup({
-    input: 'self-package-import',
+    // TODO previously this worked because we'd do this internally
+    input: './self-package-import',
```
tjenkinson, line 355: @guybedford this is needed now because the export map algo only runs on the import the user provided. So is this ok? If it is ok, does this make it a breaking change now? :/

Previously we 'fixed' `self-package-import` to `./self-package-import` and attempted that first

packages/node-resolve/src/index.js
```diff
@@ -121,30 +121,17 @@ export function nodeResolve(opts = {}) {
...
-    const importeeIsBuiltin = builtins.has(importee);
-
-    if (importeeIsBuiltin) {
-      // The `resolve` library will not resolve packages with the same
-      // name as a node built-in module. If we're resolving something
-      // that's a builtin, and we don't prefer to find built-ins, we
-      // first try to look up a local module with that name. If we don't
-      // find anything, we resolve the builtin which just returns back
-      // the built-in's name.
-      importSpecifierList.push(`${importee}/`);
```
tjenkinson, line 145: all tests pass without this now, so I think maybe this is no longer needed since the other work I did to add builtin support to `resolve`

packages/node-resolve/src/package/resolvePackageImports.js
```diff
@@ -33,7 +33,7 @@ async function resolvePackageImports({
   }
 
   if (importSpecifier === '#' || importSpecifier.startsWith('#/')) {
-    throw new InvalidModuleSpecifierError(context, 'Invalid import specifier.');
+    throw new InvalidModuleSpecifierError(context, true, 'Invalid import specifier.');
```
tjenkinson, line 36: @guybedford should this be `true` (`internal`) here?

packages/node-resolve/test/package-entry-points.js
tjenkinson: If we have
```js
input: 'self-package-import2'
```
and rename `self-package-import.js` to `self-package-import2.js` it works fine.

Or if we just put `self-package-import.js` it also works, so going to switch it to that.

I think it fails now because the first run of the resolve algorithm is treating `self-package-import` as matching the package with that name over `./self-package-import.js`, which I think is correct?

abdonrd:
Thanks for working on this! I look forward to it!

packages/node-resolve/src/package/resolvePackageImports.js
guybedford: Yes, good catch!

packages/node-resolve/test/package-entry-points.js
```diff
@@ -352,7 +352,7 @@ test('can override a star pattern using null', async (t) => {
 
 test('can self-import a package when using exports field', async (t) => {
   const bundle = await rollup({
-    input: 'self-package-import',
+    input: 'self-package-import.js',
```
guybedford, line 355: This is a tricky one from an extensions perspective, because it's unclear if we are building a package or a path so there is some ambiguity in place.

Do we try the package lookup first, then fall back to a path-like lookup? Even still, a path-like lookup should still support extension checking I think.

packages/node-resolve/test/test.js
```diff
@@ -124,6 +141,25 @@ test('supports JS extensions in TS when referring to TS imports', async (t) => {
   t.is(module.exports, 'It works!');
 });
 
+test('supports JS extensions in TS actually importing JS with export map', async (t) => {
```
guybedford, line 144: What about the following cases:

* `.js` file importing `.ts` file?
* `.ts` file importing `.ts` file?
* `.ts` file importing another `.ts` file using the `.js` extension?

guybedford (commented):
Thank you so much @tjenkinson for your work on this PR, it is really valued. Sorry I dropped the ball on this for a while will aim to be more responsive.

I've shared some more questions mostly, if you can perhaps let me know where things stand on those hopefully we can resolve them fairly quickly, also deciding which are in and out of scope for this PR as well.

packages/node-resolve/test/package-entry-points.js
tjenkinson: With `self-package-import` we get
```
Could not resolve import "self-package-import" in undefined using exports defined in /Users/tomjenkinson/Documents/GitHub/plugins/packages/node-resolve/test/fixtures/node_modules/self-package-import/package.json.
```
because it does try treating it as the package, and does not fall back then to `self-package-import.js` given the package had an export map.

Are you saying we should fall back to trying the candidates with extensions if it's the root import (no importer)?

guybedford: I see, so because it is matching the package lookup path, we don't try the path-like resolution.

That makes sense to me then.

tjenkinson: IMO we shouldn't have a special case for the root import, unless there is a good use case for it?

The probability of users having a file named the same as a package _and_ missing off the extension is probably quite low?

packages/node-resolve/test/test.js
tjenkinson: I think these should all be fine because they will all be imports with the extension included?

I think
>.ts file importing another .ts file using the .js extension?

is covered with "supports JS extensions in TS when referring to TS imports"?

guybedford: There is an ambiguity with the `.ts` file loading a `.js` file case - because it could be a `.js` or a `.ts`, and then the question is which to give precedence. Is there an existing test covering this case? If not it would be good to add one to ensure this is no regression here.

tjenkinson: Sure happy to add more tests. Still not following exactly what you mean though. Do you mean a .ts file importing `file.js`, when there is a `file.js` and a `file.ts`, and checking the `file.js` is actually the one that's loaded?

guybedford: Yes that's the case, where the `file.ts` is loaded when the `file.js` does not exist.

tjenkinson: There's
- "supports JS extensions in TS actually importing JS with export map"
- "supports JS extensions in TS when referring to TS imports"

I also just updated the test I added to have an import with the same name to ensure it prefers the one in the import and doesn't always switch to ".ts"

guybedford: Great, thanks for confirming this.

guybedford (approved):

PR #924, opened 2021-07-03
fix(commonjs)!: Correctly infer module name for any separator

### LINKED ISSUE #923: commonjs: Module ID containing forward slash is improperly normalized on Windows
- Rollup version:
- OS: Windows 10 Enterprise 20H2
- Node.js version: 12.22.1 (x64)

When extracting the base name of a module from its module ID, @rollup/plugin-commonjs does not correctly normalize paths with separators that are not `path.sep`. This happens with some custom resolvers (e.g. Vite.js) which generate module IDs with forward slashes on Windows.

For example, if the module ID is `C:/Users/Phil/Documents/GitHub/test-vite-app/node_modules/react/index.js`, the plugin generates code like this on Windows:

```js
var C__Users_Phil_Documents_GitHub_testViteApp_node_modules_react = { exports: {} };
var react_production_min = {};
```

This is undesireable because (1) it leaks the absolute path of the development machine into the bundle, and (2) it generates bundles with different hashes on different machines, hurting reproducible builds.

This is caused by `getName()`, which is used to extract the base name of a module:
https://github.com/rollup/plugins/blob/6867290ec8b983950ce20e58a34eaaee5e371517/packages/commonjs/src/utils.js#L25-L32

Here, it uses `path.sep` to split the module ID. Since `sep` is `\` on Windows, it fails to split paths that use `/`.

---

Since `basename()` and `dirname()` can handle both forward and backward slashes, I suggest that we replace lines 30-31 with:
```js
  return makeLegalIdentifier(basename(dirname(id))); 
```

See related investigation at https://github.com/vitejs/vite/issues/2623#issuecomment-873352054

pastelmind:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] no

List any relevant issue numbers: #923

### Description

When a module whose file name is `index.js` is resolved to a module ID containing a directory separator that is different from [`path.sep`](https://nodejs.org/docs/latest/api/path.html#path_path_sep) of the current platform, the commonjs plugin incorrectly infers the base name of the module. This causes the absolute path of the module to be embedded into the bundle, and prevents generating reproducible builds across different machines.

For example, with a module ID of `C:/Users/Phil/Documents/GitHub/test-vite-app/node_modules/react/index.js` on Windows, the plugin generates code like this:
```js
var C__Users_Phil_Documents_GitHub_testViteApp_node_modules_react = { exports: {} };
var react_production_min = {};
```

This PR fixes the problem by using `path.basename()` to extract the final path component instead of splitting the string with `path.sep`.

With the fix, the plugin generates code like this:
```js
var react = { exports: {} };
var react_production_min = {};
```

Closes #923

### DIFF
--- a/packages/commonjs/src/utils.js
+++ b/packages/commonjs/src/utils.js
@@ -1,6 +1,6 @@
 /* eslint-disable import/prefer-default-export */
 
-import { basename, dirname, extname, sep } from 'path';
+import { basename, dirname, extname } from 'path';
 
 import { makeLegalIdentifier } from '@rollup/pluginutils';
 
@@ -27,8 +27,7 @@ export function getName(id) {
   if (name !== 'index') {
     return name;
   }
-  const segments = dirname(id).split(sep);
-  return makeLegalIdentifier(segments[segments.length - 1]);
+  return makeLegalIdentifier(basename(dirname(id)));
 }
 
 export function normalizePathSlashes(path) {
--- a/packages/commonjs/test/fixtures/samples/module-path-separator/foo/index.js
+++ b/packages/commonjs/test/fixtures/samples/module-path-separator/foo/index.js
@@ -1,0 +1,1 @@
+module.exports.a = 1;
--- a/packages/commonjs/test/fixtures/samples/module-path-separator/main.js
+++ b/packages/commonjs/test/fixtures/samples/module-path-separator/main.js
@@ -1,0 +1,2 @@
+const foo = require("./foo");
+console.log(foo.a);
--- a/packages/commonjs/test/test.js
+++ b/packages/commonjs/test/test.js
@@ -1,5 +1,6 @@
 /* eslint-disable line-comment-position, no-new-func, no-undefined */
 import * as path from 'path';
+import os from 'os';
 
 import resolve from '@rollup/plugin-node-resolve';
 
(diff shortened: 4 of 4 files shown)


Chucky-ux (commented):
I think we are on the right way. 😊

pastelmind:
I added a test. Unfortunately, it works only on Windows, where replacing backslashes with slashes is perfectly fine. On Linux, the backslash is treated as a regular character, which means that we'd get an error if we replaced slashes with backslashes.

pastelmind:
@danielgindi I apologize for pinging you, but this PR hasn't even got a chance to run workflows yet. Could you please take a look?

pastelmind:
@danielgindi Thanks for letting the CI workflow happen, but it failed and I'm not sure why. The log is at https://github.com/rollup/plugins/runs/3150391648, here's an excerpt:
> Run pnpm run ci:test --filter ...[36a461e2f9775339554b7c76b1a2a8db408b9635]
>  pnpm run ci:test --filter ...[36a461e2f9775339554b7c76b1a2a8db408b9635]
>  shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
> ERROR  Filtering by changed packages failed. fatal: bad object 36a461e2f9775339554b7c76b1a2a8db408b9635
> Error: Process completed with exit code 1.

I rebased the commit onto the current `master` and force-pushed, but now the workflow needs approval again. I'm not familiar with the CI process--could you please help me troubleshoot this?

shellscape:
@pastelmind looks like your tests all pass now.  I've requested review from a few folks to get eyes on this for you.

danielgindi:
To me this looks much more resilient than the current code. Using core path functions handles this stuff for us.

danielgindi (approved):

lukastaegert (approved):

shellscape:
thanks!

PR #928, opened 2021-07-03
chore: add eslint as a dependency

tjenkinson:
## Rollup Plugin Name: `all`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description
I can't see `eslint` listed as a dependency anywhere so I think it exists as a side effect at the moment?

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "chalk": "^4.1.0",
     "codecov-lite": "^1.0.3",
     "del-cli": "^3.0.1",
+    "eslint": "^7.30.0",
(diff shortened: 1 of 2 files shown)


shellscape:
I've been procrastinating like a MF'er on the greater eslint alignment effort, and for that I apologize. I don't think we need this PR and that the original effort will solve this need. I'll do my best to make that happen in the next few days.

PR #933, opened 2021-07-09
fix(replace): prevent self-replacement of `preventAssignment` option

ctjlewis:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

### Description
Before, the `preventAssignment` option was not stripped from the replacements, meaning if `@rollup/plugin-replace` is executed on itself (i.e., a program which itself calls Rollup and uses `@rollup/plugin-replace`), it will rewrite the `preventAssignment` declaration, making it impossible to specify that argument.

This was encountered in the wild getting [TSDX](https://github.com/ctjlewis/tsdx) to self-compile. The first pass with `replace({ ... })` would indicate `preventAssignment: true` and not warn, but running the emitted, self-compiled output would indicate `preventAssignment` had been replaced and was `undefined` (I believe `{ preventAssignment: true }`  ➞ `{ true: true }`).

### Changes
Now the `preventAssignment` option, like other reserved options in the object passed to `replace({ ... })` (such as `delimiters`, `include`, and `exclude`), is removed from the replacement options, and will not be replaced by the plugin.

### Are tests included?
No.

### Breaking Changes?

- [x] no

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,14419 @@
+{
+  "name": "@rollup/plugins",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "@rollup/plugins",
+      "hasInstallScript": true,
+      "dependencies": {
+        "conventional-commits-parser": "^3.1.0",
+        "semver": "^7.3.2",
+        "write-pkg": "^4.0.0"
+      },
+      "devDependencies": {
+        "@ava/babel": "^1.0.1",
+        "@typescript-eslint/eslint-plugin": "^4.9.0",
+        "@typescript-eslint/parser": "^4.9.0",
+        "ava": "^3.15.0",
+        "chalk": "^4.1.0",
+        "codecov-lite": "^1.0.3",
+        "del-cli": "^3.0.1",
+        "eslint-config-rollup": "^1.0.0",
+        "esm": "^3.2.25",
+        "globby": "^11.0.1",
+        "husky": "^4.2.5",
+        "lint-staged": "^10.5.2",
+        "nyc": "^15.1.0",
+        "pnpm": "5.13.5",
+        "prettier": "^2.2.1",
+        "prettier-plugin-package": "^1.3.0",
+        "ts-node": "^8.10.2",
+        "tsconfig-paths": "^3.9.0",
+        "typescript": "^3.9.7",
+        "yaml": "^1.10.0"
+      }
+    },
+    "node_modules/@ava/babel": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/@ava/babel/-/babel-1.0.1.tgz",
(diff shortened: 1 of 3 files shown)


shellscape:
This is one of the edgiest edge cases I've run across 😄  While it's not marked as a breaking change, it is a breaking change. Anyone not trying to bundle the plugin and replacing things like `exclude` or `include` are going to be broken. For that reason, I don't think you should automatically prevent replacement of option keys. Instead, provide an option that if true, would prevent those replacements. That will leave backcompat in place and allow you to move forward. Impact will be minimal.

ctjlewis:
It's not really an edge case, this key was just not removed like the others, and so if you feed emitted output as input, it breaks. Feeding in output is pretty common with bundlers as we all well know.

Also, my secondary comments about the issue template were not disrespectful to the maintainers, only to the specific person that wrote the issue template.

ctjlewis:
Also, how is this a breaking change when the current functionality is broken?

This is what I mean about software brainworms: That does not make sense. It would "break" broken functionality because it's a bugfix.

I recommend merging the one-line change.

ctjlewis:
Alright, having thought about it, I don't see how that approach or most of that feedback makes any sense. Right now, the current options are screened out, so if I wanted to rename `sourceMap` or `include` calls or something today, it would not work, and it would read them as arguments to the `replace({ ... })` function. That is what happens *currently.*

https://github.com/rollup/plugins/blob/6867290ec8b983950ce20e58a34eaaee5e371517/packages/replace/src/index.js#L17-L28

So we have two options:
- add `preventAssignment` to the current "reserved" arguments, which cannot be replaced (current PR)
- add a `rollupPreventSelfReplacement` option, which would be the only "reserved" argument (since we can safely assume nobody would ever want to replace that exact string), and then if that is `true`, all of the current options, and `preventAssignment`, will be deleted in `getReplacements()` like above
  - *but*, this would be an *actual* breaking change, and also, the new default would be broken? as options like `{ sourceMap: true, includes: [...] }` would no longer be removed in `getReplacements()`, and would start being replaced???

### Conclusions
1. The PR should be merged as-is, because the alternatives do not make any sense.
2. Needing to replace a JS identifier that happens to be named `preventAssignment` is **a much rarer "edge case"** (analogous to happening to need to replace an identifier named `sourceMap` or something under current logic) than the transform being able to feedback its own output

The best approach involves the simple change made here.

ctjlewis:
cc @lukastaegert @Andarist

shellscape (changes requested):
> This plugin doesn't even have tests, which was how an error like this was able to come up in the first place.

There are quite a few with good coverage at 92.06%, located here: https://github.com/rollup/plugins/tree/master/packages/replace/test

We'll need a test to prevent regression against this fix. This review will be blocking until that's resolved, unfortunately.

guybedford (approved):
Thanks for this - it's a highly unfortunate namespace conflation, and seems like it was just omitted by mistake when implemented.

PR #941, opened 2021-07-15
chore(repo): Enforce prettier on CI

tjenkinson:
## Rollup Plugin Name: `na`

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

List any relevant issue numbers: #787

### Description
Some changes that were in #787 and missed in the version in master.

At the moment on CI prettier is run with `--write`, meaning I think any error would not cause an issue because the files would also be 'fixed' on CI.

With this PR we now enforce that prettier has been run on the files as they are in git, and also on every file type that prettier understands.

It also adds a few more extensions to the precommit hook, and removes the `--plugin` flag because it shouldn't be needed given the plugin is listed in the config file :)


shellscape:
Thanks for opening this one! Very sorry I missed that. As soon as CI is done we'll merge.

shellscape:
OK so we're getting a weird one:

```
> @rollup/plugins@ prettier:check /Users/powella/code/rollup/plugins
> prettier --check .

Checking formatting...
[warn] packages/dynamic-import-vars/CHANGELOG.md
[warn] Code style issues found in the above file(s). Forgot to run Prettier?
```

I can reproduce this locally. However, when I run prettier on that file it doesn't change the file. Super weird.

Issue #939, opened 2021-07-15
CommonJS plugin doesn't recognize the `__esModule` field in `module.exports = { __esModule: true, ... }`


sodatea:
- Rollup Plugin Name: commonjs
- Rollup Plugin Version: 19.0.0
- Rollup Version: 2.53.1
- Operating System (or Browser): macOS 11.4
- Node Version: 16.4.2
- Link to reproduction _(⚠️ read below)_: https://replit.com/@sodatea/rollup-esm-import-compiled-cjs

### Expected Behavior

Both imports from `symbol` and `symbol2` return the `.default` field because they are both compiled from an ES module and contain the `__esModule` field.

### Actual Behavior

The CommonJS plugin only recognizes the one with `exports.__esModule`. The one with `module.exports = { __esModule: true, default: ... }` is treated as a normal CJS module.

### Additional Information

The `module.exports = { __esModule: true, default: ... }` pattern is found in `babel-runtime` v6 https://unpkg.com/browse/babel-runtime@6.26.0/core-js/symbol.js

It is used in some old codebases. So this issue causes incompatibility with them.

See also https://github.com/vitejs/vite/issues/4090

guybedford:
Fixed in https://github.com/rollup/plugins/pull/942.

PR #942, opened 2021-07-16
fix(commonjs): convert module.exports with `__esModule` property(#939)

githoniel:

This PR contains:

- [X] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [X] no

List any relevant issue numbers: #939 

### Description

transform commonjs  as `module.exports = { __esModule: true, default: ... }` to wrap runtime code


guybedford:
@githoniel as well as the unit test here, could you include a full end-to-end test of the original use case provided in https://github.com/rollup/plugins/issues/939?

It looks like the output of that test is currently:

```
{ default: 'symbol', __esModule };
'symbol2'
```

Can you clarify what the output of that same test becomes with this PR?

githoniel:
@guybedford thank for your time. 

Look at this https://replit.com/@githoniel/rollup-plugin-repro-1

many old package output code like below

```
module.exports = { 
  "default": () => { console.log('beep') },
   __esModule: true 
};
```

they are not supported by current commojs plugin

guybedford:
One issue here is this is the opposite interop pattern to Node.js itself... making this plugin unsuitable for building Node.js applications.

githoniel:
any idea？shall we add options like `target: 'browse' | 'node'`

guybedford:
@githoniel I believe this is supposed to be controled by the `defaultIsModuleExports` configuration option - https://github.com/rollup/plugins/tree/master/packages/commonjs#defaultismoduleexports. So this PR should be sure to follow the semantics of that option as well I think.

githoniel:
> @githoniel I believe this is supposed to be controled by the `defaultIsModuleExports` configuration option - https://github.com/rollup/plugins/tree/master/packages/commonjs#defaultismoduleexports. So this PR should be sure to follow the semantics of that option as well I think.

I have make it work and update new tests

guybedford (approved):
Great work checking these cases, thanks for the PR.

guybedford:
I know the prettier issue is unrelated, but are you able to get the CI to pass on this PR?

guybedford:
Finally passing - thanks again for your work on this.

Issue #622, opened 2020-10-24
It should be clarified that plugin-commonjs should be placed before plugin-babel


shrinktofit:
- Rollup Plugin Name:  `plugin-commonjs`/`plugin-babel` 
- Rollup Plugin Version: latest

Documentation Is:

- [ ] Missing
- [x] Needed
- [ ] Confusing
- [ ] Not Sure?

### Please Explain in Detail...

Since how `@rollup/plugin-commonjs` and `@rollup/plugin-babel` are implemented. It's better to hint user that when specify rollup plugins, they'd better place `@rollup/plugin-commonjs` before `@rollup/plugin-babel`.

#### Reason

Given a module in CommonJS format and its code needs to be processed by babel. For example:
```js
module.exports = typeof window;
```

`@rollup/plugin-babel` will inject helper codes:
```js
import { typeof as _typeof } from "\0rollupPluginBabelHelpers.js";
module.exports = (typeof window === "undefined" ? "undefined" : _typeof(window));
```

then, the transformed source will be passed to `@rollup/plugin-commonjs`. The CommonJS plugin then treat the source as ES6 module(because of the `import` syntax `@rollup/plugin-babel` inserted) and does not process further and leads to `*** is not exported by ***`.

### Your Proposal for Changes

Document the order issue in plugin section.

shellscape:
Thanks for the issue. Happy to review a PR which adds this to the README.

jbt:
I came across this issue while running into problems with commonjs & babel - I was actually running into a similar issue but I found that often it doesn't work whichever way around you have the plugins.

For example, if you have syntax like JSX (which the commonjs plugin can't parse), you have to put babel before commonjs. But if your commonjs code requires things like the typeof helper reported above, then commonjs has to come before babel.

I've ended up having to include the babel plugin twice in my config (once before commonjs, once after) to make this work, which feels hacky.

Perhaps the circular-dependency between the two plugins is a bigger thing that needs addressing, I'm happy to try and summarise that in a new issue. But based on my experience it doesn't seem as simple as saying one plugin or the other ought to be placed first

shellscape:
resolved in 6b4b7b6

teehemkay:
> resolved in [6b4b7b6](https://github.com/rollup/plugins/commit/6b4b7b6748d5c043b6b56d02b39eb4c7e29c78ac)

IMHO the recommandation should be further contextualized and more nuanced as pointed out by @jbt above and @Andarist in #805.

[...]

PR #971, opened 2021-08-07
fix(commonjs): Support JSX and TypeScript

onigoetz:

This PR contains:

- [x] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: #622

### Description

It is written in the documentation that the commonjs plugin goes _before_ the babel plugin.
It works fine for most cases. But in some cases, like using TypeScript or JSX; this won't work as the parsing of the source will fail.
One workaround is to do the babel transpilation before, the risk with this approach though is that babel adds helpers (using `import`) and breaks the module detection of `commonjs`

The only safe approach would be to make sure that any syntax close enough to EcmaScript is parsed properly and re-output them identically; `acorn-loose` allows to do this.

__Downside of this solution__:

`acorn-loose` will make a valid AST even from the most broken code. So syntax that shouldn't be valid at all will parse correctly. Is it a problem ? in my opinion no, I think projects use linters or other tools to ensure their code is valid and this shouldn't happen too often.

__Alternatives considered__:

- Changing babel to issue helpers using commonjs syntax; looks complicated to do as this would require to change the TypeScript and swc logic as well
- Adding a first pass of babel to remove JSX, run commonjs, then babel again for the rest; Too complex to advertise, inefficient because the code is parsed again and again, also it won't work with TypeScript and SWC.

### DIFF
--- a/packages/commonjs/package.json
+++ b/packages/commonjs/package.json
@@ -51,6 +51,7 @@
   },
   "dependencies": {
     "@rollup/pluginutils": "^3.1.0",
+    "acorn-loose": "^8.1.0",
     "commondir": "^1.0.1",
     "estree-walker": "^2.0.1",
     "glob": "^7.1.6",
--- a/packages/commonjs/src/parse.js
+++ b/packages/commonjs/src/parse.js
@@ -1,7 +1,15 @@
+import acornLoose from 'acorn-loose';
+
 export function tryParse(parse, code, id) {
   try {
     return parse(code, { allowReturnOutsideFunction: true });
   } catch (err) {
+    try {
+      return acornLoose.parse(code, { allowReturnOutsideFunction: true, ecmaVersion: 2020 });
+    } catch (err2) {
+      // We ignore this second try and just report the original error
+    }
+
     err.message += ` in ${id}`;
     throw err;
   }
--- a/packages/commonjs/test/fixtures/form/jsx/input.js
+++ b/packages/commonjs/test/fixtures/form/jsx/input.js
@@ -1,0 +1,5 @@
+const React = require("React");
+
+module.exports = function Button(props) {
+  return <button {...props} />;
+}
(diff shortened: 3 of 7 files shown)


onigoetz:
Hi @shellscape is there anything I can do to help this get merged ?
Thanks in advance

shellscape:
I've requested review by some of the folks who know this plugin well.

onigoetz:
Hi. I do understand your concern. 

I'm afraid just transpiling JSX with babel won't work. As it will add it's own imports: 

using https://babeljs.io/docs/en/babel-plugin-transform-react-jsx/ on

```
const profile = (
  <div>
    <img src="avatar.png" className="profile" />
    <h3>{[user.firstName, user.lastName].join(" ")}</h3>
  </div>
);
```

Will produce 

```
import { jsx as _jsx } from "react/jsx-runtime";
import { jsxs as _jsxs } from "react/jsx-runtime";

const profile = _jsxs("div", {
  children: [
    _jsx("img", {
      src: "avatar.png",
      className: "profile",
    }),
    _jsx("h3", {
      children: [user.firstName, user.lastName].join(" "),
    }),
  ],
});
```

Which will also break the commonjs detection

I'm pretty sure that having some TypeScript input using enums or some more advanced features will also need some helpers to be transpiled enough to be parsed well by acorn.

double-layering the Babel plugin is one more level of parsing and that seems very inefficient anyway.

shellscape:
Closing this one as it hasn't gotten any recent traction. I think the core team is in agreement with @guybedford that loose parsing isn't something we'd like to introduce into the commonjs plugin. However, there may be some room to introduce a third party plugin that does perform loose commonjs parsing. As there hasn't been a large call for that, a third party plugin (or a fork of commonjs) is probably the route to go. Thanks for your work on this and for your willingness to contribute to the project.

onigoetz:
Thanks for the feedback

Issue #974, opened 2021-08-15
[@rollup/plugin-html] `html` (default export) function's type declaration should list `options` argument as optional


jameone:
- Rollup Plugin Name: 
- Rollup Plugin Version:

### Expected Behavior / Situation
When using a `rollup.config.ts` file (as opposed to a `rollup.config.js` file), the plugin should be able to be called as outlined in the docs like so:
```ts
import typescript from '@rollup/plugin-typescript';
import html from '@rollup/plugin-html';

export default {
    input: 'src/index.ts',
    output: {
        file: 'dist/bundle.js',
        format: 'umd',
        name: 'package',
    },
    plugins: [
        typescript(),
        html(), // <- the important bit
    ],
};

```

### Actual Behavior / Situation
In actuality it needs to be called like so:
```ts
import typescript from '@rollup/plugin-typescript';
import html from '@rollup/plugin-html';

export default {
    input: 'src/index.ts',
    output: {
        file: 'dist/bundle.js',
        format: 'umd',
        name: 'package',
    },
    plugins: [
        typescript(),
        html({}), // <- notice an argument is required
    ],
};

```

### Modification Proposal
In the type declarations (`packages/html/types/index.d.ts`), the function signature of html should be updated and the `options` argument labelled as optional.

PR #975, opened 2021-08-15
refactor(html): function signature of html updated

ghost:
## Rollup Plugin Name: `html`

This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

#974 
### Description

When using a `rollup.config.ts` file I am seeing a linting error when using the plugin in the prescribed way (with no arguments). Given the function definition uses a default, I think the type declaration can be amended to list the options argument as optional. This way, people do not get linting errors/have to add a redundant default argument: eg.`html({})`.

EDIT: Added `html` to name and a fixed typo.

### DIFF
--- a/packages/html/types/index.d.ts
+++ b/packages/html/types/index.d.ts
@@ -25,4 +25,4 @@ export function makeHtmlAttributes(attributes: Record<string, string>): string;
  * @param options - Plugin options.
  * @returns Plugin instance.
  */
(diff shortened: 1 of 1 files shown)


shellscape:
thanks!

PR #1000, opened 2021-09-23
refactor(alias): avoid resolving customResolver every time handling resolveId

unbyte:
## Rollup Plugin Name: `@rollup/plugin-alias`

This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

This PR removes useless functions and avoid resolving customResolver every time handling resolveId.

### DIFF
--- a/packages/alias/src/index.ts
+++ b/packages/alias/src/index.ts
@@ -57,7 +57,8 @@ export default function alias(options: RollupAliasOptions = {}): Plugin {
 
   if (entries.length === 0) {
     return {
-      name: 'alias'
+      name: 'alias',
+      resolveId: () => null
     };
   }
 
--- a/packages/alias/src/index.ts
+++ b/packages/alias/src/index.ts
@@ -1,8 +1,6 @@
-import { PartialResolvedId, Plugin } from 'rollup';
+import { Plugin } from 'rollup';
 
-import { Alias, ResolverFunction, RollupAliasOptions } from '../types';
-
-const noop = () => null;
+import { ResolvedAlias, ResolverFunction, ResolverObject, RollupAliasOptions } from '../types';
 
 function matches(pattern: string | RegExp, importee: string) {
   if (pattern instanceof RegExp) {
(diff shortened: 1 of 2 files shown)


shellscape:
Thanks for the PR. To fully understand your case for this change, we need to see some examples of how this was faulty before your changes. typically that's done via a previously failing test, but occasionally that's not usable for a demonstration. If that's the case, please provide a fair amount more info on why this change is necessary. That's important since the plugin is considered stable.

unbyte:
I agree that this plugin is stable, and I haven't encountered any problems in using it. 

The reason why I open this PR is that, the documentation has no description of the resolving algorithm, so I looked at the code and then I found that there are redundant functions and some logic that can have readability improvemet.

In general, this PR does not involve any functional improvement, but simply makes the code concise.

shellscape:
Thanks for the explanation. We'll have a few folks take a look at the refactor.

danielgindi:
This looks much cleaner, and I love that's move usage of `await` is implemented!
I'm worried about two things:
1. The need for a "fix" in the tests - why did that happen? Will it possibly break for someone in the real world?
2. The amount of "already defined in upper scope" warnings. These warnings are not there just for fun, they can point on a confusion in the code which leads to an obscure bug later.

unbyte:
@danielgindi hi, thank you for your review❤️

The commit `fix test` has nothing to do with tests in fact. I just kept the return value when no entry passed consistent with the previous implementation. I accidentally missed it in the first commit.

And I don't know why there are so many warning; it doesn't seem to be caused by my changes🤔

danielgindi:
> As for these warnings, I am sure they are not caused by my changes, because they also exist in other PR s. Maybe other packages should also be refactored? (just a suggestion, because I see many plugins have different code styles and be in different languages.)

Nevertheless, refactoring while these warnings exists is quite dangerous.
But let's trust the tests here and pass this.

PR #1012, opened 2021-10-01
feat(auto-install): support pnpm

antfu:
## Rollup Plugin Name: `auto-install`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

Support detection for `pnpm` as a manager

### DIFF
--- a/packages/auto-install/src/index.ts
+++ b/packages/auto-install/src/index.ts
@@ -15,15 +15,16 @@ export default function autoInstall(opts: RollupAutoInstallOptions = {}): Plugin
     // intentionally undocumented options. used for tests
     commands: {
       npm: 'npm install',
+      pnpm: 'pnpm install',
       yarn: 'yarn add'
     },
-    manager: fs.existsSync('yarn.lock') ? 'yarn' : 'npm',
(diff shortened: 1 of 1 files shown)


shellscape (approved):

shellscape:
Looks like there are some failing tests you'll have to resolve before we can merge this. We'll need tests around adding pnpm as well.

antfu:
In Node v16, npm use lock v2 which makes the lock snap failed. Maybe switching to exist testing instead of lock snapshot (similar to yarn's)

shellscape:
Tests probably need a tweak to assert npm v6.x for the test. You're welcome to address that here or we can tackle it after this PR.

antfu:
I am not sure what's the initial decision to snap on `package-lock.json`, but from my point of view, I feel that having the package existed in `package.json` is already good enough to prove the auto-installation is working. Making sure the lock is stable is more like a thing that needs to be generated on the NPM/Node side. Giving the different lock formats for Node v14 / v16, I think it makes things easier to not snap on the lock.

I have made the changes to the test. Feel free to commit to the PR directly if have different thoughts.

shellscape:
I'm good with that.

PR #1043, opened 2021-11-16
test(node-resolve): add tests for mixing custom `exportConditions` with `browser: true`

Andarist:

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] yes
- [x] no

### Description

I was wondering if this would work correctly - I've figured out that the quickest way to found out would be to check the test suite but I couldn't find **any** `exportConditions`-related test. The only place where "exportCondition:" string appears in the codebase is this type test:
https://github.com/rollup/plugins/blob/53776ee3fc59587fe5ff0c756216185f48372507/packages/node-resolve/test/types.ts#L14

Since it felt like such tests are missing, I've decided to just add them 😉

### DIFF
--- a/packages/node-resolve/test/fixtures/exports-only-worker-condition-with-browser-field.js
+++ b/packages/node-resolve/test/fixtures/exports-only-worker-condition-with-browser-field.js
@@ -1,0 +1,3 @@
+import sample from 'exports-only-worker-condition-with-browser-field';
+
+export default sample;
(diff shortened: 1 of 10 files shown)


guybedford (approved):
Thanks for posting these. Note the lint issue / console.log.

Andarist:
@guybedford fixed 🤦

Andarist:
Since this is approved and CI passes - can I just land this? I'm slightly unsure and I don't want to get into the way of The Process 😉

shellscape (approved):

PR #1047, opened 2021-11-21
feat(pluginutils): support bigint and symbol in dataToEsm

LongTengDao:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

|data|oldEsm|newEsm|
|---|---|---|
|`bigint`|`undefined`|`bigint`|
|`Symbol.for()`|`undefined`|`Symbol.for()`|

### DIFF
--- a/packages/pluginutils/src/dataToEsm.ts
+++ b/packages/pluginutils/src/dataToEsm.ts
@@ -50,10 +50,9 @@ function serialize(obj: unknown, indent: Indent, baseIndent: string): string {
   if (typeof obj === 'bigint') return `${obj}n`;
   if (typeof obj === 'symbol') {
     const key = Symbol.keyFor(obj);
-    if (key === undefined) throw TypeError('dataToEsm can not serialize symbol');
-    return `Symbol.for(${stringify(key)})`;
+    return key === undefined ? 'undefined' : `Symbol.for(${stringify(key)})`;
   }
-  if (typeof obj === 'function') throw TypeError('dataToEsm can not serialize function');
+  if (typeof obj === 'function') return 'undefined';
   return stringify(obj);
 }
 
--- a/packages/pluginutils/src/dataToEsm.ts
+++ b/packages/pluginutils/src/dataToEsm.ts
@@ -4,7 +4,7 @@ import makeLegalIdentifier from './makeLegalIdentifier';
 
 export type Indent = string | null | undefined;
 
-function stringify(val: boolean | number | string): string {
+function stringify(val: unknown): string {
   return JSON.stringify(val).replace(
     /[\u2028\u2029]/g,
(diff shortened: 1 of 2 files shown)


packages/pluginutils/test/dataToEsm.ts
```diff
@@ -2,6 +2,13 @@ import test from 'ava';
 
 import { dataToEsm } from '../';
 
+test('support bigint and symbol', (t) => {
```
shellscape, line 5: please split these into separate tests. (that may seem pedantic, but this PR essentially adds two features in support for two different entities)

packages/pluginutils/test/dataToEsm.ts
```diff
@@ -2,6 +2,13 @@ import test from 'ava';
...
+  t.is(
+    dataToEsm({ bigint: 0, symbol: Symbol.for('') }),
```
shellscape, line 7: I'd like to see a few different values (within the same `t` test func) tried here for sanity check

packages/pluginutils/src/dataToEsm.ts
```diff
@@ -4,8 +4,8 @@ import makeLegalIdentifier from './makeLegalIdentifier';
 
 export type Indent = string | null | undefined;
 
-function stringify(obj: unknown): string {
-  return (JSON.stringify(obj) || 'undefined').replace(
+function stringify(val: boolean | number | string): string {
```
shellscape, line 7: why is this change necessary?

shellscape (changes requested):

shellscape:
Looks like there are some failures caused by these changes in CI. Please have a look

packages/pluginutils/src/dataToEsm.ts
LongTengDao: because rest types (`undefined` `object` `symbol` `bigint`) already and should be handled in `serialize`, I think this will make the code more strong.

if it's not necessary, this change could be reverted.

LongTengDao: I reverted this change for CI undue error.

shellscape (approved):

shellscape:
Closing as abandoned. The error is visible here: https://github.com/rollup/plugins/runs/5281573593?check_suite_focus=true#step:7:241. If you get around to cleaning that up, please update your fork and give us a ping.

LongTengDao:
@shellscape Oh, sorry, I missed. The typo error has been corrected.

shellscape:
Thanks for updating it!

shellscape:
Looks like there is a failing test to take a look at.

LongTengDao:
@shellscape Thanks for your careful guidance

PR #1054, opened 2021-11-26
fix(node-resolve): prevent passing invalid argument to fs.stat.

msyfls123:
Can't figure out why `undefined` is passed by, just undertake it.

## Rollup Plugin Name: node-resolve

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### DIFF
--- a/packages/node-resolve/src/index.js
+++ b/packages/node-resolve/src/index.js
@@ -199,7 +199,7 @@ export function nodeResolve(opts = {}) {
       browserMapCache.set(location, packageBrowserField);
     }
 
-    if (hasPackageEntry && !preserveSymlinks) {
+    if (hasPackageEntry && !preserveSymlinks && location) {
       const exists = await fileExists(location);
       if (exists) {
         location = await realpath(location);
@@ -221,7 +221,7 @@ export function nodeResolve(opts = {}) {
       }
     }
 
-    if (options.modulesOnly && (await fileExists(location))) {
+    if (options.modulesOnly && location && (await fileExists(location))) {
       const code = await readFile(location, 'utf-8');
       if (isModule(code)) {
         return {


shellscape:
Thanks for the PR!

> _bugfixes and features will not be merged without tests_

We'll need a test that replicates the situation, or at the very least guards against that default parameter being removed thus introducing a regression. We will not merge the PR without one.

msyfls123:
> We'll need a test that replicates the situation, or at the very least guards against that default parameter being removed thus introducing a regression. We will not merge the PR without one.

I'll try to make such a test, but it may cost some time since I'm a little busy.

msyfls123:
Finally, I decide to put a guard statement here since it will prevent `undefined` being passed to `fs.stat`.
```diff
export async function fileExists(filePath) {
  try {
+   if (!filePath) return false;
    const res = await stat(filePath);
    return res.isFile();
  } catch {
    return false;
  }
}
```

Since the `try ... catch` block won't throw TypeError but my VSCode Debug task just catch it, I can't present a situation without modifying raw code.

__Before:__ both `ENOTDIR` error and `TypeError` are caught, and `TypeError` means nothing.
image

__After:__ only `ENOTDIR` error is caught.
image

shellscape:
@msyfls123 we'd like to merge this but we must have an accompanying test. If you don't have the time to add one we will understand completely.

guybedford:
Yes, we need to know what is triggering the scenario in the first place, this should really be the contract of the function with the error thrown before this function is even called.

msyfls123:
Well, I just saw the comment. Since
the error only occurred in Typescript checks, I'm not very sure how to create a test to cover this situation. 

Maybe such a test: There should be a more aggressive guard to prevent any 'undefined' to be passed to the function, and try to prove that there is at least one 'undefined' being
 passed to the function, and the guard works.

guybedford:
There's no guarantee that these JS files pass TS typechecking, I'm not sure what that is trying to achieve?

shellscape:
Closing as abandoned.

PR #1073, opened 2021-12-23
refactor(dynamic-import-vars): replace globby with fast-glob

TrySound:

This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

Looks like globby features are not used in this plugin and internal
fast-glob can be used instead. It also has a bit smaller install size.

https://packagephobia.com/result?p=globby
https://packagephobia.com/result?p=fast-glob

Bumped pluginutils too.

### DIFF
--- a/packages/dynamic-import-vars/package.json
+++ b/packages/dynamic-import-vars/package.json
@@ -51,9 +51,9 @@
     "rollup": "^1.20.0||^2.0.0"
   },
   "dependencies": {
-    "@rollup/pluginutils": "^3.1.0",
+    "@rollup/pluginutils": "^4.1.2",
     "estree-walker": "^2.0.1",
-    "globby": "^11.0.1",
+    "fast-glob": "^3.2.7",
     "magic-string": "^0.25.7"
   },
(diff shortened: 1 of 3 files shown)


shellscape:
Generally speaking we don't care so much about dependency size in plugins. They're not meant to be used anywhere than CI/local dev environments and it's just not a priority worry. Globby is always my go-to for globs so I'm curious what features we're not using.

TrySound:
- streaming api
- expanding directories
- gitignore support

And non of these features are used. This is the only plugin using globby.

Less dependencies -> more disk space and faster install. Rollup was always the best in handling dependencies by bundling them. With plugins it's easy to fall into webpack trap with million of dependencies and very long install time and even worse with dependent tools. This is my opinion anyway.

shellscape (approved):
I'd chalk this up to a complexity reduction, since this makes the plugin use the base dependency of a prior dependency, without all of the sugar we don't need.

PR #1084, opened 2022-01-08
feat(replace): typeofReplacements option

guybedford:
## Rollup Plugin Name: `plugin-replace`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

Currently `process.env.NODE_ENV` replacements only replace that exact string. But as came up in https://github.com/reactjs/react-transition-group/pull/777, it can be useful for browser code to guard these replacements like:

```js
if (typeof process !== 'undefined' && process.env.NODE_ENV === 'production') {...}
```

by default this plugin will not handle these other guards properly.

This PR adds a new `typeofReplacements: true` option (disabled by default, although it would be nice to enable by default in future), which will replace the `typeof process` with `"object"` in the case of identifier member expression replacement strings.

The user only needs to define `process.env.NODE_ENV` as the replacement key, and it will automatically expand the parent object typeof replacements when using this option.

See the README for a further example.


guybedford:
It's very tricky to get these snapshots to line up when working on Windows. Everything is passing, but would be appreciated if someone with a unix machine could update the snapshots here. Alternatively when I next have access to WSL I can take a look.

shellscape (approved):

shellscape:
The concept is sound. Not so sure about the option name. To anyone not familiar with the use case, I don't think the option name conveys what the option does.

guybedford:
Thanks for the review @shellscape I've renamed the feature to `objectGuards` if that seems like a more descriptive name?

shellscape:
Sounds good to me

packages/replace/test/snapshots/form.js.md
```diff
@@ -39,6 +39,14 @@ Generated by [AVA](https://avajs.dev).
       exclude␊
     `);`
 
+## process-check: Handles process type guards in replacements
+
+> Snapshot 1
+
+    `if (typeof process !== 'undefined' && "production" === 'production') {␊
```
privatenumber, line 46: @guybedford 

Shouldn't `process` be transformed to `'object'`? :
```
if ('object' !== 'undefined' && "production" === 'production') {
```

PR #1088, opened 2022-01-09
fix(replace)!: do not escape delimiters

### LINKED ISSUE #904: replace plugin. "delimiters" config is broken by "escape" internal mutation
- Rollup Plugin Name: `@rollup/plugin-replace`
- Rollup Plugin Version: `2.4.2` (currently latest)
- Rollup Version: `2.50.6`
- Operating System (or Browser): `macos` (but it's irrelevant in this case) 
- Node Version: `16` (but it's irrelevant in this case)
- Link to reproduction: https://stackblitz.com/edit/rollup-plugin-replace-broken-delimiters-option

### Expected Behavior

The `delimiters option` that is matching the `default value` should work the same

### Actual Behavior

Unfortunately, the `escape` internal function that is supposed to escape RegExp string malfunctions.
So it's impossible to reproduce the current default behavior because of that.
- it's also preventing to configure as this PR suggests https://github.com/rollup/plugins/pull/903

guansss:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] no

List any relevant issue numbers: #904 

### Description

Fixes #904, now delimiters given by users are no longer escaped, so it's possible to pass `['\\b', '\\b']` to revert to the default behaviour prior to #903.

It's breaking because users may have been relying on current behaviour (bug) to match special characters which are now required to be escaped.


danielgindi (approved):

shellscape:
@guansss apologies for the delay in getting to your PR. it looks like there are a few conflicts to clean up since you submitted it. Please clean those up and we'll give this another look.

@guybedford could you lend your eyeballs to this breaking change?

guansss:
Okay, I've rebased it onto master.

guybedford (approved):
Great work. Looks solid to me. Is the `escape` function still used elsewhere or can it be removed?

This is a major change so we should bump the semver version.

guybedford:
Another thing that could be useful is accepting an actual regex for delimiters by value as well eg via `if (delimiter instanceof RegEx)` kind of thing.

guansss:
Thanks @guybedford. Yeah, the `escape` function is still used on replacement keys.

The delimiters will end up being concatenated into a string to build the regex, so there may be no need to accept a regex.

guybedford:
@guansss accepting a regex may still solve the original use case of supporting an unescaped input though!

guansss:
@guybedford that makes sense, but IMO it's more like a trick and could be easily done from user-side via something like `delimiters: [/\b/.source, /\b/.source]`.

guybedford:
@guansss at the end of the day manual regex escaping is work, and something that can be a cause of errors. Making things easier where possible without sacrificing maintainability is the goal in open source!

guansss:
@guybedford agreed, I'm not quite into it though!😄

PR #1090, opened 2022-01-13
fix(babel): consider path delimeter on windows (#1089)

### LINKED ISSUE #1089: [plugin-babel] Unexpected situation when using with babel-preset-react-app on Windows
- Rollup Plugin Name: plugin-babel
- Rollup Plugin Version: 5.3.0
- Rollup Version: 2.63
- Operating System (or Browser): Windows
- Node Version: v14.15.5
- Link to reproduction: https://regex101.com/r/y6BLj8/2

### Expected Behavior
PreflightChecks are passed

### Actual Behavior
Error: An unexpected situation arose. Please raise an issue at https://github.com/rollup/plugins/issues. Thanks!

### Additional Information
https://github.com/facebook/create-react-app/blob/a422bf227cf5294a34d68696664e9568a152fd8f/packages/babel-preset-react-app/create.js#L194 is using 

```
absoluteRuntimePath = path.dirname(
      require.resolve('@babel/runtime/package.json')
    );
```

which on Windows evaluates to ```<my-source-path>\\rollup-babel-bug\\node_modules\\@babel\\runtime```. This breaks the preflightCheck at https://github.com/rollup/plugins/blob/9874740ca685b1600e4319365c21e8c1131e9d95/packages/babel/src/preflightCheck.js#L48, as the regex is not considering backslashes.

https://github.com/rollup/plugins/blob/9874740ca685b1600e4319365c21e8c1131e9d95/packages/babel/src/preflightCheck.js#L37 should be changed to:

```
const inheritsHelperRe = /[\\\/]+helpers[\\\/]+(esm[\\\/]+)?inherits/;
```

If possible to verify on Windows I can provide a reproduction repo.

No matter what, I'll create a PR for the change.

atti187:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

What it says on the tin - fixes #1089

### DIFF
--- a/packages/babel/src/preflightCheck.js
+++ b/packages/babel/src/preflightCheck.js
@@ -34,7 +34,7 @@ function helpersTestTransform() {
 const mismatchError = (actual, expected, filename) =>
   `You have declared using "${expected}" babelHelpers, but transforming ${filename} resulted in "${actual}". Please check your configuration.`;
 
-// Revert to /\/helpers\/(esm\/)?inherits/ once https://github.com/babel/babel/issues/14185 is fixed
+// Revert to /\/helpers\/(esm\/)?inherits/ when Babel 8 gets released, this was fixed in https://github.com/babel/babel/issues/14185
 const inheritsHelperRe = /[\\/]+helpers[\\/]+(esm[\\/]+)?inherits/;
 
 export default async function preflightCheck(ctx, babelHelpers, transformOptions) {
--- a/packages/babel/src/preflightCheck.js
+++ b/packages/babel/src/preflightCheck.js
@@ -34,7 +34,7 @@ function helpersTestTransform() {
 const mismatchError = (actual, expected, filename) =>
   `You have declared using "${expected}" babelHelpers, but transforming ${filename} resulted in "${actual}". Please check your configuration.`;
 
-const inheritsHelperRe = /[\\\/]+helpers[\\\/]+(esm[\\\/]+)?inherits/;
+const inheritsHelperRe = /[\\/]+helpers[\\/]+(esm[\\/]+)?inherits/;
 
 export default async function preflightCheck(ctx, babelHelpers, transformOptions) {
   const finalOptions = addBabelPlugin(transformOptions, helpersTestTransform);
--- a/packages/babel/src/preflightCheck.js
+++ b/packages/babel/src/preflightCheck.js
@@ -34,7 +34,7 @@ function helpersTestTransform() {
 const mismatchError = (actual, expected, filename) =>
   `You have declared using "${expected}" babelHelpers, but transforming ${filename} resulted in "${actual}". Please check your configuration.`;
 
-const inheritsHelperRe = /\/helpers\/(esm\/)?inherits/;
+const inheritsHelperRe = /[\\\/]+helpers[\\\/]+(esm[\\\/]+)?inherits/;
 
 export default async function preflightCheck(ctx, babelHelpers, transformOptions) {
   const finalOptions = addBabelPlugin(transformOptions, helpersTestTransform);
--- a/packages/babel/src/preflightCheck.js
+++ b/packages/babel/src/preflightCheck.js
@@ -34,6 +34,7 @@ function helpersTestTransform() {
 const mismatchError = (actual, expected, filename) =>
   `You have declared using "${expected}" babelHelpers, but transforming ${filename} resulted in "${actual}". Please check your configuration.`;
 
+// Revert to /\/helpers\/(esm\/)?inherits/ once https://github.com/babel/babel/issues/14185 is fixed
 const inheritsHelperRe = /[\\/]+helpers[\\/]+(esm[\\/]+)?inherits/;
 
 export default async function preflightCheck(ctx, babelHelpers, transformOptions) {


shellscape:
Please correct the CI errors

atti187:
I have no idea why there are CI errors in your build pipeline...

Look at the change and tell me it's because of it!

shellscape:
Are you saying that because you don't have much experience with submitting pull requests that use Github actions to validate pull requests, or are you confused as to where the error is?

If it's the latter, you can see how/why your changes caused CI to fail in two places:

- https://github.com/rollup/plugins/pull/1090/files (the Github UX tells you that your code change caused a linting error)
- https://github.com/rollup/plugins/runs/4803955163?check_suite_focus=true#step:10:10

As a practice, you should always look for a `lint` and/or `build` in package.json for the project you're editing, and run either or both _before_ submitting a pull request. This repo uses `pnpm`, so you should have run:

```console
$ cd packages/babel
$ pnpm lint
```

atti187:
The latter - as running the linting on a Windows machine, hundreds (if not thousands) of errors showed up. I just assumed it was the same stuff showing up, and/or issues with the node images - have happened before.

Anyway, I can take care of it tomorrow on my Ubuntu instance on WSL.

atti187:
Done!

packages/babel/src/preflightCheck.js
```diff
@@ -34,7 +34,7 @@ function helpersTestTransform() {
 const mismatchError = (actual, expected, filename) =>
   `You have declared using "${expected}" babelHelpers, but transforming ${filename} resulted in "${actual}". Please check your configuration.`;
 
-const inheritsHelperRe = /\/helpers\/(esm\/)?inherits/;
+const inheritsHelperRe = /[\\/]+helpers[\\/]+(esm[\\/]+)?inherits/;
```
Andarist, line 37: Hm, I'm rather surprised that this would be needed. I have never seen a windows path delimiter in the import specifier. Isn't this normalized by Babel and all the tools to use POSIX delimiter? 

Do you have a repro case of the problem? I would love to take a look at that.

Andarist: Ok, I've found out the repro case in the linked ticket - thanks for that and for the detailed description of the problem there.

I think that the best thing to do would be to first try using [`normalize-path`](https://github.com/jonschlinkert/normalize-path) on this computed path:
https://github.com/facebook/create-react-app/blob/a422bf227cf5294a34d68696664e9568a152fd8f/packages/babel-preset-react-app/create.js#L51-L55

If a PR implementing such a change won't land in CRA cause their maintainers won't have time/bandwidth to handle this then we could consider proceeding with this PR here.

You can also try disabling the `absoluteRuntime` option in your `babel-preset-react-app`'s options. I didn't have time to dig into why this is even the default etc but I think that for the vast majority of use cases this actually shouldn't be used.

atti187: Ok - I'll create a PR for them and reference this issue, and we'll see where it takes us. I the meantime, I'll just skip the preflight checks.

atti187: I dug into this a little bit more... seems like the windows-style import comes from https://github.com/babel/babel/blob/6ec66d8d2249e345a587d8d47722abd779be6416/packages/babel-plugin-transform-runtime/src/index.ts#L241. The absoluteRuntime config _value_ isn't used in this computation, so changes in CRA has no effect.

So the question is then - should we try to land a change in babel-plugin-transform-runtime, or stick with the regex fix in this PR?

Andarist: > So the question is then - should we try to land a change in babel-plugin-transform-runtime, or stick with the regex fix in this PR?

I'm leaning towards... both 😅 I would appreciate it if you could:
- prepare a PR to Babel
- add a comment in the code about this issue and link to the created PR

packages/babel/src/preflightCheck.js
```diff
@@ -34,7 +34,8 @@ function helpersTestTransform() {
 const mismatchError = (actual, expected, filename) =>
   `You have declared using "${expected}" babelHelpers, but transforming ${filename} resulted in "${actual}". Please check your configuration.`;
 
-const inheritsHelperRe = /\/helpers\/(esm\/)?inherits/;
+// Revert to /\/helpers\/(esm\/)?inherits/ once https://github.com/babel/babel/issues/14185 is fixed
```
Andarist, line 37:
```suggestion
// Revert to /\/helpers\/(esm\/)?inherits/ when Babel 8 gets released, this was fixed in https://github.com/babel/babel/issues/14185
```

Andarist: Since the problem turned out to be related to Babel itself we might want to keep this adjusted regex for now - given how popular Babel is we might as well go with this patch and end up being more friendly to users

shellscape:
@Andarist @atti187 checking in on this PR.

Andarist (approved):

Andarist:
@shellscape I've approved the changes - feel free to land this

PR #1104, opened 2022-02-10
feat(node-resolve): add new option, modulePaths

ajhyndman:
## Rollup Plugin Name: `plugin-node-resolve`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

This change introduces a new option for @rollup/plugin-node-resolve: **`modulePaths`**.

Unlike `moduleDirectories`, `modulePaths` allows the user to specify an absolute path to alternative locations (in addition to the importer's local directory) that should be searched for node modules.

This can be useful (for example) to support a CI system that caches node_modules in a non-standard location.

[In node, this is supported via the `NODE_PATH` environment variable](https://nodejs.org/api/modules.html#loading-from-the-global-folders).  Conveniently, this also brings @rollup/plugin-node-resolve into alignment with Jest's configuration options.  See: https://jestjs.io/docs/configuration#modulepaths-arraystring


shellscape:
@tjenkinson please have a peek at this one

packages/node-resolve/README.md
```diff
@@ -75,7 +75,14 @@ If `true`, instructs the plugin to use the browser module resolutions in `packag
...
 
-One or more directories in which to recursively look for modules.
+A list of directory names in which to recursively look for modules.
+
+### `modulePaths`
+
+Type: `Array[...String]`<br>
+Default: `[]`
+
+A list of absolute paths to additional locations to search for modules. This is analogous to setting the `NODE_PATH` environment variable for node.
```
tjenkinson, line 85:
```suggestion
A list of absolute paths to additional locations to search for modules. [This is analogous to setting the `NODE_PATH` environment variable for node](https://nodejs.org/api/modules.html#loading-from-the-global-folders).
```

packages/node-resolve/test/test.js
```diff
@@ -257,6 +257,36 @@ test('allows custom moduleDirectories with legacy customResolveOptions.moduleDir
   t.snapshot(warnings);
 });
 
+test('custom moduleDirectories do not support nested dependencies', async (t) => {
```
tjenkinson, line 260: I think it's good to test this but it would be nice to also actually flag this as a proper error if the user tries to provide a path instead of name for this option.

Please could we update the code to something like:
```js
if (moduleDirectories.some((name) => name.includes('/'))) {
  throw new Error(
    '`moduleDirectories` option must only contain directory names. If you want to load modules from a custom directory see `modulePaths`.'
  );
}
```
in index.js

Then this test can just check that
```js
nodeResolve({
  moduleDirectories: ['some/path']
})
```
throws that error

packages/node-resolve/test/test.js
```diff
@@ -257,6 +257,36 @@ test('allows custom moduleDirectories with legacy customResolveOptions.moduleDir
...
+    input: 'custom-module-path/main.js',
+    onwarn: failOnWarn(t),
+    plugins: [
+      nodeResolve({
+        modulePaths: [join(process.cwd(), 'custom-module-path/node_modules')]
+      })
+    ]
+  });
+
+  t.is(bundle.cache.modules.length, 3);
```
tjenkinson, line 287:
```suggestion
  const { dependency } = await evaluateBundle(bundle);
  t.is(dependency, 'DEPENDENCY');
```
is a bit clearer IMO and checks the actual export is there :)

You'll also need to add the import for `evaluateBundle` from `../../../util/test`.

tjenkinson (commented):
thanks @ajhyndman!

Looks good. Got a few suggestions

packages/node-resolve/test/test.js
ajhyndman: Yeah, that sounds awesome!  Would this change be breaking though?  Should it require a major version bump?

tjenkinson: If you use a path doesn’t it just get ignored currently?

If that’s the case and it silently fails (with the warning when module not found) IMO this doesn’t need to be a breaking change as it’s highlighting an invalid config

tjenkinson: What’s the existing behaviour you’re relying on?

ajhyndman: If you pass a relative path, it appears that the same logic for folder names is applied.  i.e.  When searching for node modules, rollup will recursively walk up the filesystem, at each level looking for a the full subpath specified.

I think we inadvertently assumed that moduleDirectories would allow us to specify a NODE_PATH, and for a little while we were accidentally getting away with it.

tjenkinson: Interesting.  This readme says it's meant to just contain directory names, and the doc for the `resolve` package we use says

>opts.moduleDirectory - directory (or directories) in which to recursively look for modules. default: "node_modules"

and [the tests for it are only for names](https://github.com/browserify/resolve/blob/dc8ad1945c663ba47821a269aefb5549f6177cf7/test/module_dir.js).

Given that, I think the fact that it did kind of work with a path was unintentional/undocumented, and therefore IMO erroring would be better and this would be fine as a patch. Especially given if people were misusing `moduleDirectories` when they see the error there will be the correct option available for them to to move too.

@shellscape might have a different opinion though :)

shellscape: sounds reasonable to me

ajhyndman: I have rebased and pushed a change addressing this feedback.  Let me know what you think!

shellscape:
@tjenkinson PR is awaiting your review of most recent changes in response to your comments. please take a look.

tjenkinson (approved):
Lgtm!

packages/node-resolve/test/test.js
tjenkinson: Sorry missed your reply

shellscape:
thanks!

PR #1124, opened 2022-03-02
fix(node-resolve): support `node:` protocol

isker:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

#1120

### Description

Imports with the [`node:` protocol](https://nodejs.org/api/esm.html#node-imports) are currently undetected by this plugin.

Switch to using [is-builtin-module](https://github.com/sindresorhus/is-builtin-module), which handles this protocol and also submodule imports like `fs/promises`.  Functionality is otherwise identical.

I confirmed the added test fails without the other changes in this PR.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -99,18 +99,6 @@ importers:
       rollup: 2.67.3
       typescript: 4.1.2
 
-  packages/auto-install/test/fixtures/pnpm:
-    specifiers:
-      node-noop: ^1.0.0
-    dependencies:
-      node-noop: 1.0.0
-
-  packages/auto-install/test/fixtures/pnpm-bare:
-    specifiers:
-      node-noop: ^1.0.0
-    dependencies:
-      node-noop: 1.0.0
-
   packages/babel:
     specifiers:
       '@babel/core': ^7.10.5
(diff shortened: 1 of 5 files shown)


pnpm-lock.yaml
```diff
@@ -99,18 +99,6 @@ importers:
...
-      node-noop: ^1.0.0
-    dependencies:
-      node-noop: 1.0.0
-
-  packages/auto-install/test/fixtures/pnpm-bare:
-    specifiers:
-      node-noop: ^1.0.0
-    dependencies:
-      node-noop: 1.0.0
-
```
isker, line 113: I have no idea what this is about.  `pnpm install` removed these lines.  I'm happy to manually revert.

shellscape: as long she builds, we're good

isker:
Apologies, but I changed the authorship on the commit and now CI needs a button press again.

shellscape:
@guybedford I think this could use your input given that it's introducing resolution for a new (to rollup) protocol

guybedford (approved):
Seems good. Note we should really just cut out the dependency and use something like:

```js
function isBuiltinModule (name) {
  if (name.startsWith('node:')) name = name.slice(5);
  return require('module').builtinModules.includes(name);
}
```

shellscape:
@isker what are your thoughts on that suggestion?

isker:
What are @lukastaegert’s thoughts?  This solution was his suggestion: https://github.com/rollup/plugins/issues/1120#issuecomment-1052115824.

shellscape:
I took a look at the `is-builtin-modules` and `builtin-modules` deps (both sindre packages, both very small) and I don't see that there'd be a significant perf hit loading either in the global space. For now, I think we're safe to merge as-is. If we notice a performance issue or we get reports of per issues, an optimization would be to dynamically require/import that package in the future.

PR #1132, opened 2022-03-07
feat(wasm): Add target environment configuration

### LINKED ISSUE #1129: [@rollup/plugin-wasm] allow omitting node.js specific modules from output with targeted environment
- Rollup Plugin Name: @rollup/plugin-wasm
- Rollup Plugin Version: v5.1.2 (latest)

### Feature Use Case

I am using rollup and the rollup wasm plugin to create libraries that leverage wasm.

When a downstream developer imports this library using default webpack config, the build will fail due to helper module requiring Node.js modules:

https://github.com/rollup/plugins/blob/4c5e8fd582d7f2e173bfacd7c8cb066af4edd916/packages/wasm/src/helper.ts#L20-L21

webpack will error with "Error: Can't resolve 'fs' in ..."

### Feature Proposal

I'm proposing that the rollup wasm plugin exposes an additional configuration value where the library author can dictate if node.js specific modules can be used to initialize the wasm. This way the library author can use package.json conditional exports to properly expose wasm initialization depending on the environment.

```js
import { wasm } from "@rollup/plugin-wasm";
export default {
  plugins: [
    wasm({
      targetEnv: "browser" | "node" | "auto"
    })
  ]
}
```

- "auto" is the default and doesn't change the behavior
- "browser" removes the `isNode` conditional and associated branches
- "node" can assume the `isNode` conditional is true and remove `fetch`

nickbabcock:
## Rollup Plugin Name: `wasm`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

Closes #1129

### Description

Currently the Wasm rollup plugin emits code that includes calls to
require builtin node modules (`fs` and `path`) as well as builtin
`fetch` in browser environments.

Even though there is an check before the environment specific code is
executed, the mere presence of `require` for builtin node modules will
cause downstream bundlers like `esbuild` and `webpack` to emit errors
when they statically analyze the output.

This commit adds in a `targetEnv` that configures what code is emitted
to instantiate the Wasm (both inline and separate)

  - `"auto"` will determine the environment at runtime and invoke the correct methods accordingly
  - `"auto-inline"` always inlines the Wasm and will decode it according to the environment
  - `"browser"` omits emitting code that requires node.js builtin modules that may play havoc on downstream bundlers
  - `"node"` omits emitting code that requires `fetch`

"auto" is the default behavior and preserves backwards compatibility.

### DIFF
--- a/packages/wasm/README.md
+++ b/packages/wasm/README.md
@@ -69,6 +69,18 @@ Default: (empty string)
 
 A string which will be added in front of filenames when they are not inlined but are copied.
 
+### `targetEnv`
+
+Type: `"auto" | "browser" | "node"`<br>
+Default: `"auto"`
+
+Configures what code is emitted to instantiate the Wasm (both inline and separate):
+
+- `"auto"` will determine the environment at runtime and invoke the correct methods accordingly
+- `"auto-inline"` always inlines the Wasm and will decode it according to the environment
+- `"browser"` omits emitting code that requires node.js builtin modules that may play havoc on downstream bundlers
+- `"node"` omits emitting code that requires `fetch`
+
 ## WebAssembly Example
 
 Given the following simple C file:
--- a/packages/wasm/src/helper.ts
+++ b/packages/wasm/src/helper.ts
@@ -1,6 +1,98 @@
+import { TargetEnv } from '../types';
+
 export const HELPERS_ID = '\0wasmHelpers.js';
 
-export const getHelpersModule = () => `
+const nodeFilePath = `
+var fs = require("fs")
+var path = require("path")
+
+return new Promise((resolve, reject) => {
+  fs.readFile(path.resolve(__dirname, filepath), (error, buffer) => {
+    if (error != null) {
+      reject(error)
+    }
+
+    resolve(_instantiateOrCompile(buffer, imports, false))
+  });
+});
+`;
+
+const nodeDecode = `
+buf = Buffer.from(src, 'base64')
+`;
+
+const browserFilePath = `
+return _instantiateOrCompile(fetch(filepath), imports, true);
+`
+
+const browserDecode = `
+var raw = globalThis.atob(src)
+var rawLength = raw.length
+buf = new Uint8Array(new ArrayBuffer(rawLength))
+for(var i = 0; i < rawLength; i++) {
+   buf[i] = raw.charCodeAt(i)
+}
+`
+
+const autoModule = `
+var buf = null
+var isNode = typeof process !== 'undefined' && process.versions != null && process.versions.node != null
+
+if (filepath && isNode) {
+  ${nodeFilePath}
+} else if (filepath) {
+  ${browserFilePath}
+}
+
+if (isNode) {
+  ${nodeDecode}
+} else {
+  ${browserDecode}
+}
+`
+
+const nodeModule = `
+var buf = null
+if (filepath) {
+  ${nodeFilePath}
+}
+
+${nodeDecode}
+`
+
+const browserModule = `
+var buf = null
+if (filepath) {
+  ${browserFilePath}
+}
+
+${browserDecode}
+`
+
+const autoInlineModule = `
+var buf = null
+var isNode = typeof process !== 'undefined' && process.versions != null && process.versions.node != null
+if (isNode) {
+  ${nodeDecode}
+} else {
+  ${browserDecode}
+}
+`
+
+const envModule = (env: TargetEnv) => {
+  switch (env) {
+    case "auto":
+      return autoModule;
+    case "auto-inline":
+      return autoInlineModule;
+    case "browser":
+      return browserModule;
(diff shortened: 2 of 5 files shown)


nickbabcock:
This PR is marked as draft as it lacks test and it wasn't immediately obvious how other part of rollup tests similar functionality (ie: testing emitting code). 

Not sure if @shellscape have a recommendation of a plugin or test case that I should take inspiration from.

nickbabcock:
Apologies for flipping this PR back to draft and then marking it ready for review. I decided to add in the `auto-inline` target environment that will always inline the Wasm as base64, and will omit both `require` statements and `fetch`.

Using `auto-inline` is becoming my preferred way to author libraries that encompass Wasm algorithms as it allows me to export both a fat and slim distribution. The fat one will have base64 and won't require any setup, while the slim one exposes the Wasm module as a package export and relies on the downstream developer to configure how best to bundle wasm.

shellscape:
Thanks for opening this PR and no worries about the state changes. We're woefully short on WASM knowledge on the Rollup team and I'm afraid there's no active maintainer for that plugin. Would it be possible to ping a few of your colleagues that might work with WASM to weigh in on the changes?

nickbabcock:
I've reached out to a couple contacts to see if they have bandwidth to review. And feel free to ping your own. In the meantime, [I wrote an article](https://nickb.dev/blog/recommendations-when-publishing-a-wasm-library) that mentioned this new feature to make it easier to publish Wasm libraries. The relevant section is abbreviated below.

> The Wasm plugin for rollup has a serious flaw. It will emit code that requires builtin Node.js modules:
> 
> ```ts
> var fs = require("fs") 
> var path = require("path") 
> ```
> 
> These statements will cause downstream bundlers (ie: webpack) to fail to package a browser application as they won't be able to tell if the modules are unused and thus can be eliminated.
> 
> Furthermore, the `fs` and `path` modules were included by the rollup plugin to allow downstream developers to pass an argument to the Wasm initialization in order to read Wasm from disk or fetch it, which is a use case that is rendered unnecessary when the library is structured such that it is known at build time if either the Wasm is inlined or initialized externally using the `slim` package entrypoint.
> 
> So [I created a PR](https://github.com/rollup/plugins/pull/1132) for the plugin that will allow the following situations:
> 
> - When using the slim package, we don't have to worry about the Wasm at all as the downstream developer will be handling that.
> - If we're targeting a node environment, copy the Wasm to a separate file and omit browser specific code (like `fetch`) from the Wasm helper module. 
> - Else we're inlining the Wasm as base64, so decode it using environment specific code (which doesn't use Node.js builtin modules). I called this method `auto-inline`, as it AUTOmatically decodes INLINE Wasm based on if we're in a node environment or not (since it should still be possible to run the UMD or any other importable build with Wasm inlined in node).
> 
> 
> ```js
> import { wasm } from "@nickbabcock/plugin-wasm";
> export {
>   plugins: [
>     env != "slim" &&
>       wasm(
>         env == "node"
>           ? { maxFileSize: 0, targetEnv: "node" }
>           : { targetEnv: "auto-inline" }
>       ),
>   ]
>   // ...
> }
> ```

shellscape (approved):

shellscape:
I'm baffled as to why, but one of the tests is failing in CI after we merged this. @nickbabcock please take a look when you're able to. https://github.com/rollup/plugins/runs/6232657474?check_suite_focus=true We'll leave the changes on master for now, but we won't be able to publish an update until this is resolved.

nickbabcock:
- I guess my contacts are a bit swamped to look into this. Sorry about that.
- Yes, I'm happy to keep a watchful eye over the wasm plugin
- Have you been able to reproduce the CI error? Seems like it only occurs on node 14. Maybe bump the release action to use Node 16?

shellscape:
Well we support v14 until it's out of LTS, so that isnt an option unfortunately.

nickbabcock:
Apologies, I'm not too well versed in how this repo's github actions are plumbed, but my suggestion was along the lines that if the [validate action ensures compatibility node 12-16](https://github.com/rollup/plugins/blob/50b618bcae9f58a33445b95c4b24768d2dbdfefa/.github/workflows/validate.yml#L20), then I'm not aware what makes the release action different and why it needs to use only node 14. I'm also not familiar with pnpm testing syntax, so I'll poke at it, but the answer isn't apparent as of right now.

shellscape:
v14 was used as a baseline for speed, and we (in the Node realm) rarely ran into conflicts between 12 and 14. v12 actually goes out of LTS on Sunday, so it's not really a concern. (We'll end up dropping it shortly). What the CI run did reveal is that the changes have a bit of an issue on Node v14 which we'll have to support for another year. My best advice would be to install Node v14 with NVM or the like, and see if you can reproduce the failing test locally.

PR #1134, opened 2022-03-11
fix(typescript): Drive letter casing on win32 platforms

lovettchris:
## Rollup Plugin Name: `plugin-typescript`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

As in several other places in this code base the regex for matching windows drive letters should be `/^([a-zA-Z]+):/` but one case was missing in the implementation of TSCache.

### DIFF
--- a/packages/typescript/src/tscache.ts
+++ b/packages/typescript/src/tscache.ts
@@ -16,7 +16,7 @@ export default class TSCache {
 
   /** Returns the path to the cached file */
   cachedFilename(fileName: string): string {
-    return path.join(this._cacheFolder, fileName.replace(/^([A-Z]+):/, '$1'));
+    return path.join(this._cacheFolder, fileName.replace(/^([a-zA-Z]+):/, '$1'));
   }
 
   /** Emits a file in the cache folder */


shellscape:
Thanks for the PR. I see you checked the box for tests included, but don't see any tests in your PR. (Nearly) All fixes and features require an accompanying test to be accepted, even if the change looks obvious. It's required to protect against regression.

lovettchris:
@shellscape, the existing tests failed on my machine and with my fixes they pass.  The problem is in figuring out how to automate the machine setup.  One needs a windows drive installed in such way that node.js `fs.realpath` returns a different case from the `path.resolve` function and I haven't figured out how or why my machine is doing that.  I supposed one could implement some kind of fs hook during testing that randomizes file name case on win32 platforms... and this might catch more bugs...

lovettchris:
Can you suggest the right title for this PR, I've tried a million variations they all failed...

shellscape:
_(Title taken care of. It didn't like the hyphen in the scope)_

OK Thanks for the explanation on the test. That's pretty much all we need to archive to let that slide. Anyone wondering why it has no test can point back to your comment. WRT drives on Windows, it's not exactly easy to create new mounts. Your best bet would be creating a virtual drive/ramdisk, but I went through that process for https://github.com/shellscape/webpack-plugin-ramdisk#windows-users and came up short. I'm not aware of any new capabilities in the newer versions of Windows either. You're welcome to try, but we'll give it a pass if you'd rather not.

lovettchris:
> didn't like the hyphen in the scope

thanks for fixing the title.  I will pass on figuring out how to test this for now since I'm just trying to unblock another project that is consuming all my time, but I could add a new work item in case someone else has time to test properly all windows file system case insensitivities...

lovettchris:
@shellscape it might be relatively easy to reproduce the bug on any windows machine.  Create a shortcut that launches `cmd.exe` with the `Start in` path set to the opposite of what you get from nodejs `fs.realpath` in that folder.  In my case the opposite is lowercase, so this: 
image

Then try and run the rollup tests with `npm run ci:test:only` and they will all fails like this:

image

shellscape (approved):

lovettchris:
the existing tests failed on my machine and with my fixes they pass.  The problem is in figuring out how to automate the machine setup.  One needs a windows drive installed in such way that node.js "fs.realpath" returns a different case from the path.resolve function and I haven't figured out how or why my machine is doing that.

I supposed one could implement some kind of fs hook during testing that randomizes file name case on win32 platforms... and this might catch more bugs...
________________________________

PR #1140, opened 2022-03-16
fix(eslint): include and exclude options typing allow to be a regex

tinymins:

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

ESLint plugin, `exclude` and `include` got a wrong type define. See the screenshot for details, we must cast regex to string for omit the error.

image

Actually, those options will pass to [`createFilter` function](https://github.com/rollup/plugins/blob/5363f55aa1933b6c650832b08d6a54cb9ea64539/packages/eslint/src/index.ts#L37), so its typing should follow this function's parameters typing.

https://github.com/rollup/plugins/blob/5363f55aa1933b6c650832b08d6a54cb9ea64539/packages/eslint/src/index.ts#L37

### DIFF
--- a/packages/eslint/types/index.d.ts
+++ b/packages/eslint/types/index.d.ts
@@ -1,5 +1,6 @@
 import { Plugin } from 'rollup';
 import { CLIEngine } from 'eslint';
+import { CreateFilter } from '@rollup/pluginutils';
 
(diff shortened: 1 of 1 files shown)


shellscape:
thanks!

PR #1144, opened 2022-03-18
chore(url): update pluginutils deps, extend tests

r0b1n:
## Rollup Plugin Name: `url`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] yes (_Included test that requires update of dependencies_)
- [ ] no

Breaking Changes?

- [x] yes (_Dependencies update, major version bump_)
- [ ] no

### Description

Fixes issue with `url` module by bumping `@rollup/pluginutils` dependencies to v4. Tests that reproduce the issue on v3 are included. The issue is related to incorrect [handling of absolute pathes](https://github.com/rollup/plugins/blob/master/packages/pluginutils/CHANGELOG.md#v400) in `@rollup/pluginutils` v3. 

This is a major dependencies bump, so the version of this module is bumped as well to `7.0.0`.


shellscape:
Looks like there's a failure in CI on Windows from your changes, please have a look.

r0b1n:
@shellscape thanks for running the CI checks. I found an issue. It is an oversight in `pluginutils@4.2.0`. I added a workaround.

[Here](https://github.com/rollup/plugins/blob/master/packages/pluginutils/src/createFilter.ts#L10-L24) if `id` is an absolute path, it doesn't do anything to it. That is I think not what should happen, is should still normalize paths regardless. That is also true for paths starting with `*`, it should normalize path starting with `*` because those may still contain windows style `\﻿﻿` slashes inside. It still should not append resolution base of course, but normalize.

shellscape:
We got #1161 published yesterday so you should be good to update for that fix. Nice work finding the issue.

shellscape (approved):

PR #1148, opened 2022-03-22
fix(node-resolve): update side effects logic to be deep when glob doesn’t contain `/`

TheMcMurder:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

- https://github.com/rollup/rollup/issues/4084

### Description

node-resolve doesn't include files from nested folders even when they match files in `sideEffects`.

Example:

```
root-dir
- index.js
- deep-dir
---- other-file.js
```
```json
{
  "main": "index.js",
  "sideEffects": ["*.js"]
}
```

the side-effects of `other-file.js` will not be included because `picomatch` only matches items in the root dir unless we change the `sideEffects` block to include wildcards.

This should fix [this issue](https://github.com/rollup/rollup/issues/4084) by doing a deep search when the glob does not contain a `/`. This matches the same behaviour as webpack.


TheMcMurder:
> 👋 I’m wondering if this is this something we should do or should packages themselves already be including `**/` if they also want to match children?

I've debated this as well. The disconnect for me which lead me down this path is that, in my testing, this behavior works this way with `webpack` and `parceljs` without adding `**/` to `sideEffects`.

I'll be honest and say I'm not sure if that alone is a good reason to change it. I couldn't find the official spec for this so I wasn't sure if webpack/parcel was doing it incorrectly or this plugin was doing it incorrectly. 

I chose to start here because I use rollup more out of preference and am (slightly) more familiar with the codebase.

TheMcMurder:
I'm okay with making it the default behavior over time, I wanted to make sure it was a non-breaking change when introduced.

> One concern I have with the flag at the moment is that it applies to everything, so you could end up with an annoying problem where one dependency needs it to be true to work and another only works with false. Maybe the chance of that is so unlikely it’s not worth worrying about though?

I hadn't even thought about that scenario. How could we address something like that? I don't want to add _even more_ configuration but my first idea goes to another configuration item:

```js
// Brainstorming
// assuming it's default behavior in the future

  const bundle = await rollup({
    input: 'deep-side-effects/index.js',
    onwarn: failOnWarn(t),
    plugins: [
      nodeResolve({
        rootDir: 'deep-side-effects',
        ignoreDeepEffectsForPackages: [/* list of packages to ignore the deep effects */]
      })
    ]
  });
```

TheMcMurder:
Whoops, good call. I added a brief description to the readme. Happy to edit the description or include an example if you think that would be beneficial.

TheMcMurder:
I'm not sure. I'll experiment with how webpack handles it and use that to inform if we should add the `**/` suffix. It may take me a few days to experiment with that though. In the meantime I'm going to push a new commit to fix the prettier issues.

TheMcMurder:
@tjenkinson fantastic! I updated the MR.
- removed new option and associated docs
- updated tests

packages/node-resolve/src/util.js
```diff
@@ -144,7 +144,18 @@ export function getPackageInfo(options) {
     if (typeof packageSideEffects === 'boolean') {
       internalPackageInfo.hasModuleSideEffects = () => packageSideEffects;
     } else if (Array.isArray(packageSideEffects)) {
-      internalPackageInfo.hasModuleSideEffects = createFilter(packageSideEffects, null, {
+      let finalPackageSideEffects = packageSideEffects;
+      finalPackageSideEffects = packageSideEffects.map((sideEffect) => {
```
tjenkinson, line 148: Please could you make this `const final…` and initialise the variable here, then it could be one line not 2

tjenkinson (commented):
Looks great. One comment

tjenkinson (approved):
Lgtm!

shellscape:
thanks!

Issue #1150, opened 2022-03-26
[@rollup/plugin-node-resolve] Extend resolveOnly property to allow for denylists / miscellaneous use cases. 


TotalTechGeek:
- Rollup Plugin Name:  node-resolve
- Rollup Plugin Version: 13.1.3

### Feature Use Case

Currently, it appears that the `resolveOnly` property allows you to pass in an array of strings or regular expressions to validate if a plugin should be bundled in. 

I had a use case where I wanted to bundle in everything except for certain dependencies. (`node-rdkafka`, `sqlite3`). 

I would like to extend the `resolveOnly` property to allow developers more control. 

This feature may already exist under another property, and if so, I apologize for the unnecessary recommendation. 

### Feature Proposal

While it is semi-possible to work around this by creating a complex regular expression to attempt to match everything except the given string, there are a few ways to introduce this. 

**Suggestion 1**
I believe the best way to introduce this is to also allow `resolveOnly` to be a `(plugin: string) => boolean` function, thus allowing a developer to write something like:

```js 
function forbid (...plugins) {
    const set = new Set(plugins)
    return plugin => !set.has(plugins)
}

nodeResolve({
    resolveOnly: forbid('sqlite3', 'node-rdkafka')
})
```

By allowing the user to provide their own function, it gives them complete control to customize however they see fit.

**Suggestion 2**
While I believe suggestion 1 should be simple enough to implement, if you wanted to change just literally 1 type / 1 line of code, all you'd need to do is change out `RegExp` for `{ test: (plugin: string) => boolean }`. I do not like this as much :) 

```js
function forbid (...plugins) {
    const set = new Set(plugins)
    return [{ test: plugin => !set.has(plugins) }]
}

nodeResolve({
    resolveOnly: forbid('sqlite3', 'node-rdkafka')
})
```

**My current workaround**

In order to work around this, I ended up doing the following:

```js
function forbid (...plugins) {
    const regex = new RegExp('');
    const set = new Set(plugins)
    regex.test = (pattern) => !set.has(pattern)
    return [regex]
}
```

While this works for now, I know it's a hack and could break at any minute :P 

If someone likes the idea, I'm 100% willing to file the MR!

shellscape:
tagging a few folks that might have some input @guybedford @tjenkinson @lukastaegert

tjenkinson:
Supporting a function too that returns a boolean sounds good to me

TotalTechGeek:
Closing the issue as the pull request has been merged :)

PR #1152, opened 2022-03-27
feat(node-resolve): Add the ability to pass a function into resolveOnly

TotalTechGeek:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

#1150 
### Description

Currently, the `resolveOnly` property allows you to pass in an array of strings or regular expressions to validate if a module should be bundled in.

I had a use case where I wanted to bundle in everything except for certain dependencies. (node-rdkafka, sqlite3).

I've extended the resolveOnly property to allow developers more control. With this PR, it should be possible to pass in a function into `nodeResolve` to override the check that takes place, thus allowing something like the following:

```js
function forbid (...modules) {
    const set = new Set(modules)
    return module => !set.has(module)
}

nodeResolve({
    resolveOnly: forbid('sqlite3', 'node-rdkafka')
})
```


packages/node-resolve/src/index.js
```diff
@@ -112,11 +121,8 @@ export function nodeResolve(opts = {}) {
       isRelativeImport = true;
     }
 
-    if (
-      !isRelativeImport &&
-      resolveOnly.length &&
```
tjenkinson, line 117: Just checking have we definitely kept the behaviour where the array is empty the same?

I think before if it was empty we would resolve everything (which imo seems backwards but we should keep it the same).

tjenkinson (approved):
Lgtm! I think the empty array behaviour is the same but just wanted to double check

packages/node-resolve/src/index.js
```diff
@@ -63,13 +63,22 @@ export function nodeResolve(opts = {}) {
...
+  // creates a function from the patterns to test if a particular module should be bundled.
+  const allowPatterns = (patterns) => {
+    const regexPatterns = patterns.map((pattern) => {
+      if (pattern instanceof RegExp) {
+        return pattern;
+      }
+      const normalized = pattern.replace(/[\\^$*+?.()|[\]{}]/g, '\\$&');
+      return new RegExp(`^${normalized}$`);
+    });
+    return (id) => !regexPatterns.length || regexPatterns.some((pattern) => pattern.test(id));
```
TotalTechGeek, line 75: It is :) 

Removing the `!regexPatterns.length` causes tests to fail. I thought it was a bit backwards at first too, but I think it makes sense in the context of "zero elements in the array means there are no filters, and no filters means include everything"

tjenkinson: Cool. Hmm I see it as zero items in an allow list meaning there’s nothing to allow 🤷 

also follows the semantics of `[].some(…)` then too

shellscape:
thanks!

PR #1161, opened 2022-04-12
fix(pluginutils): fix path normalisation for createFilter

r0b1n:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

Discovered in https://github.com/rollup/plugins/pull/1144

### Description

`createFilter` function is not normalizing paths correctly. Everything is expected to be converted to `/` while on windows it may not be the case as it depends on users input.

### DIFF
--- a/packages/pluginutils/src/createFilter.ts
+++ b/packages/pluginutils/src/createFilter.ts
@@ -9,7 +9,7 @@ import normalizePath from './normalizePath';
 
 function getMatcherString(id: string, resolutionBase: string | false | null | undefined) {
   if (resolutionBase === false || isAbsolute(id) || id.startsWith('*')) {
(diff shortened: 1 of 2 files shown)


packages/pluginutils/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rollup/pluginutils",
-  "version": "4.2.0",
+  "version": "4.3.0",
```
shellscape, line 3: please revert this change. our automatic publish process handles version bumps.

shellscape (changes requested):

PR #1040, opened 2021-11-11
fix(typescript): add error location to warnings

YonatanKra:
## Rollup Plugin Name: `{name}`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] yes (_bugfixes and features will not be merged without tests_)
- [ ] no

Breaking Changes?

- [ ] yes (_breaking changes will not be merged unless absolutely necessary_)
- [x] no

If yes, then include "BREAKING CHANGES:" in the first commit message body, followed by a description of what is breaking. 

List any relevant issue numbers: closes #1041 

### Description

When running the plugin and running into a typescript error, only the error message appears without the location of the error. This PR adds the location of the error to the logged message so the developer would be able to easily fix the typescript error.

PR #1167, opened 2022-04-22
fix(typescript): add precise file/line/column in warning message

zarov:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

I reimplemented #1040 as it seems abandoned. I fixed linting errors and modified tests that were missing from the initial PR.

### DIFF
--- a/packages/typescript/src/diagnostics/toWarning.ts
+++ b/packages/typescript/src/diagnostics/toWarning.ts
@@ -21,12 +21,15 @@ export default function diagnosticToWarning(
   if (diagnostic.file) {
     // Add information about the file location
     const { line, character } = diagnostic.file.getLineAndCharacterOfPosition(diagnostic.start!);
+    const { fileName } = diagnostic.file;
 
     warning.loc = {
       column: character + 1,
       line: line + 1,
-      file: diagnostic.file.fileName
+      file: fileName
     };
+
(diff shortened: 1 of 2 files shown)


NotWoods (approved):
LGTM, although I expected Rollup to handle this since we provide line/column info to it

zarov:
That is what I thought at first, maybe something wasn't correctly used. After looking at rollup (which I didn't in the first place), `loc` seems [supported](https://github.com/rollup/rollup/blob/030c56fd6b186a0ddfd57d143ad703f34fd2c17a/src/utils/options/normalizeInputOptions.ts#L84). After some logging in this part though, `loc` seems to have disappeared from `warning` right where I linked it.

I don't have a lot of knowledge of the rollup codebase and its plugins, so I haven't found anything yet. I'll take another look as soon as I can because I am convinced that we can make this thing work correctly without the tricky message edition.

zarov:
So after having another look at it, I think this PR should not be, as the output of `onwarn` needs to be handled by the user, as seen [in the documentation](https://rollupjs.org/guide/en/#onwarn). A simple `onwarn(warning) { console.warn(warning); }` solves it for the end-user.

PR #1179, opened 2022-04-29
fix(wasm): Reference atob polyfill in browser wasm test

nickbabcock:
The "target environment browser" test invoked the bundled output, which
leads to an undefined error on node 14 as [`atob` was introduced in node
16][0].

This commit uses the same atob polyfill that was introduced in #334

[0]: https://developer.mozilla.org/en-US/docs/Web/API/atob#browser_compatibility

## Rollup Plugin Name: `wasm`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

Addresses the release github action breakage as reported in #1132

### DIFF
--- a/packages/wasm/test/fixtures/async.js
+++ b/packages/wasm/test/fixtures/async.js
@@ -1,6 +1,9 @@
 import fixture from './sample.wasm';
 
 // eslint-disable-next-line no-undef
+globalThis.atob = (str) => Buffer.from(str, 'base64').toString('binary');
+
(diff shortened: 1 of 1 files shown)


shellscape:
Thanks for jumping on this. Will review as soon as my kids are down for.bed tonight.

shellscape (approved):

PR #1200, opened 2022-06-07
chore: Set permissions for GitHub actions

neilnaveen:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: neilnaveen <42328488+neilnaveen@users.noreply.github.com>

### DIFF
--- a/.github/workflows/node-windows.yml
+++ b/.github/workflows/node-windows.yml
@@ -11,6 +11,9 @@ on:
       - '*'
       - '!master'
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: windows-2019
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,9 +5,14 @@ on:
     branches:
       - master
 
+permissions:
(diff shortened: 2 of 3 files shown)


shellscape:
I noticed that nearly all of your GitHub contribs have been identical PRs to this one to 30+ various projects. That leads me to wonder about motivation. We'll need to research these changes before considering the PR.

shellscape:
Seems the author's M.O. is to drop a copy of this PR into various repos, and not follow up. Closing for now.

PR #1212, opened 2022-06-19
feat(node-resolve): dedupe preferBuiltins protocol

bluwy:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

When bundling a library that uses both `node:fs` and `fs`, it can be deduped as a single `node:fs` or `fs` import instead, slightly reducing the bundle size. This can often happen when using third-party libraries.

To enable this optimization, `preferBuiltins` now also accepts `prefer-protocol` or `prefer-no-protocol` to dedupe to either `node:fs` or `fs`, depending on the users node target version that supports it.

### DIFF
--- a/packages/node-resolve/README.md
+++ b/packages/node-resolve/README.md
@@ -126,10 +126,12 @@ Specifies the properties to scan within a `package.json`, used to determine the 
 
 ### `preferBuiltins`
 
-Type: `Boolean`<br>
+Type: `Boolean | 'prefer-protocol' | 'prefer-no-protocol'`<br>
 Default: `true` (with warnings if a builtin module is used over a local version. Set to `true` to disable warning.)
 
 If `true`, the plugin will prefer built-in modules (e.g. `fs`, `path`). If `false`, the plugin will look for locally installed modules of the same name.
+
+When set to `true`, `prefer-protocol` and `prefer-no-protocol` can also be used to deduplicate, e.g. `fs` and `node:fs`, as the same module. The final import path would be `node:fs` for `prefer-protocol` or `fs` for `prefer-no-protocol`.
 
 ### `modulesOnly`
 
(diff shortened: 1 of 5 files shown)


bluwy:
> I’m wondering if this really should be part of this plug-in though or a separate plug-in?

I thought about it too, but if it's a separate plugin, the only place it'll be useful are in projects that externalize all deps and builtin modules (bundle source code only). But in that case, the source code would've been written with only `node:` or not in the first place, not needing the plugin.

Deduping protocols would mostly be handy when bundling third-party packages which `node-resolve` handles, so it may be nice to have this as a first-class feature.

> Also I reckon one of the reasons for the new prefix was to prevent name collisions between node built-ins and real packages, so I’m not sure we should provide an option to go in the direction from prefix to no prefix?

Good point. It would a problem if the project/package used `node:url` and the `url` at the same time, but that sounds like an edge case, and that if they have a configuration like this, they should be avoiding both `prefer-no-protocol` and `prefer-protocol` (as both is destructive). But for most projects I think it would be a handy addition.

bluwy:
> A bit related, we should probably always have this plug-in mark `node:*` as external, regardless of the `preferBuiltins` option

It looks like it's already handled as `node:*` resolves to `null` when calling `resolveImportSpecifiers`, so it won't match a local package that would trigger a warning.

shellscape:
> I’m still not sure it needs to be in this plug-in vs another one. I get that the prefix is a node thing which is why it could fall in here, but also it could easily be an isolated plug-in.

I agree with this

bluwy:
I'll make this as a separate plugin then, thanks for the feedback!

shellscape:
Thanks. Please make sure to add it to `rollup/awesome` so others can discover it

PR #1237, opened 2022-08-09
feat(multi-entry): add preserveModules option

814k31:
## Rollup Plugin Name: `@rollup/plugin-multi-entry`

This PR contains:
Probably somewhere in the grey area of a bugfix and a feature
- [x] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

Its "kind of" a breaking change? Breaking in the fact that people who use this plugin with preserveModules=true will experience a change by upgrading, however the change will be probably an outcome that they would expect from this plugin when using preverveModules=true

- [ ] no

### Description

When preserveModules is set on the output, its expected that the filenames and structure will not be altered.
This library is always setting the files names to be `multi-entry.js` and if there are multiple it is `multi-entry1.js` and so on.

My desired out come is that I can use the multi-entry plugin to target all src files within my project to be built so when using the library I can easily do
`import { ExampleComponent } from 'example-library/nested-directory/exampleComponent';`

when the src file structure is:
example-library -> src -> nested-directory -> exampleComponent.js


814k31:
Because I dont have pnp on my machine, I haven't actually been able to run the tests unfortunately... just fyi but that should at least describe my expectation

packages/multi-entry/src/index.js
```diff
@@ -53,7 +53,7 @@ export default function multiEntry(conf = {}) {
     outputOptions(options) {
       return {
         ...options,
-        entryFileNames: config.entryFileName
+        entryFileNames: (options.preserveModules && config.entryFileName === DEFAULT_OUTPUT) ? options.entryFileName : config.entryFileName
```
814k31, line 56: I dont know if you would prefer this to actually fall back to `options.entryFileNames` or `options.entryFileName`
So raising here to highlight the potential change

```suggestion
        entryFileNames: (options.preserveModules && config.entryFileName === DEFAULT_OUTPUT) ? options.entryFileNames : config.entryFileName
```

814k31:
I am open to other suggestions on how to solve this, one being allowing config.entryFileName to be set to false or null to fallback to the options which would go into making this a non breaking change. However I wanted to try provide a solution that doesnt bloat configuration options

packages/multi-entry/src/index.js
814k31: using entryFileName had a warning so filenames is the go

814k31:
@shellscape not to be a pain but could I bug you to have a look at this
It's actually stopping my enterprise from adopting rollup in general

I'm very flexible on the implementation here

Really I just need a method to make sure the plugin doesn't always use the default entry filename so that I can use rollups default behaviour with preserveModules and this plugin together

shellscape:
I generally don't have much stake in this plugin. Please check out the history and which maintainers have contributed to it in the past for their feedback.

814k31:
@shellscape will do thanks for replying so quickly

814k31:
Hello @guybedford @NotWoods

sorry to drag you guys in here randomly
Could I ask you for some reviews?

As stated before I'm pretty flexible on implementation here

shellscape:
Since the other two maintainers are quiet at the moment, I'll chime in to try and get this moving forward. I generally don't like to merge breaking changes unless there's a large, community-wide reason to do so. That said, I think you could add another option such as `preserveOutputFilenames` so that when true, the same behavior is applied without a breaking change. I would have no issue merging that for you.

814k31:
Thanks @shellscape I'll update the pr when I have time and tag you when its done

814k31:
@shellscape all updated to use a new config option instead

I've also updated the pr title to hopefully match the required format, though tbh I'm not totally aware of what the required format is...

either way, look forward to hearing from you :)

shellscape:
The change looks good now. We're going to be blocked from merging until #1270 lands (sorry about that). One thing to do in the mean time is to update the README with the new option.

814k31:
Sorry about the delay myself, README is updated :)

814k31:
Hey @shellscape

I noticed there were some pr conflicts so I've gone through and fixed them up
is it possible to get this merged in ASAP? It's now become a priority for my work as we are going forward with adopting rollup

Thanks!

shellscape (approved):

shellscape:
thanks!

PR #1242, opened 2022-08-14
feat(typescript): allow override of forced `noEmit` and `emitDeclarationOnly` compiler options

Harris-Miller:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

This MR is a replacement for the closed https://github.com/rollup/plugins/pull/1206. the suggested feature of a `transpilerMode` option, which unfortunately was discovered to be a breaking change.

**Why**

Currently, this plugin forced the `compilerOptions` `noEmit` and `emitDeclarationOnly` to be false. There is a note `// Typescript needs to emit the code for us to work with.`

This is **false**

The way that the plugin is written, the `load` hook looks for if typescript did transpilation for a file and returns that if found, null otherwise. This means for `emitDeclarationOnly` rollup will simply load from the output of the previous plugin or from the file system. This is ideal to mimic the following:
```jsonc
// package.json
{
  "scripts": {
    "build": "babel src --out-dir dist" 
    "typecheck": "tsc --emitDeclarationOnly"
  }
}
```
Where build uses babel to do all transpilation, including typescript via `@babel/preset-typescript`, and `tsc` is used to only emit declaration files.

Currently, rollup can more or less do this by running all code though `@rollup/plugin-typescript` first and then `@rollup/plugin-babel`, however with the forced `emitDeclarationOnly: false` being on, all the code is run through the `typescript` transpiler first. If `@rollup/plugin-babel` is set up to do to `typescript`, there isn't a need to actually do this

By allowing `emitDeclarationOnly: true`, `@rollup/plugin-typescript` is drastically faster! Since it only has to emit declaration file (which are emitted by rollup separately as "asset" files in the generateBundle step). On the large projects I tested this out at my work, I say initial builds drop from 20 seconds down to 2 seconds


NotWoods (approved):
Thanks for the change!

packages/typescript/README.md
```diff
@@ -238,6 +247,7 @@ These compiler options are ignored by Rollup:
 
 - `noEmitHelpers`, `importHelpers`: The `tslib` helper module always must be used.
 - `noEmit`, `emitDeclarationOnly`: Typescript needs to emit code for the plugin to work with.
+  - While this was true for early iterations of `@rollup/plugin-typescript`, it is no longer. To override this behavior, and defer to `tsconfig.json` for these options, see the `noForceEmit` options
```
shellscape, line 250:
```suggestion
- `noEmit`, `emitDeclarationOnly`: Typescript needs to emit code for the plugin to work with.
  - _Note: While this was true for early iterations of `@rollup/plugin-typescript`, it is no longer. To override this behavior, and defer to `tsconfig.json` for these options, see the [`noForceEmit`](#noForceEmit) option_
```

shellscape:
Looks like there are some snapshot mismatches in the tests for this change. Please see the CI output for Validate Monorepo.

packages/typescript/README.md
Harris-Miller: Suggestion applied

Harris-Miller:
I fixed the tests. Let me know if you need anything additional

Harris-Miller:
Anything needed from me to get this merged?

shellscape:
nope :)

PR #1249, opened 2022-08-28
fix: support .mts and .cts out of the box

RebeccaStevens:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

Currently out of the box the `typescript` plugin only supports `.ts` and `.tsx` extensions. This PR adds support for `.mts` and `.cts` as well without the user needing to modify the `includes` option.

Additionally only `.d.ts` were supported as type declaration files, this PR also adds support for `.d.mts` and `.d.cts` files.

### DIFF
--- a/packages/typescript/src/index.ts
+++ b/packages/typescript/src/index.ts
@@ -17,6 +17,9 @@ import { preflight } from './preflight';
 import createWatchProgram, { WatchProgramHelper } from './watchProgram';
 import TSCache from './tscache';
 
+/**
+ * The glob pattern used to test if a given file extension corresponds to a TypeScript file.
+ */
 const defaultTsExtensionsGlob = '?(m|c)ts+(|x)';
 
 export default function typescript(options: RollupTypescriptOptions = {}): Plugin {
--- a/packages/typescript/src/index.ts
+++ b/packages/typescript/src/index.ts
@@ -3,7 +3,7 @@ import * as path from 'path';
 import { createFilter } from '@rollup/pluginutils';
 
 import { Plugin, RollupOptions, SourceDescription } from 'rollup';
-import type { Watch } from 'typescript';
+import type { ResolvedModuleFull, Watch } from 'typescript';
 
 import { RollupTypescriptOptions } from '../types';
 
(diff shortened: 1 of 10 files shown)


packages/typescript/src/index.ts
```diff
@@ -17,6 +17,8 @@ import { preflight } from './preflight';
 import createWatchProgram, { WatchProgramHelper } from './watchProgram';
 import TSCache from './tscache';
 
+const defaultTsExtensionsGlob = '?(m|c)ts+(|x)';
```
NotWoods, line 20: It would be great to have a doc comment here to explain what patterns this glob is matching, so future contributors know :)

packages/typescript/src/index.ts
```diff
@@ -110,7 +116,8 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
       const resolved = resolveModule(importee, containingFile);
 
       if (resolved) {
-        if (resolved.extension === '.d.ts') return null;
+        if (resolved.extension.startsWith('.d.') && resolved.extension.match(/\./g)?.length === 2)
```
NotWoods, line 119: I think we should allow files with multiple `.`s, since `file.with.lots.of.periods.ts` is a valid path

NotWoods (changes requested):

packages/typescript/src/index.ts
RebeccaStevens: added

packages/typescript/src/index.ts
RebeccaStevens: fixed/improved.

perrin4869:
For what it's worth, I think https://github.com/rollup/plugins/pull/1194 already adds support for this?

RebeccaStevens:
Yeah, that's true. It would be good if we can get one of these merged soon as currently libraries using the `.d.mts` extension cannot be used in rollup build environment.

perrin4869:
I think #1194 is more complete and merging this one will only make it harder to resolve conflicts
@NotWoods any chance you can review that PR?
I can resolve the new conflicts but i don't want to keep resolving conflicts indefinitely...

shellscape:
Thanks for the work on this PR. https://github.com/rollup/plugins/pull/1194 supersedes this PR and we've just merged that.

PR #1254, opened 2022-09-03
fix: html template type

flying-sheep:
## Rollup Plugin Name: `html`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

List any relevant issue numbers: N/A

### Description

An object is always passed.


shellscape:
Thanks for the potential catch. That typing was explicit because in rare cases it would be undefined, and the potential exists in the future for it to be undefined. I'd be happy to add a note to the code to indicate as much.

Issue #1267, opened 2022-09-21
browser: true ignored when plugin-node-resolve 14 is used with plugin-typescript


FranklinWhale:
- Rollup Plugin Name: plugin-node-resolve
- Rollup Plugin Version: 14.1.0
- Rollup Version: 2.79.0
- Operating System (or Browser): Windows 10 Professional 21H2 19044.2006
- Node Version: 16.17.0
- Link to reproduction: https://replit.com/@FranklinWhale/NodeResolve14-TypeScript, https://replit.com/@FranklinWhale/NodeResolve13-TypeScript

### Expected Behavior
The `diff` returns no result

### Actual Behavior
In NodeResolve14-TypeScript, TS and JS output are not the same. The outputs are the same in NodeResolve13-TypeScript.

### Additional Information

This is probably caused by the use of `order: "post"` as part of #1245. Both plugin-typescript and [plugin-typescript2 require plugin-node-resolve be run before themselves](https://github.com/ezolenko/rollup-plugin-typescript2#rollupplugin-node-resolve).

pedrotorchio:
This could finally explain the issues I'm having. +1

FranklinWhale:
[rollup-plugin-typescript2](https://github.com/ezolenko/rollup-plugin-typescript2) has just released [a version](https://github.com/ezolenko/rollup-plugin-typescript2/releases/tag/0.34.1) that works properly with plugin-node-resolve 14

dkuhnert:
Same issue here. My workaround is to disable `@rollup/plugin-typescript`'s `resolveId` and add `.ts` to `@rollup/plugin-node-resolve` extensions:

```typescript
import typescriptPlugin from "@rollup/plugin-typescript";
import {nodeResolve as nodeResolvePlugin} from "@rollup/plugin-node-resolve";

const typescript = {
  ...typescriptPlugin(/* ... */),
  // do not resolve anything instead leave this to `@rollup/plugin-node-resolve`!
  // see also: https://github.com/rollup/plugins/issues/1267
  resolveId: undefined,
};
const nodeResolve = nodeResolvePlugin({
  extensions: [".mts", ".ts", ".cts", ".mjs", ".js", ".cjs", ".json", ".node"],
  mainFields: ["browser", "module", "jsnext:main", "main"],
});

// ...
```

PR #1310, opened 2022-10-10
fix(typescript)!: don't resolve filtered files (#1267)

tk-1io:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] no

#1267 

### Description

BREAKING CHANGES: The signature of the error being thrown has changed.

This should fix the issue #1267, which is caused by the use of `order: "post"` as part of https://github.com/rollup/plugins/pull/1245. It will prevent the typescript plugin from resolving files which are not included by its provided configuration/options. Instead the `node-resolve` plugin will again handle these cases, which is able to correctly handle the `browser` config parameter. The same fix has been done for the `rollup-plugin-typescrip2` here: https://github.com/ezolenko/rollup-plugin-typescript2/commit/f6db59613a66f58c48310aa8fa785951970b5d6d

### DIFF
--- a/packages/typescript/src/index.ts
+++ b/packages/typescript/src/index.ts
@@ -125,6 +125,7 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
 
       if (resolved) {
         if (/\.d\.[cm]?ts/.test(resolved.extension)) return null;
+        if (!filter(resolved.resolvedFileName)) return null;
         return path.normalize(resolved.resolvedFileName);
       }
 
--- a/packages/typescript/test/test.js
+++ b/packages/typescript/test/test.js
@@ -999,8 +999,8 @@ test.serial('does it fail for filtering with incorrect rootDir in nested project
       plugins: [typescript({ tsconfig: 'tsconfig.json' })]
     })
   );
-  // Will throw parse error because it includes a typescript file outside CWD
-  t.is(error.code, 'PARSE_ERROR');
+  // It imports a typescript file outside CWD, hence will not get resolved
+  t.is(error.code, 'UNRESOLVED_IMPORT');
 });
 
 test.serial('does manually setting filterRoot resolve nested projects', async (t) => {


shellscape:
This is actually a breaking change, because the error being thrown has changed signature for that scenario. I'd also like to see a test with a fixture containing the `browser` setting so that any possible future regressions guard against edge cases with that as well.

shellscape:
@lukastaegert getting some new CI errors. could they be ralated to your recent work?

tk-1io:
> This is actually a breaking change, because the error being thrown has changed signature for that scenario. I'd also like to see a test with a fixture containing the `browser` setting so that any possible future regressions guard against edge cases with that as well.

Yes, you are correct about the changed error signature (though, `rollup-plugin-typescript2` released the same fix as a patch, I think this is probably because the plugin was never supposed to resolve any filtered paths in the first place).

Regarding the test, I'm not sure how I should cover the `browser` setting in a typescript-plugin test, because typescript does not seem to care about this setting at all when resolving paths (but the `node-resolve` plugin does). This means, even with this fix, the `browser` setting problem can still occur when someone configures the `typescript` plugin or his `tsconfig.json` in a way so it won't exclude imports to other packages (i.e. `node_modules/...`). To completely get rid of the `browser` setting problem, typescript itself would need to be able to handle this setting.
The point here is that, with a typical typescript setup, only the projects `*.(m|c)ts` files are meant to get compiled, while the files of dependencies inside `node_modules` are excluded from this process, hence the `typescript` plugin should also not be the one who is responsible to resolve the paths into these dependencies. Instead the `node-resolve` plugin is here to do this.
So in a nutshell, the problem arose because the typescript plugin tries to resolve a path which it is not supposed to do, hence "stealing" the resolve-job from the `node-resolve` plugin (which would correctly respect the `browser` setting). The problem was not apparent before, because in a typical rollup configuration the plugins are ordered in a way so that the `node-resolve` plugin gets executed before the `typescript` plugin, therefore prevented the `typescript` plugin from executing its flawed resolve behaviour, but this changed due to the mentioned pull https://github.com/rollup/plugins/pull/1245. This was also the reason why I just updated the existing test, as I think it addresses exactly this case: checking that the `typescript` plugin doesn't do any work on files that aren't its business. It now just exits a step earlier (when trying to resolve, instead of afterward trying to parse it).

Regarding the CI problems: I could not run the tests without some changes (e.g. `packages/typescript/ava.config.js`  symlink was broken due to a renamed file), so I too think the recent changes are the ones to blame here.

lukastaegert:
Sorry about the CI issues, if you rebase against latest master and run `pnpm i`, everything should be working again. I was working on a batch update of all plugins that was taking a little longer than I expected and you happened to stumble into the middle of it, sorry for that.

tk-1io:
Thanks Lukas, after the rebase everything worked fine again.

@shellscape Based on my previous comment, do you have any suggestions on how we can proceed with this issue?

shellscape:
Looks like CI cleaned itself up

thanks!

shellscape (approved):
Thanks for being patient on this one. I'd like one more maintainer to put eyes on this before we merge.

NotWoods (approved):

Issue #1230, opened 2022-08-01
[@rollup/plugin-typescript] tsconfig declarationDir is ignored when using single file output


CarrotB:
- Rollup Plugin Name: plugin-typescript
- Rollup Plugin Version: 8.3.4
- Rollup Version: 2.77.2
- Operating System (or Browser): Windows
- Node Version: v14.19.1
- Link to reproduction _(⚠️ read below)_: https://replit.com/@BobCarrot/rollup-plugin-typescript-declarationDir-error

### Expected Behavior
Declarations generated at the specified location

### Actual Behavior
Declarations generated at the same directory as the output file.

### Additional Information
I'm using api-extractor to rollup d.ts files into a single one, so the declarations are output at a place that will be removed after 
 processing.
This works at version 8.3.1 with a problem that it generates file at relative location under output directory, that is when `declarationDir` is `types` and output.file is `dist/index.js`, declarations will be under `dist/types`. But after my collegue updated to 8.3.4 (by using ^8.3.1 in package.json), the behavior changes to that declarations generated at `dist/`, and the build progress breaks since api-extractor can't find the desired entry for d.ts rollup.

dominikbulaj:
Also confirmed. 

In `tsconfig.json` I have: `"declaration": false` and in `rollup.config.js` I got 
```
    typescript({
      compilerOptions: { declaration: true, declarationDir: './types' }
    }),
 ```
 
 When rollup runs declaration file(s) are placed in the output directory and not specified (`'./types'`)

himanshuchawla009:
+1

shellscape:
@himanshuchawla009 please do not post "me too," "+1" type replies. They spam anyone who has notifications on for issues (e.g. all maintainers). Please use the reaction buttons on the original post in the issue instead.

Alpha0723:
I got the same issue with 8.3.4.  Now I have to create new tsconfig.types.json with declaration config
  "compilerOptions": {
    "declaration": true,
    "declarationDir": "./dist/types",
    "emitDeclarationOnly": true,
  }
then use tsc command to generate d.ts to specific folder.
It works but it's ugly. Hope it could be fixed asap.

Basssiiie:
I got the issue as well. Reverting back to version 8.3.3 fixes the issue, so it seems it was introduced in the 8.3.4 release. The only change in that release is this; #1201, which may very well be the cause (also there's other people talking about problems there as well).

[...]

PR #1313, opened 2022-10-12
fix(typescript)!: incorrect declarations directory

georg94:
## Rollup Plugin Name: `plugin-typescript`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

- #1230

### Description
The commit 5f78173 changed the behaviour of the typescript plugin. This introduced the issue #1230: The `declarationDir` was now ignored for single file setups. I made some analysis and found the problem:
* In the construction of the plugin the typescript configuration is parsed
* while parsing all paths are made absolute, prepended with `process.cwd()` ([packages/typescript/src/options/tsconfig.ts#L147](https://github.com/rollup/plugins/blob/master/packages/typescript/src/options/tsconfig.ts#L147))
* in order to output declaration files the `baseDir` is determied from the parsed config
* the `id` (filename generated by typescript) of the declaration file is also an absolute path, generated with the config
* as the "prefixes" match the result file misses the `declarationDir`

Example:
```javascript
let tsConfig = {
  // [...]
  "declaration": true,
  "declarationDir": "types"
  // [...]
};
// After initial config processing
tsConfig = {
  // [...]
  "declaration": true,
  "declarationDir": "/home/user/app/types"
  // [...]
};
// declaration file generation
const id = '/home/user/app/types/index.d.ts';
const baseDir = parsedOptions.options.declarationDir; // -> '/home/user/app/types'
path.relative(baseDir, id) // gives `index.d.ts` instead of `types/index.d.ts`
```

There is another description of this error in https://github.com/rollup/plugins/issues/1230#issuecomment-1233630234

### DIFF
--- a/packages/typescript/src/index.ts
+++ b/packages/typescript/src/index.ts
@@ -157,6 +157,15 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
             (parsedOptions.options.declaration
               ? parsedOptions.options.declarationDir || parsedOptions.options.outDir
               : null);
+          const cwd = normalizePath(process.cwd());
+          if (
+            parsedOptions.options.declaration &&
+            parsedOptions.options.declarationDir &&
+            baseDir?.startsWith(cwd)
+          ) {
+            const declarationDir = baseDir.slice(cwd.length + 1);
+            baseDir = baseDir.slice(0, -1 * declarationDir.length);
+          }
           if (!baseDir && tsconfig) {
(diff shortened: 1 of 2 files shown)


NotWoods (approved):

shellscape:
Because this fix is modifying the output of the plugin, it's a breaking change and will be released as a new major.

shellscape (approved):

PR #1325, opened 2022-10-19
chore(repo): enable `consistent-type-imports` for typescript files

jerry-lllman:

This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

we added eslint rule `'@typescript-eslint/consistent-type-imports': 'error'` 
It can make some optimizations.
For example, if you `import typescript`, the whole typescript library will be imported, but the `import type typescript` this library will not be imported, saving a lot of volume. 

`typescript` documentation is described as follows
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,7 +9,8 @@ module.exports = {
     'sort-keys': 'off',
     'typescript-sort-keys/interface': 'off',
     'import/extensions': 'off',
-    'import/no-unresolved': 'off'
+    'import/no-unresolved': 'off',
+    '@typescript-eslint/consistent-type-imports': 'error'
   },
   overrides: [
     {
--- a/packages/alias/src/index.ts
+++ b/packages/alias/src/index.ts
@@ -1,4 +1,4 @@
-import { Plugin } from 'rollup';
+import type { Plugin } from 'rollup';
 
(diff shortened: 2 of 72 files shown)


lukastaegert (commented):
Thank you. IMO though, doing this manually is an unmaintainable chore. If we want this, we should throw

```
'@typescript-eslint/consistent-type-imports': 'error'
```

into the ESLint config. This rule has an auto-fixer.

jerry-lllman:
> 

ok, i add it, and changed the code that does not conform to this rule.

btw, my first commit is not manually, it's fixed by this rule~

lukastaegert:
> btw, my first commit is not manually, it's fixed by this rule~

😅 Makes sense

shellscape:
@jerry-lllman Please fill in the Description of the pull request template. I'm familiar with the rule and it's intent and purpose, but why do we need to enforce this? Please make your case in the Description.

jerry-lllman:
> @jerry-lllman Please fill in the Description of the pull request template. I'm familiar with the rule and it's intent and purpose, but why do we need to enforce this? Please make your case in the Description.

@shellscape I have added the description🥳

shellscape:
thanks!

Issue #1334, opened 2022-11-01
[@rollup/plugin-terser] Increase performance by running in parallel workers


dasa:
- Rollup Plugin Name: @rollup/plugin-terser
- Rollup Plugin Version: 0.1.0

### Feature Use Case

This is missing compared to the `rollup-plugin-terser` [implementation](https://github.com/TrySound/rollup-plugin-terser/blob/master/rollup-plugin-terser.js). This was noted at: https://github.com/rollup/plugins/pull/1323#issuecomment-1285935612

I'm seeing a four minute slowdown as compared to using the `rollup-plugin-terser` plugin when building about 3,500 files.

### Feature Proposal

This would be automatic as it was in the `rollup-plugin-terser` plugin. `rollup-plugin-terser` relied on [jest-worker](https://github.com/facebook/jest/tree/main/packages/jest-worker), but maybe https://nodejs.org/api/worker_threads.html could now be used instead.

tada5hi:
I will devote myself to that :v: :relaxed:

tada5hi:
Sorry for the delay guys.
But i was busy working on a online- & federated-learning architecture at my University :sweat_smile: .

Besides that i "wasted" my time on [routup](https://github.com/tada5hi/routup)  lightweight and extendable http interface based routing framework (alternative to express) :grin: .
I would be very happy if you would try it out, leave feedback or contribute to it :heart: 

https://github.com/rollup/plugins/pull/1341

tada5hi:
This issue should be solved soon and closed soon ✌️😊. Waiting fo shellscape to merge the pr.

tada5hi:
i would not call it work, rather suggestions from @dasa.

michaelfaith:
Fair enough.  I just meant it was still WIP and not simply waiting on merge.  Appreciate your work on it. 🍻

PR #1341, opened 2022-11-10
feat(terser): parallel execution

tada5hi:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

[1334](https://github.com/rollup/plugins/issues/1334)

### Description

Execute minify process in parallel


shellscape:
Thanks for opening this one. Is this based on existing work somewhere or brand new work? If existing, please document from where so we have a reference point. If it's new, please add some code comments for code in spots you think people may have questions or that may raise eyebrows.

tada5hi:
> Thanks for opening this one. Is this based on existing work somewhere or brand new work? If existing, please document from where so we have a reference point. If it's new, please add some code comments for code in spots you think people may have questions or that may raise eyebrows.

The idea of merging and populating the nameCache of child processes is from @TrySound, but it is basically the idea not the code. Nevertheless, i already mentioned him in the Readme.

shellscape:
That was less about attribution and more about helping the other maintainers understand the change.

tada5hi:
> That was less about attribution and more about helping the other maintainers understand the change.
No problem :blush: 
If you or anyone else have any questions, I'm more than happy to help :v: :innocent:

packages/terser/src/module.ts
```diff
@@ -0,0 +1,67 @@
...
+      if (outputOptions.format === 'es') {
+        defaultOptions.module = true;
+      }
+
+      if (outputOptions.format === 'cjs') {
+        defaultOptions.toplevel = true;
+      }
+
+      try {
+        const { code: result, nameCache } = await callWorker(__filename, {
```
dasa, line 25: Is this going to create a new Worker for each call to `renderChunk`?

From the [Worker threads](https://nodejs.org/api/worker_threads.html) doc:

> In practice, use a pool of Workers for these kinds of tasks. Otherwise, the overhead of creating Workers would likely exceed their benefit.

The doc also links to an example pool implementation: https://nodejs.org/api/async_context.html#using-asyncresource-for-a-worker-thread-pool

michaelfaith:
> > That was less about attribution and more about helping the other maintainers understand the change.
> 
> No problem 😊 If you or anyone else have any questions, I'm more than happy to help ✌️ 😇

I think the request was for more documentation.

packages/terser/src/module.ts
tada5hi: Yes, at the moment that is the case.
The idea was to let the operating system decide when enough resources are available to run unprocessed worker threads.
But in general it might be better to have more control over it, to better control the cpu usage.

tada5hi: @dasa  it is now possible to cotnrol how many workers can run at the same time. By default it is equal to the amount of cpu cores.

packages/terser/src/index.ts
```diff
@@ -1,24 +1,8 @@
-import { NormalizedOutputOptions, RenderedChunk } from 'rollup';
-import { minify, MinifyOptions } from 'terser';
+import { runWorker } from './worker';
+import terser from './module';
 
-export default function terser(options?: MinifyOptions) {
-  return {
-    name: 'terser',
+Promise.resolve().then(() => runWorker());
```
dasa, line 4: Is this line needed?

packages/terser/src/module.ts
dasa: It doesn't appear to me that the workers are being reused. If that's the case, it seems this would still be true, "the overhead of creating Workers would likely exceed their benefit."

packages/terser/src/index.ts
tada5hi: In theory not, but my ide (webstorm) is complaining about it and eslint in some configurations might too, due the fact that the `runWorker` method is async.

tada5hi: was speaks against the usage of `Promise.resolve()` ? 
Are there any impacts or undesired behaviour ?

tada5hi: good point :blush: . I removed it :v:

packages/terser/README.md
```diff
@@ -47,6 +47,26 @@ Then call `rollup` either via the [CLI](https://www.rollupjs.org/guide/en/#comma
 The plugin accepts a terser [Options](https://github.com/terser/terser#minify-options) object as input parameter,
 to modify the default behaviour.
 
+Besides, the terser options, it is also possible to pass the option `maxWorkers`,
```
shellscape, line 50:
```suggestion
In addition to the `terser` options, it is also possible to provide the following options:
```

Following this line, the `maxWorkers` should be documented in the same format as the options in other plugins. e.g. https://github.com/rollup/plugins/blob/master/packages/alias/README.md#options

packages/terser/README.md
```diff
@@ -55,5 +75,6 @@ to modify the default behaviour.
 
 ## Credits
 
-This package was originally developed by [https://github.com/TrySound](TrySound) but is not
+The original terser plugin was developed by [TrySound](https://github.com/TrySound) but is not
 maintained anymore.
+The current implementation is developed and maintained by [tada5hi](https://github.com/tada5hi).
```
shellscape, line 80:
```suggestion
Based on [`rollup-terser-plugin`](https://www.npmjs.com/package/rollup-plugin-terser), moved here with permission.
```

We tend not to tag individual maintainers.

shellscape:
@TrySound please have a look through this one. once merged, it will probably be a good idea to deprecate the other plugin (https://www.npmjs.com/package/rollup-plugin-terser) and archive the old repo, pointing people here. If you need a hand with that, please feel free to add me to each and I'll take care of it.

packages/terser/README.md
tada5hi: That is fine, but than i would rather remove this line. Besides, sharing the logic for 10 lines of source code, the implementation vary and the term "moved" is defintly wrong!

packages/terser/src/index.ts
```diff
@@ -1,24 +1,8 @@
-import { NormalizedOutputOptions, RenderedChunk } from 'rollup';
-import { minify, MinifyOptions } from 'terser';
+import { runWorker } from './worker';
+import terser from './module';
 
-export default function terser(options?: MinifyOptions) {
-  return {
-    name: 'terser',
+runWorker();
```
dasa, line 4: Does this need to be called here? It seems like it'll just return since it's on the main thread.

packages/terser/src/worker-pool.ts
```diff
@@ -0,0 +1,117 @@
...
+      }
+      called = true;
+
+      this.workers -= 1;
+
+      task.cb(err, output);
+      this.emit(symbol);
+    };
+
+    const worker = new Worker(this.filePath, {
```
dasa, line 96: Are Workers ever reused and returned to the pool? If not, this would be better to avoid the overhead of creating new workers.

packages/terser/src/index.ts
tada5hi: Yes, it is required, due the fact that the files are bundled to one file. 
Thererfore, we must always expect, that the bundled file is called as worker.

packages/terser/README.md
```diff
@@ -9,11 +9,11 @@
 
 # @rollup/plugin-terser
 
-🍣 A Rollup plugin to generate a minified output bundle.
+🍣 A Rollup plugin to generate a minified output bundle wiht terser.
```
dasa, line 12:
```suggestion
🍣 A Rollup plugin to generate a minified output with terser.
```

packages/terser/src/worker-pool.ts
tada5hi: No, it is more or less a fake worker pool, to ensure that not more than x workers run i parallel and the event loop is not blocked, for other applications or tasks, by blocking the event loop, with open workers.

https://nodejs.org/en/docs/guides/dont-block-the-event-loop/

packages/terser/README.md
tada5hi: done

packages/terser/src/worker-pool.ts
dasa: Ah OK, maybe a future enhancement would be to create a true worker pool like the one shown here: https://nodejs.org/api/async_context.html#using-asyncresource-for-a-worker-thread-pool

tada5hi:
@shellscape is TrySound's feedback required or how long do you plan to wait for his review?

shellscape:
It may be that Github doesn't update that in realtime. You're also on the merge commit. https://github.com/rollup/plugins/commit/2d67e34e9b27adf235a637b0b2a957abaf931c65. I wouldn't get too wrapped up with vanity metrics.

You should however add yourself to https://github.com/rollup/plugins/blob/master/CODEOWNERS for this plugin so you get automated notifications when someone opens a PR against it.

> is TrySound's feedback required or how long do you plan to wait for his review?

I would like it, yes. he's a busy guy so it may take some time. moving forward that won't be the case.

shellscape:
Welp. It's been a week so we'll move forward. Thanks for your patience.

PR #1346, opened 2022-11-14
fix(wasm): rejecting with error on node.js file read

nickbabcock:
## Rollup Plugin Name: `wasm`

This PR contains:

- [x] bugfix

Are tests included?

- [x] yes

Breaking Changes?

- [x] no

### Description

When node.js is reading the wasm file and encounters an error (like if the file doesn't exist or user doesn't have permissions to read). It will still try and call `_instantiateOrCompile` with `undefined`, which results in an uncaught exception, possibly obscuring the error. As instead of the file read error, the following error will return:

> TypeError: WebAssembly.compile(): Argument 0 must be a buffer source

This can be replicated with the following script (don't execute at a REPL):

```js
(async () => {
  await new Promise((resolve, reject) => {
    reject(new Error("permission error"));
    resolve(WebAssembly.compile(undefined));
  });
})();
```

Node.js will exit with an error code and print only the type error to stderr.

If we try and catch the error at the outer level

```js
(async () => {
  try {
    await new Promise((resolve, reject) => {
      reject(new Error("permission error"));
      resolve(WebAssembly.compile(undefined));
    });
  } catch (ex) {
    console.log(`this is the error: ${ex}`)
  }
})();
```

The log statement will exclude the type error:

> this is the error: Error: permission error

But node.js will still exit with an error code and the type error is printed to stderr.

This is the behavior of node.js default [uncaught exception handler](https://nodejs.org/api/process.html#event-uncaughtexception)

> By default, Node.js handles such exceptions by printing the stack trace to stderr and exiting with code 1

This commit avoids the uncaught exception by only invoking `resolve` behind a conditional on an absent error.

This commit doesn't change the status of the promise as calling `resolve` after `reject`, while legal, doesn't really do anything [(ref)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/Promise#description):

> Only the first call to `resolveFunc` or `rejectFunc` affects the promise's eventual state, and subsequent calls to either function can neither change the fulfillment value/rejection reason nor toggle its eventual state from "fulfilled" to "rejected" or opposite

So depending on the runtime and how the user is overriding the default handler, uncaught exceptions may be handled differently

The test executes in a worker so that it can require the appropriate node js modules. I saw that another Wasm test used a worker, but would never execute as require is not defined, so this commit also fixes that test case.

### DIFF
--- a/packages/wasm/src/helper.ts
+++ b/packages/wasm/src/helper.ts
@@ -10,9 +10,9 @@ return new Promise((resolve, reject) => {
   fs.readFile(path.resolve(__dirname, filepath), (error, buffer) => {
     if (error != null) {
       reject(error)
+    } else {
+      resolve(_instantiateOrCompile(buffer, imports, false))
     }
-
-    resolve(_instantiateOrCompile(buffer, imports, false))
   });
 });
 `;
--- a/packages/wasm/test/test.mjs
+++ b/packages/wasm/test/test.mjs
@@ -1,6 +1,7 @@
 import { createRequire } from 'module';
 import { sep, posix, join } from 'path';
 import { fileURLToPath } from 'url';
+import { Worker } from 'worker_threads';
 
 import { rollup } from 'rollup';
 import globby from 'globby';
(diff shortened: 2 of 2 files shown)


shellscape:
> I'm secretly hoping that the changes are trivial enough to not warrant a test as this bugfix revolves around the error message that is propagated and may be runtime dependent.

What can we do to protect against a regression (e.g. _"hey this doesn't do anything on the runtime I'm using, let's remove useless code, for some reason"_) without a test?

nickbabcock:
I was able to write a test, but it required fixing an existing test that wasn't executing. Hopefully this is ok.

I've updated the description so that it is more accurate (the problem is that the type error is an uncaught exception, which can obscure the rejected promise).

Thanks for the patience 😄

shellscape (approved):

shellscape:
Thanks for putting the work in on the plugin, and apologies for the delay in review.

thanks!

Issue #1366, opened 2022-12-05
[@rollup/plugin-terser] esm module uses __filename


jakearchibald:
- Rollup Plugin Name: @rollup/plugin-terser
- Rollup Plugin Version: 0.2.0
- Rollup Version: 3.6.0
- Link to reproduction _(⚠️ read below)_: https://github.com/rollup/plugins/blob/master/packages/terser/src/module.ts#L9

Error:

```
[!] ReferenceError: __filename is not defined
ReferenceError: __filename is not defined
    at terser (file:///…/node_modules/@rollup/plugin-terser/dist/es/index.js:121:19)
```

Should use `import.meta.url` instead.

Andrew-web-coder:
This package is broken. Does the authors even care?

shellscape:
@jakearchibald for future reference, please keep all bits of the templates in the issues.

jakearchibald:
It felt redundant, but:

### Expected Behavior

Run without throwing an error.

### Actual Behavior

Threw the following error:

```
[!] ReferenceError: __filename is not defined
ReferenceError: __filename is not defined
    at terser (file:///…/node_modules/@rollup/plugin-terser/dist/es/index.js:121:19)
```

shellscape:
All good. @tada5hi is the lead for that plugin, so hopefully they see this soon

josdejong:
For reference: I think I'm hitting this issue, when upgrading `@rollup/plugin-terser` from `0.1.0` to `0.2.0` in an ES project with  `"type": "module"`. I get the following error:

```
[!] RollupError: Node tried to load your configuration as an ES module even though it is likely CommonJS. To resolve this, change the extension of your configuration to ".cjs" or pass the "--bundleConfigAsCjs" flag.

Original error: __filename is not defined in ES module scope
This file is being treated as an ES module because it has a '.js' file extension and 'C:\...\package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
https://rollupjs.org/guide/en/#--bundleconfigascjs
ReferenceError: __filename is not defined in ES module scope
This file is being treated as an ES module because it has a '.js' file extension and 'C:\...\package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
    at terser (file:///C:/.../node_modules/@rollup/plugin-terser/dist/es/index.js:121:19)
    at file:///C:/.../rollup.config.bundle.js:50:19
[...]

PR #1367, opened 2022-12-06
fix(terser): __filename not defined

tada5hi:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

fixes https://github.com/rollup/plugins/issues/1366

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -3,3 +3,4 @@
 /packages/babel/ @Andarist
 /packages/image/ @tjenkinson
 /packages/node-resolve/ @tjenkinson
+/packages/terser/ @tada5hi
--- a/packages/terser/src/module.ts
+++ b/packages/terser/src/module.ts
@@ -1,3 +1,5 @@
+import { fileURLToPath } from 'url';
+
 import type { NormalizedOutputOptions, RenderedChunk } from 'rollup';
 import { hasOwnProperty, isObject, merge } from 'smob';
 
@@ -5,8 +7,25 @@ import type { Options } from './type';
 import { WorkerPool } from './worker-pool';
 
 export default function terser(options: Options = {}) {
+  let filePath: string | undefined;
+
+  if (typeof __filename !== 'undefined') {
+    filePath = __filename;
+  } else {
+    // @ts-ignore
+    // eslint-disable-next-line no-lonely-if
+    if (typeof import.meta !== 'undefined') {
+      // @ts-ignore
+      filePath = fileURLToPath(import.meta.url);
+    }
+  }
+
+  if (typeof filePath === 'undefined') {
+    throw new Error('File path could not be determined.');
+  }
+
   const workerPool = new WorkerPool({
-    filePath: __filename,
+    filePath,
     maxWorkers: options.maxWorkers
   });
 
--- a/packages/terser/src/module.ts
+++ b/packages/terser/src/module.ts
@@ -6,27 +6,12 @@ import { hasOwnProperty, isObject, merge } from 'smob';
 import type { Options } from './type';
 import { WorkerPool } from './worker-pool';
 
-export default function terser(options: Options = {}) {
-  let filePath: string | undefined;
-
-  if (typeof __filename !== 'undefined') {
-    filePath = __filename;
-  } else {
-    // @ts-ignore
-    // eslint-disable-next-line no-lonely-if
-    if (typeof import.meta !== 'undefined') {
-      // @ts-ignore
-      filePath = fileURLToPath(import.meta.url);
-    }
-  }
-
-  if (typeof filePath === 'undefined') {
-    throw new Error('File path could not be determined.');
-  }
+export default function terser(input: Options = {}) {
+  const { maxWorkers, ...options } = input;
 
   const workerPool = new WorkerPool({
-    filePath,
-    maxWorkers: options.maxWorkers
+    filePath: fileURLToPath(import.meta.url),
+    maxWorkers
   });
 
   return {
--- a/packages/terser/tsconfig.json
+++ b/packages/terser/tsconfig.json
@@ -2,6 +2,7 @@
   "extends": "../../tsconfig.base.json",
   "include": ["src/**/*", "types/**/*"],
   "compilerOptions": {
-    "noUnusedParameters": false
+    "noUnusedParameters": false,
+    "module": "ESNext"
   }
 }


packages/terser/src/module.ts
```diff
@@ -1,12 +1,31 @@
...
 
 export default function terser(options: Options = {}) {
+  let filePath: string | undefined;
+
+  if (typeof __filename !== 'undefined') {
+    filePath = __filename;
+  } else {
+    // @ts-ignore
+    // eslint-disable-next-line no-lonely-if
+    if (typeof import.meta !== 'undefined') {
```
jakearchibald, line 17: Could be an `else if`?

tada5hi: In general of cause, but i couldn't provide `@ts-ignore` to the `else if` statement :/

shellscape: in general we shouldn't have to use `// @ts-ignore`. I can take a look at your fork later tonight or tomorrow to find a better solution.

tada5hi: That would be great ☺️. The issue here is this that tsc is complaining about the usage of imoort.meta because the package is not an esm module.

dasa: Node throws a syntax error for `import.meta` in CJS modules.

ntkme: It is very picky on mixing CJS and ESModule code, it will complain one way or the other and just refuse to run. One solution I found is to have two completely different entrypoint files, one for CJS and one for ESModule. Compile them separately, then in `package.json`, define `main` for CJS and `module` for ESModule.

shellscape: does it complain about using `global['import']`?

ntkme: @shellscape `global['import']` do not complain in CJS mode, but it does not work in MJS mode either, because `import` is a special keyword, that `global['import'].meta.url` is different than `import.meta.url`.

Having the following code:

``` js
console.log(global['import'])
```

Run it as `node test.cjs`:
```
undefined
```
Run it as `node test.mjs`:
```
undefined
```

mxdvl: You could also introduce an isomorphic `__filename` as suggested in [this blog post](https://antfu.me/posts/isomorphic-dirname):

```ts
import { fileURLToPath } from 'url'

const filePath = typeof __filename !== 'undefined'
    ? __filename
    : fileURLToPath(import.meta.url)
```

ntkme: @mxdvl That article is outdated. It won’t work anymore in latest node because keyword import detected in anywhere in CSJ mode is now a syntax error.

mxdvl: Oh, could [`@rollup/plugin-replace`](https://www.npmjs.com/package/@rollup/plugin-replace) be an alternative? That might be lighter touch than two separate entry points.

shellscape: that was my initial thought but _would_ double the size of the package

tada5hi: > @mxdvl That article is outdated. It won’t work anymore in latest node because keyword import detected in anywhere in CJS mode is now a syntax error.

Besides, that was exactly what i have implemented in the pr. ^^

tada5hi: > that was my initial thought but _would_ double the size of the package

but wouldn't it just affect the size during development/build, because it would be a defined as dev dependency ?

UmamiAppearance:
Without having testet this, the worker thread seem to also take [urls](https://nodejs.org/api/worker_threads.html#new-workerfilename-options).  
  
A simple switch between `__filename` and `import.meta.url` should be enough for the different entry points.

jonkoops:
Got an alternative fix for this in #1374. Looks like `import.meta.url` is transpiled when compiling to CommonJS, so it should remain backwards compatible.

tada5hi:
> Got an alternative fix for this in #1374. Looks like `import.meta.url` is transpiled when compiling to CommonJS, so it should remain backwards compatible.

can confirm 👍

CODEOWNERS
```diff
@@ -3,3 +3,4 @@
 /packages/babel/ @Andarist
 /packages/image/ @tjenkinson
 /packages/node-resolve/ @tjenkinson
+/packages/terser/ @tada5hi
```
jonkoops, line 6: Can @tada5hi get write access to the repo so that this is no longer a problem to add, or should this be removed from the PR?

packages/terser/src/module.ts
```diff
@@ -1,13 +1,17 @@
...
+
 import type { NormalizedOutputOptions, RenderedChunk } from 'rollup';
 import { hasOwnProperty, isObject, merge } from 'smob';
 
 import type { Options } from './type';
 import { WorkerPool } from './worker-pool';
 
-export default function terser(options: Options = {}) {
+export default function terser(input: Options = {}) {
+  const { maxWorkers, ...options } = input;
```
jonkoops, line 10: I believe the destructuring for this could happen in the same place the parameter is defined.

```suggestion
export default function terser({ maxWorkers, ...options }: Options = {}) {
```

jonkoops:
@tada5hi just FYI you can use a keyword in the description of your PR to link it to the issue (see the [GitHub documentation](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)). That way it will be closed automatically when the PR is merged.

tada5hi:
> @tada5hi just FYI you can use a keyword in the description of your PR to link it to the issue (see the [GitHub documentation](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)). That way it will be closed automatically when the PR is merged.

Oh cool, thank you :blush:.

CODEOWNERS
tada5hi: @jonkoops i think that should be fine. @shellscape told me to add this line.

shellscape:
OK are we in agreement this can be merged?

jonkoops (approved):

jonkoops:
This looks good, no objections from me 👍

tada5hi:
@shellscape yes from my side.

shellscape:
Thanks all. This will be automatically published shortly.

PR #1372, opened 2022-12-08
feat(wasm): add wasm fileName customization

nickbabcock:
## Rollup Plugin Name: `wasm`

This PR contains:

- [x] feature

Are tests included?

- [x] yes

Breaking Changes?

- [x] no

### Description

The current implementation of the plugin always emits a wasm file with a filename containing the hash of the contents. This behavior is counter-productive to libraries that distribute Wasm; it would be better if the Wasm is in a predictable location.

Thus this commit partially ports the `fileName` option of the url plugin to the wasm plugin, exposing the following substitutable variables: `[hash]`, `[name]`, and `[extname]` (which is may be argued as redundant as it should always be ".wasm").

This commit is backwards compatible by keeping the default behavior of emitting `[hash][extname]`.


shellscape (approved):

shellscape:
thanks!

PR #1374, opened 2022-12-12
fix(terser): replace `__filename` with `import.meta.url`

### LINKED ISSUE #1366: [@rollup/plugin-terser] esm module uses __filename

jonkoops:
## Rollup Plugin Name: terser

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

- Closes #1366
- Supersedes #1367

### Description

Because `__filename`  does not exist in the ES modules this plugin errors out when used in a Rollup config written in the same format. This PR replaces the `__filename` constant with `import.meta.url` instead, which is available in ES modules.

When building the CommonJS version `import.meta.url` is replaced with `__dirname` at compile time, making sure that it remains compatible.

### DIFF
--- a/packages/terser/tsconfig.json
+++ b/packages/terser/tsconfig.json
@@ -2,6 +2,7 @@
   "extends": "../../tsconfig.base.json",
   "include": ["src/**/*", "types/**/*"],
   "compilerOptions": {
+    "module": "ES2020",
     "noUnusedParameters": false
   }
 }
--- a/packages/terser/src/module.ts
+++ b/packages/terser/src/module.ts
@@ -1,3 +1,5 @@
+import { fileURLToPath } from 'url';
+
 import type { NormalizedOutputOptions, RenderedChunk } from 'rollup';
 import { hasOwnProperty, isObject, merge } from 'smob';
 
@@ -6,7 +8,7 @@ import { WorkerPool } from './worker-pool';
 
 export default function terser(options: Options = {}) {
   const workerPool = new WorkerPool({
-    filePath: __filename,
+    filePath: fileURLToPath(import.meta.url),
     maxWorkers: options.maxWorkers
   });
 


jonkoops:
Here is an example of the transpiled CommonJS code that initializes the worker. I believe this is enough to keep backwards compatibility:

```js
const workerPool = new WorkerPool({
    filePath: url.fileURLToPath((typeof document === 'undefined' ? new (require('u' + 'rl').URL)('file:' + __filename).href : (document.currentScript && document.currentScript.src || new URL('index.js', document.baseURI).href))),
    maxWorkers: options.maxWorkers
});
```

packages/terser/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "../../tsconfig.base.json",
   "include": ["src/**/*", "types/**/*"],
   "compilerOptions": {
+    "module": "ES2020",
```
jonkoops, line 5: Had to set this flag, otherwise the TypeScript compiler will start complaining about using `import.meta.url` in something that compiles to a CommonJS target. Which of course is a non-issue as Rollup is used to compile the code, not the TypeScript compiler.

That said, should this not be the default? Then opting back into CommonJS where necessary (not even sure if that is needed).

jonkoops:
Closing this PR in favor of #1367, which has now been updated with a similar implementation.

PR #1376, opened 2022-12-14
fix(terser): print message when terser worker throws an error

12Me21:

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

(Extremely minor change, just adding a `console.error()` call)

Currently, if an error occurs inside a worker, it exits without printing a useful error message.  
e.g. `Error: Minify worker stopped with exit code 1`, in https://github.com/rollup/plugins/issues/1366#issuecomment-1347708877, when the actual error was `DefaultsError: 'maxWorkers' is not a supported option`

### DIFF
--- a/packages/terser/src/worker.ts
+++ b/packages/terser/src/worker.ts
@@ -42,6 +42,7 @@ export async function runWorker() {
 
     parentPort.postMessage(output);
   } catch (e) {
+    console.error(e);
     process.exit(1);
   }
 }


shellscape (changes requested):
While this may be immediately useful to you, it doesn't pass muster for a quality addition. Namely, errors should be printed using [Rollup's native method](https://www.rollupjs.org/guide/en/#thiserror). And this will need an accompanying test that checks the output, in order to be accepted.

tada5hi:
closed in favour of #1394

Issue #1238, opened 2022-08-10
Error: Unexpected token (Note that you need plugins to import files that are not JavaScript)


khteh:
`tsc --project tsconfig.build.json && npm run rollup`:
```
[!] Error: Unexpected token (Note that you need plugins to import files that are not JavaScript)
../core/greetings.ts (7:11)
5: import url from 'url'
6: export class Greetings {
7:     public greetings(req: Request, res: Response, next: NextFunction) {
              ^
8:         var url_parts = url.parse(req.url, true);
9:         var query = url_parts.query;
Error: Unexpected token (Note that you need plugins to import files that are not JavaScript)
    at error (/usr/src/Node.JSRestAPI/node_modules/rollup/dist/shared/rollup.js:198:30)
    at Module.error (/usr/src/Node.JSRestAPI/node_modules/rollup/dist/shared/rollup.js:12543:16)
    at Module.tryParse (/usr/src/Node.JSRestAPI/node_modules/rollup/dist/shared/rollup.js:12920:25)
    at Module.setSource (/usr/src/Node.JSRestAPI/node_modules/rollup/dist/shared/rollup.js:12825:24)
    at ModuleLoader.addModuleSource (/usr/src/Node.JSRestAPI/node_modules/rollup/dist/shared/rollup.js:22267:20)

npm ERR! Lifecycle script `rollup` failed with error: 
npm ERR! Error: command failed 
npm ERR!   in workspace: webapi@1.0.0 
npm ERR!   at location: /usr/src/Node.JSRestAPI/src/webapi 
npm ERR! Lifecycle script `build` failed with error: 
npm ERR! Error: command failed 
npm ERR!   in workspace: webapi@1.0.0 
npm ERR!   at location: /usr/src/Node.JSRestAPI/src/webapi 
```
- Rollup Plugin Name: @rollup/plugin-typescript, 
- Rollup Plugin Version: 8.3.3
- Rollup Version: rollup v2.70.1
- Operating System (or Browser): Ubuntu 22.04, Chrome
- Node Version: v18.7.0
- Link to reproduction _(⚠️ read below)_: https://github.com/khteh/Node.JSRestAPI

### Expected Behavior
No error.

### Actual Behavior
Error.

shellscape:
Thanks for opening an issue. Citing the issue template:

>   🚨 Issues WITHOUT a valid reproduction WILL BE CLOSED!
> 
>   Please provide one by:
[...]

Issue #1083, opened 2022-01-07
Error: Unexpected token (Note that you need plugins to import files that are not JavaScript)


hotamateurxxx:
- Rollup Plugin Name: @rollup/plugin-typescript
- Rollup Plugin Version: ^8.3.0
- Rollup Version: ^2.63.0
- TypeScript Version: ^4.5.4
- Node Version: v17.2.0
- Operating System (or Browser): Windows 10
- Link to reproduction: https://github.com/hotamateurxxx/rollup-typescript-bug

### Expected Behavior
Successfull build.

### Actual Behavior
```CMD
yarn build

...

Error: Unexpected token (Note that you need plugins to import files that are not JavaScript)
    at error (D:\Projects\github.com\rollup-typescript-bug\node_modules\rollup\dist\shared\rollup.js:158:30)
    at Module.error (D:\Projects\github.com\rollup-typescript-bug\node_modules\rollup\dist\shared\rollup.js:12423:16)
    at Module.tryParse (D:\Projects\github.com\rollup-typescript-bug\node_modules\rollup\dist\shared\rollup.js:12826:25)
    at Module.setSource (D:\Projects\github.com\rollup-typescript-bug\node_modules\rollup\dist\shared\rollup.js:12729:24)
    at ModuleLoader.addModuleSource (D:\Projects\github.com\rollup-typescript-bug\node_modules\rollup\dist\shared\rollup.js:22202:20)
```

### Additional Information

Got closed issue [ Typescript plugin 4.0.0 Error: Unexpected token (Note that you need plugins to import files that are not JavaScript) #287](https://github.com/rollup/plugins/issues/287) with plugin version that should to contain fix:

`node_modules\@rollup\plugin-typescript\CHANGELOG.md`
> ## v8.2.1
> 
> _2021-03-26_
> 
> ### Bugfixes
> 
> - fix: bump TypeScript version (#818)
> - fix: update readme and peerDeps version (#830)

shellscape:
Please update your issue with the full template information.

hotamateurxxx:
Update reproduction repo. Bug cause in case when you need to push build scripts into subfolder.

@shellscape, is that issue corresponds to full template information now?

xenobytezero:
Rollup Plugin Name: @rollup/plugin-typescript
Rollup Plugin Version: ^8.3.1
Rollup Version: ^2.68.0
TypeScript Version: ^4.5.5
Node Version: v16.13.2
[...]

Issue #892, opened 2021-06-01
[TypeScript] Error: Unexpected token


Jack-Works:
- Rollup Plugin Name: @rollup/plugin-typescript
- Rollup Plugin Version: 8.2.1
- Rollup Version: 2.50.5
- Operating System (or Browser): Windows
- Node Version: 16.2.0
- Link to reproduction _(⚠️ read below)_: https://github.com/Jack-Works/rollup-plugin-bug-287-reproduce/

```shell
> npx rollup -c
```

### Expected Behavior

Can compile

### Actual Behavior

> [!] Error: Unexpected token (Note that you need plugins to import files that are not JavaScript)

### Additional Information

related #287

shellscape:
Duplicate of #287. Please comment on the other issue. If youd like to see it reopened/resolved pleade consider contributing.

Jack-Works:
> Please comment on the other issue

It's locked and unresolved.

Jack-Works:
Oh is this resolved in #818?

shellscape:
Thanks, I didn't realized it was locked. Yes, #818 was targeted to resolve _most_ of the problems that caused the error. Given the size of the framework, we couldn't claim there aren't cases which it didn't resolve. That fix did land in https://github.com/rollup/plugins/blob/master/packages/typescript/CHANGELOG.md#v821. I'll reopen this as a separate issue, separate reproduction.

waynevanson:
> I'll reopen this as a separate issue, separate reproduction.

Hi @shellscape, I'm looking to follow the trail but can't see the issue you've mentioned. If you've made it, what issue was it?

ZYinMD:
For those who still get this issue, fwiw, check if the parent directory or grand parent directory is a symlink. If yes, cd into the actual directory and try again.

kibin:
I had the same problem (plugin-typescript@11.0.0, rollup@3.20.2, node@18.15.0, typescript@4.9.5) and could only resolve it after extensive reading of original thread, especially [this comment](https://github.com/rollup/plugins/issues/287#issuecomment-701923921). thought I need to share my findings for others who struggle as well.
seems like `outDir` field is still a major offender for this plugin. code isn't compiling until it's removed from `tsconfig.json`.

PR #1377, opened 2022-12-15
Fix(typescript): Unexpected Token

glenn2223:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: #892 #1083 #1238 

### Description

The `id` passed to `async load(id)` has certain edge cases when drives letters can de different cases or file**names** can have its case changed. By normalising the `emittedFiles` we will always find the `code` we need and, as such, will not throw

#### Proof of issue/concept

**Machine1:** Windows 10.0.19044

I was able to compile [TopMarksDevelopment/JavaScript.Autocomplete](https://github.com/TopMarksDevelopment/JavaScript.Autocomplete), but not [TopMarksDevelopment/JavaScript.HoverBox](https://github.com/TopMarksDevelopment/JavaScript.HoverBox). The only difference was that the index file was named `Index` instead of `index`. Changed that and then it compiled.

Rather than trudging on - because I was to **improve open source**, when I can 😁 - I started tinkering. When debugging I noticed that the `code` wasn't being returned as there was this name inconsistency between that in `emittedFiles` and the `id` received in `load(id)`.

Applied the change replaced the compiled code in my project's `node_modules` and it compiled, whilst keeping the file name as `Index`.

**Machine 2:** Windows Server 2019 10.0.17763

I have a server that was having the same issue - this time attempting with `gulp`.

Debugging this time showed me that the drive letter character case was being changed for some reason - and not on every file that it was working on at the time.

Again, replaced the source with the changed compile code and it once again compiled. YAY!


shellscape:
I'd like to get some clarity about what causes this edge case - why `emittedFiles` contains asset names that pass; is it TypeScript? is it this plugin?

glenn2223:
> I'd like to get some clarity about what causes this edge case - why `emittedFiles` contains asset names that pass; is it TypeScript? is it this plugin?

The name that gets passed to the `load(id)` function is wrong - the `emittedFiles[]` is right - as shown further down
image

This means that when the `getEmittedFile(...)` is called the filename isn't found and no code is returned and, as such, rollup thinks there's no parser installed
image

So, strictly speaking, it's a `rollup` issue. But, personally, I think normalizing should transform the case too, when comparing, to stop any edge cases, wherever they may be/pop-up

I'm unable to replicate/locate the issue with the windows server, but it was randomly getting a lowercase drive letter for some files - which again would mean `find` doesn't work

> Also looks like these changes broke some existing tests on Windows.

Sorry, I'll take a look at the tests

glenn2223:
> Also looks like these changes broke some existing tests on Windows

Looks like the `tsconfig.tsbuildinfo` files are being dumped inside a cache folder `.rollup.cache` - which isn't accounted for in the tests
image

glenn2223:
Not sure why the typescript test failed, worked in my run. Plus completely irrelevant to my changes 🤷‍♂️
image

shellscape:
> So, strictly speaking, it's a rollup issue

This was my suspicion based on the original description of the problem. I'd like to get this addressed in Rollup core before we make concessions for that in plugins.

lukastaegert:
> So, strictly speaking, it's a rollup issue

I do not think this is the case. Rollup core does not change imported ids. It does not change slashes and it definitely does not do anything with drive letters. Between emission and `load` hook, the id is passed through all `resolveId` hooks that all have the full power to do that. I strongly suspect TypeScript/the TypeScript plugin rewrote at least the slashes in the `resolveId` hook of the TypeScript plugin.

I remember having some older issue around TypeScript normalizing slashes while Rollup does not.

glenn2223:
> I do not think this is the case. Rollup core does not change imported ids. It does not change slashes and it definitely does not do anything with drive letters. Between emission and `load` hook, the id is passed through all `resolveId` hooks that all have the full power to do that. I strongly suspect TypeScript/the TypeScript plugin rewrote at least the slashes in the `resolveId` hook of the TypeScript plugin.
> 
> I remember having some older issue around TypeScript normalizing slashes while Rollup does not.

Thanks for the reply. Sorry, I'm not too familiar with the rollup structure/architecture.

So I've found the actual cause of the issue. In my config, I have `input: 'src/index.ts'` which is the cause of the naming inconsistency.

I tested with `src/iNdex.ts` in the config and `resolveId(imported, ...)` matched. As the  `resolveId(..., importer)` is `null` this typescript plugin returns `null` so that rollup (or I suppose another plugin) can resolve it. After the first return, the same function is called again, but this time with the _KIND OF_ resolved path `"C:\\Users\\Glenn\\Documents\\GitHub\\HoverBox\\src\\iNdex.ts"` (note the name actually hasn't been resolved to match the actual file. **In my opinion** the actual path should be resolved correctly - as rollup already knows the file exists.

---

In regards to the drive letter changing randomly. There are some inconsistencies in npm's path resolutions - like nodejs/node#37307. Again, **in my opinion**, comparisons of any form should be case normalized too.

glenn2223:
_**Note:**_ with the change I get the below
image

glenn2223:
So case sensitivity is ignored by node (yes, I know there's more to it). Below is an article touching on the random case changes in paths and drive letters:  
https://mnaoumov.wordpress.com/2019/10/18/windows-case-insensitive-file-names-nightmares-with-node-js/

Basically, I think it's now at the point of doing any `includes` type checks with the case normalised too. Which maybe should be mentioned to all plugins, noted on the plugin template and FAQ'd too 🤷‍♂️.

I'll take a look and see if there are any other `includes` in the typescript plugin. Maybe it's worth having a function/class in the utility module that can handle case-insensitive checks - I suppose I'd need both (in the case of this plugin), a function for arrays and something for key-value pairs

glenn2223:
> I'll take a look and see if there are any other `includes` in the typescript plugin

Don't think there is any, checked the cache version (right after my change) and that does file checks rather than string comparisons - so I left it alone.

glenn2223:
Hey @shellscape/@lukastaegert, sorry to pester but have you any thoughts/progress on this? Would you like me to create a PR for some insensitive comparison utility tool(s)? Would you like two types - as I mentioned above (copied below)?

>Maybe it's worth having a function/class in the utility module that can handle case-insensitive checks - I suppose I'd need both (in the case of this plugin), a function for arrays and something for key-value pairs

shellscape:
I read through the Node core issue and have read all of the info you've dropped here. Thanks for putting all of this together.

> So case sensitivity is ignored by node

Yes. It's a legacy issue that they really should have fixed by now, but the rate of error is still considered an edge case. They could sneak a fix into a major since the release schedule is so much more frequent now, but not sure why they don't. 

This is precisely why tools like ESLint have plugins like `eslint-plugin-import` to assert that files exists at the location one is trying to require or import from. This of course, doesn't catch things like setting the entry file for a Rollup config, however.

> So I've found the actual cause of the issue. In my config, I have input: 'src/index.ts' which is the cause of the naming inconsistency.

This is really the crux; user error. (should not be read as an insult)

---

My thoughts are that this isn't really something that I'm keen on merging, because I think that case sensitivity is a good thing in the majority of cases. Paths are only normalized based on separator (https://github.com/rollup/plugins/blob/9b231e7fbe26bc4f4c7c1fc6e68f3ea3d64de920/packages/pluginutils/src/normalizePath.ts#L5) but not casing. 

That said, this might also be a cause of woe in the scenario for this PR https://github.com/rollup/plugins/blob/1a4f2bd4203e4cacc8be5908b513e2f1515d9cf6/packages/typescript/src/diagnostics/host.ts#L36-L38

glenn2223:
Sorry for the delay. Not a problem @shellscape, thanks for considering & no insult taken 😉 - we are all custom to user error (ESPECIALLY MINE!)

I needed it for a typescript extension I've been working on but since the issue has gone 🤷‍♂️

Issue #1253, opened 2022-09-02
@rollup/plugin-typescript  Incorrect DTS path


CroatiaParanoia:
- Rollup Plugin Name:  @Rollup/plugin-typescript
- Rollup Plugin Version: 8.4.0
- Rollup Version: 2.79.0
- Operating System (or Browser): macos
- Node Version: v16.16.0
- Link to reproduction _(⚠️ read below)_:

When I set the `declarationDir` value to `./types`, but the output path of the type file is still in `dist`

### Expected Behavior
Type files should be generated in `types`

### Actual Behavior
Type files are generated in `dist`

CroatiaParanoia:
https://github.com/rollup/plugins/pull/1252

danieleds:
As a temporary workaround, you can pin @rollup/plugin-typescript to v8.3.3

CroatiaParanoia:
> 作为临时解决方法，您可以将@rollup/plugin-typescript 固定到v8.3.3

CroatiaParanoia:
> As a temporary workaround, you can pin @rollup/plugin-typescript to v8.3.3

Thank you for your idea. I tried it, but it didn't work. now I plan to use tsc to build type files

danieleds:
Can we reopen this issue? The plugin is still broken.

CroatiaParanoia:
> Can we reopen this issue? The plugin is still broken.

Sorry, yes, I may have misoperated...

daiyam:
I'm having the issue with either `8.5.0` or `9.0.2`. `8.3.3` is working as expected.

susnux:
The workaround mentioned here does not work either:
https://github.com/rollup/plugins/tree/master/packages/typescript#declaration-output-with-outputfile

If using `.` then no declarations are generated at all.

susnux:
@CroatiaParanoia BTW if you plan to do something like this:
```
/dist/main.js
/types/main.d.ts
/src/main.ts
```

Meaning the output types are not within the same base directory as the js output, than this is **not** possible.
This is not a typescript plugin issue but a rollup limitation, [see this comment](https://github.com/rollup/plugins/pull/1378#discussion_r1131770253).

shellscape:
#1378 has been merged and should address this issue

Issue #1049, opened 2021-11-22
[typescript] declarationDir not working as expected


jpoehnelt:
- Rollup Plugin Name: typescript
- Rollup Plugin Version: 8.3.0
- Rollup Version: 2.60.0
- Operating System (or Browser): MacOS
- Node Version: v17.0.1
- Link to reproduction: https://replit.com/join/chxcoykjaf-jpoehnelt

### Expected Behavior
The typescript plugin places the typings in the same path as running tsc without rollup.

### Actual Behavior

The typings are nested incorrectly where `outDir` and `declaratationDir` are joined. These options are exclusive of one another.

```json
{
  "compilerOptions": {
    "declaration": true,
    "declarationDir": "./dist",
    "outDir": "./dist",
  },
  "include": ["src/**/*"]
}
```

`./dist/dist/index.ts` instead of `/dist/index.d.ts`

### Additional Information

Attempting to bypass with `declarataionDir: '.'` doesn't work because `.` is then excluded.

Copied from #978 
```
// tsconfig.json
{
  "compilerOptions": {
    "baseUrl": ".",
    "outDir": "./lib",
    "target": "esnext",
    "module": "esnext",
    "moduleResolution": "node",
    "allowJs": false,
    "strict": true,
    "noUnusedLocals": true,
    "esModuleInterop": true,
    "experimentalDecorators": true,
    "removeComments": false,
    "declaration": true,
    "declarationDir": "./lib",
    "declarationMap": false,
    "lib": [
      "esnext",
      "dom"
    ],
    "types": [
      "node",
    ]
  },
  "include": [
    "src/*"
  ]
}
```

```
// rollup.config.ts
export default {
  input: 'src/index.ts',
  output: [
    {
      file: 'lib/index.js',
      format: 'cjs',
    },
    {
      file: 'lib/index.esm.js',
      format: 'esm',
    },
    {
      file: 'lib/index.umd.js',
      format: 'umd',
      name: 'ShawUtil',
    },
  ],
  plugins: [
    nodeResolve(),
    commonjs(),
    babel({
      babelHelpers: 'bundled',
      exclude: 'node_modules/**',
    }),
    typescript({
      tsconfig: './tsconfig.json'
    }),
    eslint({
      include: ['src/**'],
      exclude: 'node_modules/**',
    })
  ]
};
```

shellscape:
Thanks for opening an issue. Citing the issue template:

[...]

PR #1378, opened 2022-12-15
feat(typescript): write declaration files in configured directory for `output.file`

susnux:

This PR contains:

- [x] bugfix
- [x] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

* #1049 (even if closed the issue was not resolved until now)
* #1253 (if `output.file` is used, if `output.dir` is used everything should, as it does, be placed inside that directory)

### Description
If you configure rollup with `output.file = 'dist/main.js'` and typscript with `'declarationDir: "dist"`
Instead of writing the declarations into `dist/` they  are written into `dist/dist`, and the workaround from the README does not work anymore, if you set 'declarationDir: "."` no declarations are written at all.

So this is fixes by only enforcing declarations inside `output.dir` but not inside `dirname(output.file)` (see added test case).

Additionally:
* I had to add the types package for `resolve` to get the package build.
* I added a new test helper that returns the output file names as rollup would write the files (emulating `handleGenerateWrite` and `writeOutputFile`)

### DIFF
--- a/packages/typescript/src/index.ts
+++ b/packages/typescript/src/index.ts
@@ -159,13 +159,16 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
             (parsedOptions.options.declaration
               ? parsedOptions.options.declarationDir || parsedOptions.options.outDir
               : null);
-          const cwd = normalizePath(process.cwd());
+          const outputDir = normalizePath(
+            outputOptions.dir ||
+              (outputOptions.file ? path.dirname(outputOptions.file) : process.cwd())
+          );
           if (
             parsedOptions.options.declaration &&
             parsedOptions.options.declarationDir &&
-            baseDir?.startsWith(cwd)
+            baseDir?.startsWith(outputDir)
           ) {
-            const declarationDir = baseDir.slice(cwd.length + 1);
+            const declarationDir = baseDir.slice(outputDir.length + 1);
             baseDir = baseDir.slice(0, -1 * declarationDir.length);
           }
           if (!baseDir && tsconfig) {
--- a/packages/typescript/src/index.ts
+++ b/packages/typescript/src/index.ts
@@ -154,27 +154,30 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
         const output = findTypescriptOutput(ts, parsedOptions, fileName, emittedFiles, tsCache);
         output.declarations.forEach((id) => {
           const code = getEmittedFile(id, emittedFiles, tsCache);
-          let baseDir =
-            outputOptions.dir ||
-            (parsedOptions.options.declaration
-              ? parsedOptions.options.declarationDir || parsedOptions.options.outDir
-              : null);
-          const outputDir = normalizePath(
-            outputOptions.dir ||
-              (outputOptions.file ? path.dirname(outputOptions.file) : process.cwd())
-          );
-          if (
-            parsedOptions.options.declaration &&
-            parsedOptions.options.declarationDir &&
-            baseDir?.startsWith(outputDir)
-          ) {
-            const declarationDir = baseDir.slice(outputDir.length + 1);
-            baseDir = baseDir.slice(0, -1 * declarationDir.length);
+          if (!code || !parsedOptions.options.declaration) {
+            return;
           }
-          if (!baseDir && tsconfig) {
-            baseDir = tsconfig.substring(0, tsconfig.lastIndexOf('/'));
+
+          let baseDir: string | undefined;
+          if (outputOptions.dir) {
+            baseDir = outputOptions.dir;
+          } else if (outputOptions.file) {
+            // find common path of output.file and configured declation output
+            const outputDir = path.dirname(outputOptions.file);
+            const configured = path.resolve(
+              parsedOptions.options.declarationDir ||
+                parsedOptions.options.outDir ||
+                tsconfig ||
+                process.cwd()
+            );
+            const backwards = path
+              .relative(outputDir, configured)
+              .split(path.sep)
+              .filter((v) => v === '..')
+              .join(path.sep);
+            baseDir = path.normalize(`${outputDir}/${backwards}`);
           }
-          if (!code || !baseDir) return;
+          if (!baseDir) return;
 
           this.emitFile({
             type: 'asset',
--- a/packages/typescript/test/test.js
+++ b/packages/typescript/test/test.js
@@ -6,7 +6,7 @@ const test = require('ava');
 const { rollup, watch } = require('rollup');
 const ts = require('typescript');
 
-const { evaluateBundle, getCode, onwarn } = require('../../../util/test');
+const { evaluateBundle, getCode, getFiles, onwarn } = require('../../../util/test');
 
 const typescript = require('..');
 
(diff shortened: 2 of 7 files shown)


shellscape:
Please see the Validate Monorepo workflow result. In a monorepo you cannot add a dependency manually without also running `pnpm install` at the root.

susnux:
> Please see the Validate Monorepo workflow result. In a monorepo you cannot add a dependency manually without also running `pnpm install` at the root.

Oh thank you! Should be fixed now :)

packages/typescript/src/index.ts
```diff
@@ -158,13 +158,17 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
             (parsedOptions.options.declaration
               ? parsedOptions.options.declarationDir || parsedOptions.options.outDir
               : null);
-          const cwd = normalizePath(process.cwd());
+          const outputDir = normalizePath(
+            outputOptions.dir || !outputOptions.file
+              ? process.cwd()
```
NotWoods, line 163: Shouldn't the output dir be `outputOptions.dir`?

packages/typescript/test/test.js
```diff
@@ -139,39 +139,66 @@ test.serial('supports emitting types also for single file output', async (t) =>
   // as that would have the side effect that the tsconfig's path would be used as fallback path for
   // the here unspecified outputOptions.dir, in which case the original issue wouldn't show.
   process.chdir('fixtures/basic');
+  const outputOpts = { format: 'es', dir: 'dist' };
```
NotWoods, line 142: Since a `dir` is now being used, this test no longer tests single file output

packages/typescript/test/test.js
```diff
@@ -139,39 +139,66 @@ test.serial('supports emitting types also for single file output', async (t) =>
...
-    ['main.js', 'dist/main.d.ts']
+    ['dist/main.js', 'dist/main.d.ts']
+  );
+  t.is(warnings.length, 0);
+});
+
+test.serial('supports emitting declarations in correct directory for output.file', async (t) => {
+  // Ensure even when no `output.dir` is configured, declarations are emitted to configured `declarationDir`
+  process.chdir('fixtures/basic');
+  const outputOpts = { format: 'es', file: 'dist/main.esm.js' };
```
NotWoods, line 166: I'd like to also see a test where the output file doesn't match the declaration dir

packages/typescript/test/test.js
```diff
@@ -139,39 +139,66 @@ test.serial('supports emitting types also for single file output', async (t) =>
...
+    output: outputOpts,
     plugins: [typescript({ tsconfig: 'fixtures/relative-dir/tsconfig.json' })],
     onwarn
   });
-  const output = await getCode(bundle, { format: 'es', dir: 'fixtures/relative-dir/dist' }, true);
+  const output = await getFiles(bundle, outputOpts);
 
   t.deepEqual(
-    output.map((out) => out.fileName),
+    output.map((out) => out.fileName.replace(`${outputOpts.dir}/`, '')),
```
NotWoods, line 197: Rather than stripping the dir, the test expected output should be changed instead

util/test.js
```diff
@@ -7,13 +10,37 @@ const getCode = async (bundle, outputOptions, allFiles = false) => {
...
   return code;
 };
 
+/**
+ * @param {import('rollup').RollupBuild} bundle
+ * @param {import('rollup').OutputOptions} [outputOptions]
+ */
+const getFiles = async (bundle, outputOptions) => {
+  if (!outputOptions.dir && !outputOptions.file)
+    throw new Error('You must specify "output.file" or "output.dir" for the build.');
```
NotWoods, line 31: If `output.file` isn't specified, can we figure it out based on input? Since Rollup doesn't require this option I want to make sure our tests support that case too

packages/typescript/src/index.ts
susnux: If `outputOptions.dir` is set, then *it is used*.

If `outputOptions.dir` is **not** set, then - if `outputOptions.file` is set, the base directory of it is used, if not then the current working dir is used.

util/test.js
susnux: Well rollup requires either `output.file` or `output.dir` so throwing is valid, moreover this is just copied from rollup.
We do not write files so this emulates the rollup code for writing assets, see also this line:
https://github.com/rollup/rollup/blob/b179695aab154bcb9b7e35e0a16f9afd7f1264f1/src/rollup/rollup.ts#L166

packages/typescript/src/index.ts
NotWoods: I see. Is the execution like this?
```js
let dir = outputOptions.dir;

if (!dir) {
  dir = !outputOptions.file ? process.cwd() : path.dirname(outputOptions.file)
}

const outputDir = normalizePath(dir);
```

If so, it would be nice to have parenthesis around the latter `!outputOptions.file ? process.cwd() : path.dirname(outputOptions.file)` part so the code execution is clearer. If prettier removes it, then you can ignore this comment.

as a nit, I would prefer to write `outputOptions.file ? path.dirname(outputOptions.file) : process.cwd()` rather than `!outputOptions.file ? process.cwd() : path.dirname(outputOptions.file)`

shellscape:
Please note that we'll close this as stale if the PR hasn't received any attention from the author in 60 days.

susnux:
My problem is: I do not have a Windows machine to test this PR so I will have to blind guess how to fix the windows node tests

shellscape:
I understand. This looks like path differences between windows and nix. I'll see if we can't get you some help for this.

vighnesh153:
Is it possible to verify on docker windows image? https://hub.docker.com/_/microsoft-windows

halfmatthalfcat:
So I took these changes for a test drive in my Vite library project and was still having problems. What I also changed to fix things was to replace `baseDir` with `outputDir` on L181. My base dir was resolving to `.` which was causing the eventual `fileName` to still include the `dist` in my output path resulting in the nested `dist/dist/...` we've been experiencing.

packages/typescript/test/test.js
susnux: What do you mean? Something like this?:
```
output.file = 'dist/main.js'
declarationDir = 'types'
```

This will always result in `dist/types/main.d.ts`, because rollup does not allow files outside the `output.dir` or the base directory of the `output.file`, see this [rollups source](https://github.com/rollup/rollup/blob/bceb3539bab95ba92688f29a7c9f3e9effc44698/src/rollup/rollup.ts#L268).

susnux:
@shellscape @NotWoods I think I fixed the issues, can you trigger the CI?

susnux:
Yes that would be really helpful, but I might also have fixed that issue as a rewrote the code while fixing the issues.

shellscape:
I think we're good to go. Thanks for working through this with the team.

rm116:
Hello @susnux , 

If I have this rollup config : 

```
export default [
  {
    input: "src/index.ts",
    output: [
      {
        file: packageJson.main,
        format: "cjs",
        sourcemap: true,
      },
      {
        file: packageJson.module,
        format: "esm",
        sourcemap: true,
      },
    ],
    plugins: [
      peerDepsExternal(),
      resolve({
        browser: true
      }),
      commonjs(),
      typescript({ tsconfig: "./tsconfig.json" }),
      json(),
      postcss(),
      terser(),
    ],
  },
  {
    input: "lib/esm/types/index.d.ts",
    output: [{ file: "lib/index.d.ts", format: "esm" }],
    plugins: [dts()],
    external: ['react', 'react-dom', /\.css$/]
  },
];
```

And this package.json : 

```
{
  "name": "test",
  "version": "0.0.1",
  "description": "",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1",
    "rollup": "rollup -c"
  },
  "main": "lib/cjs/index.js",
  "module": "lib/esm/index.js",
  "files": [
    "lib"
  ],
  "types": "lib/index.d.ts",
  "devDependencies": {
    "@rollup/plugin-commonjs": "^22.0.1",
    "@rollup/plugin-json": "^4.1.0",
    "@rollup/plugin-node-resolve": "^13.3.0",
    "@rollup/plugin-typescript": "^8.3.3",
    "@types/react": "^18.0.15",
    "postcss": "^8.4.14",
    "rollup": "^2.77.0",
    "rollup-plugin-dts": "^4.2.2",
    "rollup-plugin-ignore": "^1.0.10",
    "rollup-plugin-peer-deps-external": "^2.2.4",
    "rollup-plugin-postcss": "^4.0.2",
    "rollup-plugin-terser": "^7.0.2",
    "tslib": "^2.4.0",
    "typescript": "^4.7.4"
  },
  "dependencies": {
    "axios": "^0.27.2"
  },
  "peerDependencies": {
    "react": "^17.0.2",
    "react-dom": "^17.0.2"
  }
}
```

And this tsconfig.json : 

```
{
  "compilerOptions": {
    "target": "es5", 
    "esModuleInterop": true, 
    "forceConsistentCasingInFileNames": true,
    "strict": true, 
    "skipLibCheck": true,
    "baseUrl": "./src",
    "paths": {
      "@services/*": ["services/*"],
      "@root/*": ["./*"]
    },
    "jsx": "react", 
    "module": "ESNext",  
    "declaration": true,
    "declarationDir": "types",
    "sourceMap": true,
    "outDir": "lib",
    "moduleResolution": "node",
    "allowSyntheticDefaultImports": true,
    "emitDeclarationOnly": true,
  }
}
```

The rollup command is failing because it didn't find the file : "lib/esm/types/index.d.ts"

image

rm116:
cc @NotWoods

susnux:
@rm116 You should upgrade `@rollup/plugin-typescrip` as you still use version 8 and this PR is only included in version 11.1.0+.

Using the latest version everything from your example works for me.

PR #1389, opened 2022-12-25
fix(eslint): fix async support for `formatter` option and TypeScript types

surmon-china:

This PR contains:

- [x] bugfix

Are tests included?

Breaking Changes?

- [x] no

### Description

1. ESLint's [`CLIEngine` class has been deprecated](https://eslint.org/docs/latest/user-guide/migrating-to-7.0.0#-the-cliengine-class-has-been-deprecated), so the latest `@types/eslint` 8.x no longer exports the `CLIEngine` type, so the latest version of `@rollup/plugin-eslint` formatter is inferred to have the wrong argument type.

2. The latest type of `ESLint.Formatter` allows to return async values, while the current `@rollup/plugin-eslint` only supports sync function.
    - [Change Request: Support async formatter](https://github.com/eslint/eslint/issues/15242)
    - [feat: support async formatters](https://github.com/eslint/eslint/commit/f1b7499a5162d3be918328ce496eb80692353a5a)

### DIFF
--- a/packages/eslint/README.md
+++ b/packages/eslint/README.md
@@ -67,6 +67,8 @@ Default: `stylish`
 
 Custom error formatter, the name of a built-in formatter, or the path to a custom formatter.
 
+When the type is a function, the function should return `String` or `Promise<String>`.
+
 ### include
 
 Type: `String | String[]`<br>
--- a/packages/eslint/README.md
+++ b/packages/eslint/README.md
@@ -62,12 +62,10 @@ If true, will auto fix source code.
 
 ### formatter
 
-Type: `Function | String`<br>
+Type: `Function<String> | Function<Promise<String>> | String`<br>
 Default: `stylish`
 
 Custom error formatter, the name of a built-in formatter, or the path to a custom formatter.
-
-When the type is a function, the function should return `String` or `Promise<String>`.
 
 ### include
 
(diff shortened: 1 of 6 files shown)


surmon-china:
@shellscape

shellscape:
You just put your PR up yesterday. Please exercise patience. Much of the US is on vacation with their families right now.

packages/eslint/types/index.d.ts
```diff
@@ -37,7 +37,7 @@ export interface RollupEslintOptions extends ESLint.Options {
    * Custom error formatter or the name of a built-in formatter.
    * @default stylish
    */
-  formatter?: CLIEngine.Formatter | string;
+  formatter?: Awaited<ReturnType<ESLint['loadFormatter']>>['format'] | string;
```
shellscape, line 40: the docs currently say that a regular `Function` is valid here, and it still technically is even with `await`. so the docs need to be updated to reflect this change. docs should also reflect that the return value of the function should be `String | Promise<String>`

Given that the last update to the plugin missed the broken types, it would be great (but not required for this PR) to add type tests (other plugins have these) or to update the tests to use TypeScript. again, not required for approval but would be a big improvement.

shellscape (changes requested):

packages/eslint/README.md
```diff
@@ -67,6 +67,8 @@ Default: `stylish`
 
 Custom error formatter, the name of a built-in formatter, or the path to a custom formatter.
 
+When the type is a function, the function should return `String` or `Promise<String>`.
```
shellscape, line 70: This isn't a necessary line. Line 65 should look like:

```
Type: `Function<String> | Function<Promise<String>> | String`<br>
```

shellscape (approved):

shellscape:
thanks!

PR #1393, opened 2023-01-05
Change commonjs strictRequires default from "auto" to true

birkskyum:

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

https://github.com/rollup/rollup/issues/4787

### Description

Having strictRequires: true is a safer default option than "auto", as auto can break in cases where **true** will not (see linked issue). Minification is omnipresent at this point, so it's not an issue to go with this option in terms of filesize.

### DIFF
--- a/packages/commonjs/README.md
+++ b/packages/commonjs/README.md
@@ -49,7 +49,7 @@ When used together with the node-resolve plugin
 ### `strictRequires`
 
 Type: `"auto" | boolean | "debug" | string[]`<br>
-Default: `"auto"`
+Default: `true`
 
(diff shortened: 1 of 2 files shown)


shellscape:
This will need significant support from the community at large to change, since this is a proposed breaking change based on opinion.

birkskyum:
@shellscape , if it's breaking, then absolutely. It wasn't my impression that it would be breaking.

shellscape:
changing the default surface layer of an API is always a breaking change. this has potential to be an issue for communities like Vite, in addition to those people who use rollup directly

birkskyum:
I see. My intention was to make it easier for people using vite / vite-derived meta-frameworks, as this is necessary otherwise:

https://github.com/solidjs/solid-start/pull/626

shellscape:
I would recommend closing this PR and opening a discussion with Vite to add that option value by default within Vite. If that project adopts that setting, it's more compelling to release a new major version with that change here. (That would amount to wide community support)

PR #1394, opened 2023-01-06
feat(terser): emit worker error message (+stack) to parent instance

tada5hi:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -3,4 +3,5 @@
 /packages/babel/ @Andarist
 /packages/image/ @tjenkinson
 /packages/node-resolve/ @tjenkinson
+/packages/swc/ @tada5hi
 /packages/terser/ @tada5hi
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ This repository houses plugins that Rollup considers critical to every day use o
 | [run](packages/run)                                 | Run your bundles in Node once they're built                                               |
 | [strip](packages/strip)                             | Remove debugger statements and functions like assert.equal and console.log from your code |
 | [sucrase](packages/sucrase)                         | Compile TypeScript, Flow, JSX, etc with Sucrase                                           |
+| [terser](packages/swc)                              | Transpile TypeScript/JavaScript with the speedy-web-compiler                              |
 | [terser](packages/terser)                           | Generate a minified output bundle with terser                                             |
 | [typescript](packages/typescript)                   | Integration between Rollup and Typescript                                                 |
 | [url](packages/url)                                 | Import files as data-URIs or ES Modules                                                   |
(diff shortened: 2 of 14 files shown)


tada5hi:
@shellscape would be great if you could have a look :v:

shellscape:
I'm codeowner on the whole repo so I get email notifications of everything.

tada5hi:
I will remember that ^^. Did you see the change in the last comment?

tada5hi:
@shellscape is something missing?

packages/terser/src/type.ts
```diff
@@ -23,6 +23,12 @@ export interface WorkerOutput {
   sourceMap?: Record<string, any>;
 }
 
+export interface WorkerError {
```
shellscape, line 26: this is pretty loose. being more constrained would probably be a good idea. Adding an `error` property to `WorkerOutput` is more semantic. `error` can be an object with `message` and `stack`, and then all you need do is check if `data.error` is truthy on line 111. that will remove some complexity I don't think is needed.

tada5hi: @shellscape but that would imply that all other properties of the `WorkerOutput` must be nullable, in case an error occoured and the error property is set and than it would be harder to distinguish between a success and error state.

shellscape: not sure I follow:

`error` is nullish: no error
`error` is truthy: error

tada5hi: sure, but the `code` property e.g. could then no longer be a mandatory property.

packages/terser/test/test.js
```diff
@@ -122,7 +122,7 @@ test.serial('throw error on terser fail', async (t) => {
     await bundle.generate({ format: 'esm' });
     t.falsy(true);
   } catch (error) {
-    t.is(error.toString(), 'Error: Minify worker stopped with exit code 1');
+    t.is(error.toString(), 'Error: Name expected');
```
dasa, line 125: PR #1409 also improves this error message.

packages/terser/src/type.ts
shellscape: yes. why is that a concern?

tada5hi: I like to map different states using different types and then merge them with a union type if necessary :innocent: .
Even with callbacks, there are also two parameters (error, result), instead of the error being an optional property of the result parameter. The same applies to promises. But I have no problem doing it the way you suggested, if you prefer :v: :relaxed: .

PR #1402, opened 2023-01-13
feat(alias)!: resolve entry id

### LINKED ISSUE #1190: Cannot use an alias as an `input`
- Rollup Plugin Name: @rollup/plugin-alias
- Rollup Plugin Version: 3.1.9
- Rollup Version: 2.74.0
- Operating System (or Browser): OSX
- Node Version: 16
- Link to reproduction _(⚠️ read below)_: https://replit.com/@matthewp17/rollup-plugin-alias-bug

### Expected Behavior

You should be able to pass an alias as an `input`.

### Actual Behavior

An error for not being able to resolve the alias is shown in the console.

### Additional Information

This is the code that causes the bug: https://github.com/rollup/plugins/blob/36de7cc9c3915667bc0dd203ee7b139da8858959/packages/alias/src/index.ts#L79-L81

Was introduced in this PR: https://github.com/rollup/plugins/pull/1024

Doesn't seem like the PR was intentionally trying to make an alias as an input not be allowed, so I think this part was a mistake.

bluwy:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] no

If yes, then include "BREAKING CHANGES:" in the first commit message body, followed by a description of what is breaking. - Added

fix #1190

### Description

When the Rollup input is passed an aliased path, the `alias` plugin doesn't resolve it because it skips if there's no `importer`. I tracked this behaviour, and found it was not done in the past due to https://github.com/rollup/rollup-plugin-alias/pull/29, but since the internals changed now, that PR's issue isn't relevant and we can continue to alias entry paths now.

### DIFF
--- a/packages/alias/src/index.ts
+++ b/packages/alias/src/index.ts
@@ -81,9 +81,6 @@ export default function alias(options: RollupAliasOptions = {}): Plugin {
       );
     },
     resolveId(importee, importer, resolveOptions) {
-      if (!importer) {
-        return null;
-      }
       // First match is supposed to be the correct one
       const matchedEntry = entries.find((entry) => matches(entry.find, importee));
       if (!matchedEntry) {
--- a/packages/alias/test/test.mjs
+++ b/packages/alias/test/test.mjs
@@ -163,14 +163,14 @@ test('Will not confuse modules with similar names', (t) =>
     ]
   ).then((result) => t.deepEqual(result, [null, null, null])));
 
-test('Leaves entry file untouched if matches alias', (t) =>
+test('Alias entry file', (t) =>
   resolveAliasWithRollup(
     {
       entries: [{ find: 'abacaxi', replacement: './abacaxi' }]
     },
     // eslint-disable-next-line no-undefined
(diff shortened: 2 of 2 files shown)


bluwy:
My bad. Didn't notice my syntax in the other PR would also close this. Re-opened and it should still be good to go.

shellscape:
> but since the internals changed now

could you link to the internals you mentioned?

bluwy:
The old plugin code used to rely on the `importer` at https://github.com/rollup/rollup-plugin-alias/blob/0204677c0e1112cba7a5a0f60204ff1df81493b0/src/index.js#L80. Now we're not doing that anymore (we don't use the `path` nodejs module either). We're only passing the `importer` to the `customResolver`, which is currently correctly typed that `importer` could be defined.

shellscape (approved):
Makes sense, thanks. I'd like to get another approval from another maintainer before we merge.

PR #1410, opened 2023-01-22
refactor: prefix plugins names with `rollup:`

antfu:
## Rollup Plugin Name: All of them

This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [ ] no

### Description

This is a proposal.

While Rollup becomes bigger to have Vite and its ecosystems (meta frameworks on top of vite etc.), I personally think it could be great to endorse the conventions of adding namespaces to the plugin's names. For example, when we print the plugins names from Nuxt, we have:

image

Since both Vite and Nuxt uses prefix on the plugins, it's clear to tell which plugins are from where. While you see `replace` and `virtual` are unprefixed, it may be confusing the tell where they are from.

While I think it's OK for Rollup to treat the official rollup plugins specially without the prefix. I think having the name prefix would also help to distinguish from users' local plugins (which I imaging most of the time could be unprefixed).

This **can** be a breaking change, but only for some plugins authors or advanced usage, but overall I consider this worth it, and it also brings more consistency to the repo (e.g. `rollup-plugin-dynamic-import-variables` to `rollup:dynamic-import-vars`, matching its package name, `dataUri` to `rollup:data-uri` for cases).


shellscape:
Thanks for the proposal. We have a feature and modification issue template for discussion on these types of changes before work is done on them, so that folks don't spend time on new work in the event that a PR isn't accepted.

For me personally, I don't see this as a need. Rollup is used more than just within frontend-focused ecosystems like Vite and Nuxt, and an arbitrary spec that a FE ecosystem has defined isn't compelling enough for me to approve a breaking change across all plugins. If there were an issue with internal plugin identification and confusion, I'm sure it would have come up in the last 6+ years.

For now, I don't see this as a need. Let's see what the other maintainers think.

antfu:
Since it's a straightforward change, I don't actually mind if this PR eventually didn't get accepted. I don't have high hope for this either, but I just want to bring it up to the table. We could use this PR as the thread for discussion.

While I agree with Rollup itself, this is probably not necessary, I also see Vite and the ecosystem pushing Rollup to improve to some extent. I honestly don't think this is an FE-specific thing, but instead, having a convention would help the plugin/framework in the long term. E.g. [tsup](https://github.com/egoist/tsup) and [unbuild](https://github.com/unjs/unbuild) are some examples of tools built on top of Rollup for libraries. For example, without the namespace, https://github.com/unjs/unbuild/blob/main/src/builder/plugins/esbuild.ts#L67 could be confusing from the runtime to know if the `name: 'esbuild'` is from `@rollup/plugin-esbuild`, `rolluo-plugin-esbuild`, `vite-plugin-esbuild` or just the local user plugin.

This is not something we have to have, but I think it could benefit the ecosystem in the long term (we could update the docs if this ever get accepted to make it a official convention)

shellscape:
I see what you're saying. But conventions require enforcement via specifications and implementation. There's nothing stopping `rollup-plugin-myvite-foo` from using `vite:foo` while `@vite/plugin-foo` uses `vite:foo`

I'd start in the main rollup repo with a discussion on an enforceable spec for this. If core adopts something, then the plugins would follow suit.

antfu:
Yeah, that makes sense. Opened at https://github.com/rollup/rollup/issues/4817

tada5hi:
> Thanks for the proposal. We have a feature and modification issue template for discussion on these types of changes before work is done on them, so that folks don't spend time on new work in the event that a PR isn't accepted.
> 
> 
> 
> For me personally, I don't see this as a need. Rollup is used more than just within frontend-focused ecosystems like Vite and Nuxt, and an arbitrary spec that a FE ecosystem has defined isn't compelling enough for me to approve a breaking change across all plugins. If there were an issue with internal plugin identification and confusion, I'm sure it would have come up in the last 6+ years.
> 
> 
> 
> For now, I don't see this as a need. Let's see what the other maintainers think.

I couldn't agree more with what Andrew said.
In general I like the idea 💡 of domain/namespace prefixes to provide a clearer differentiation of publishers. But that shouldn't necessarily be the case for official plugins. Moreover, in general this prefix can't be verified either. 
Verification would also only be possible if, for example, you no longer allow on-the-fly defined modules and the plugin name reflects the name in the package.json file.

shellscape:
I'm going to mark this as waiting for upstream until the rollup issue mentioned above is resolved

shellscape:
Closing for housekeeping as their hasn't been much movement on the upstream issue.

PR #1412, opened 2023-01-23
feat(swc): initial release

tada5hi:
## Rollup Plugin Name: `swc`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

https://github.com/rollup/plugins/issues/1413

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ This repository houses plugins that Rollup considers critical to every day use o
 | [run](packages/run)                                 | Run your bundles in Node once they're built                                               |
 | [strip](packages/strip)                             | Remove debugger statements and functions like assert.equal and console.log from your code |
 | [sucrase](packages/sucrase)                         | Compile TypeScript, Flow, JSX, etc with Sucrase                                           |
-| [terser](packages/swc)                              | Transpile TypeScript/JavaScript with the speedy-web-compiler                              |
+| [swc](packages/swc)                                 | Transpile TypeScript/JavaScript with the speedy-web-compiler                              |
 | [terser](packages/terser)                           | Generate a minified output bundle with terser                                             |
 | [typescript](packages/typescript)                   | Integration between Rollup and Typescript                                                 |
 | [url](packages/url)                                 | Import files as data-URIs or ES Modules                                                   |
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ This repository houses plugins that Rollup considers critical to every day use o
 | [run](packages/run)                                 | Run your bundles in Node once they're built                                               |
 | [strip](packages/strip)                             | Remove debugger statements and functions like assert.equal and console.log from your code |
 | [sucrase](packages/sucrase)                         | Compile TypeScript, Flow, JSX, etc with Sucrase                                           |
+| [terser](packages/swc)                              | Transpile TypeScript/JavaScript with the speedy-web-compiler                              |
 | [terser](packages/terser)                           | Generate a minified output bundle with terser                                             |
 | [typescript](packages/typescript)                   | Integration between Rollup and Typescript                                                 |
 | [url](packages/url)                                 | Import files as data-URIs or ES Modules                                                   |
--- a/packages/swc/README.md
+++ b/packages/swc/README.md
@@ -12,7 +12,7 @@
 🍣 A Rollup plugin to transpile TypeScript/JavaScript with the speedy-web-compiler (swc).
 
 The plugin makes it possible to avoid the usage of `@rollup/plugin-babel` and `@rollup/plugin-typescript`.
-It is also blazingly fast 🔥 ( between 20 - 70 times faster than Babel depending on how many cpu cores are used).
+It is also blazingly fast 🔥 (between 20 - 70 times faster than Babel depending on how many cpu cores are used).
 
 ## Requirements
 
@@ -61,7 +61,7 @@ declare type Options = {
 ## Alternatives
 
 Alternative transpiler/transformer, which are also officially offered, are [Babel](https://www.npmjs.com/package/@rollup/plugin-babel) and [Sucrase](https://www.npmjs.com/package/@rollup/plugin-sucrase).
-Both Babel and Sucrase are written in JavaScript, whereas the core of Swc is written in Rust. In addition, Sucrase transforms code to ES6+, whereas babel and SWC also support ES3 upwards.
+Both Babel and Sucrase are written in JavaScript, whereas the core of SWC is written in Rust. In addition, Sucrase transforms code to ES6+, whereas babel and SWC also support ES3 upwards.
 
 ## Meta
 
--- a/packages/swc/README.md
+++ b/packages/swc/README.md
@@ -50,6 +50,11 @@ Then call `rollup` either via the [CLI](https://www.rollupjs.org/guide/en/#comma
 The plugin accepts a swc [Options](https://swc.rs/docs/configuration/swcrc) object as input parameter,
 to modify the default behaviour.
 
+## Alternatives
+
+Alternative transpiler/transformer, which are also officially offered, are [Babel](https://www.npmjs.com/package/@rollup/plugin-babel) and [Sucrase](https://www.npmjs.com/package/@rollup/plugin-sucrase).
+Both Babel and Sucrase are written in JavaScript, whereas the core of Swc is written in Rust. In addition, Sucrase transforms code to ES6+, whereas babel and SWC also support ES3 upwards.
+
 ## Meta
 
 [CONTRIBUTING](/.github/CONTRIBUTING.md)
--- a/packages/swc/README.md
+++ b/packages/swc/README.md
@@ -1,0 +1,57 @@
+[npm]: https://img.shields.io/npm/v/@rollup/plugin-swc
+[npm-url]: https://www.npmjs.com/package/@rollup/plugin-swc
+[size]: https://packagephobia.now.sh/badge?p=@rollup/plugin-swc
+[size-url]: https://packagephobia.now.sh/result?p=@rollup/plugin-swc
+
+[![npm][npm]][npm-url]
+[![size][size]][size-url]
+[![libera manifesto](https://img.shields.io/badge/libera-manifesto-lightgrey.svg)](https://liberamanifesto.com)
+
+# @rollup/plugin-swc
+
+🍣 A Rollup plugin to transpile TypeScript/JavaScript with the speedy-web-compiler (swc).
+
+The plugin makes it possible to avoid the usage of `@rollup/plugin-babel` and `@rollup/plugin-typescript`.
+It is also blazingly fast 🔥 ( between 20 - 70 times faster than babel depending on how many cpu cores are used).
+
+## Requirements
+
+This plugin requires an [LTS](https://github.com/nodejs/Release) Node version (v10.0.0+) and Rollup v2.0+.
+
+## Install
+
+Using npm:
+
+```console
+npm install @rollup/plugin-swc --save-dev
+```
+
+## Usage
+
+Create a `rollup.config.js` [configuration file](https://www.rollupjs.org/guide/en/#configuration-files) and import the plugin:
+
+```typescript
+import swc from '@rollup/plugin-swc';
+
+export default {
+  input: 'src/index.js',
+  output: {
+    dir: 'output',
+    format: 'cjs'
+  },
+  plugins: [swc()]
+};
+```
+
+Then call `rollup` either via the [CLI](https://www.rollupjs.org/guide/en/#command-line-reference) or the [API](https://www.rollupjs.org/guide/en/#javascript-api).
+
+## Options
+
+The plugin accepts a swc [Options](https://swc.rs/docs/configuration/swcrc) object as input parameter,
+to modify the default behaviour.
+
+## Meta
+
+[CONTRIBUTING](/.github/CONTRIBUTING.md)
+
+[LICENSE (MIT)](/LICENSE)
--- a/packages/swc/README.md
+++ b/packages/swc/README.md
@@ -16,7 +16,7 @@ It is also blazingly fast 🔥 (between 20 - 70 times faster than Babel dependin
 
 ## Requirements
 
-This plugin requires an [LTS](https://github.com/nodejs/Release) Node version (v10.0.0+) and Rollup v2.0+.
+This plugin requires an [LTS](https://github.com/nodejs/Release) Node version (v10.0.0+) and Rollup v3.0+.
 
 ## Install
 
--- a/packages/swc/README.md
+++ b/packages/swc/README.md
@@ -12,7 +12,7 @@
 🍣 A Rollup plugin to transpile TypeScript/JavaScript with the speedy-web-compiler (swc).
 
 The plugin makes it possible to avoid the usage of `@rollup/plugin-babel` and `@rollup/plugin-typescript`.
-It is also blazingly fast 🔥 ( between 20 - 70 times faster than babel depending on how many cpu cores are used).
+It is also blazingly fast 🔥 ( between 20 - 70 times faster than Babel depending on how many cpu cores are used).
 
 ## Requirements
 
--- a/packages/swc/README.md
+++ b/packages/swc/README.md
@@ -16,7 +16,7 @@ It is also blazingly fast 🔥 (between 20 - 70 times faster than Babel dependin
 
 ## Requirements
 
-This plugin requires an [LTS](https://github.com/nodejs/Release) Node version (v10.0.0+) and Rollup v3.0+.
+This plugin requires an [LTS](https://github.com/nodejs/Release) Node version (v14.0.0+) and Rollup v3.0+.
 
 ## Install
 
--- a/packages/swc/README.md
+++ b/packages/swc/README.md
@@ -47,8 +47,16 @@ Then call `rollup` either via the [CLI](https://www.rollupjs.org/guide/en/#comma
 
 ## Options
 
-The plugin accepts a swc [Options](https://swc.rs/docs/configuration/swcrc) object as input parameter,
-to modify the default behaviour.
+The plugin accepts an object as input parameter to modify the default behaviour.
+This object has an `swc` property to provide library specific [Options](https://swc.rs/docs/configuration/swcrc).
+
+```typescript
+import type { Options as SWCOptions } from '@swc/core';
+
+declare type Options = {
+  swc?: SWCOptions;
+};
+```
 
 ## Alternatives
 
--- a/packages/swc/package.json
+++ b/packages/swc/package.json
@@ -53,7 +53,8 @@
     "modules"
   ],
   "peerDependencies": {
-    "rollup": "^2.x || ^3.x"
+    "@swc/core": "^1.x",
+    "rollup": "^3.x"
   },
   "peerDependenciesMeta": {
     "rollup": {
@@ -61,10 +62,10 @@
     }
   },
   "dependencies": {
-    "@swc/core": "^1.3.27",
     "smob": "^0.0.6"
   },
   "devDependencies": {
+    "@swc/core": "^1.3.27",
     "rollup": "^3.0.0-7",
     "typescript": "^4.8.3"
   },
--- a/packages/swc/package.json
+++ b/packages/swc/package.json
@@ -53,7 +53,6 @@
     "modules"
   ],
   "peerDependencies": {
-    "@swc/core": "^1.x",
     "rollup": "^2.x || ^3.x"
   },
   "peerDependenciesMeta": {
@@ -62,10 +61,10 @@
     }
   },
   "dependencies": {
+    "@swc/core": "^1.3.27",
     "smob": "^0.0.6"
   },
   "devDependencies": {
-    "@swc/core": "^1.3.27",
     "rollup": "^3.0.0-7",
     "typescript": "^4.8.3"
   },
--- a/packages/swc/package.json
+++ b/packages/swc/package.json
@@ -1,0 +1,73 @@
+{
+  "name": "@rollup/plugin-swc",
+  "version": "0.0.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "description": "Transpile JavaScript/TypeScript code with swc.",
+  "license": "MIT",
+  "repository": {
+    "url": "rollup/plugins",
+    "directory": "packages/swc"
+  },
+  "author": "Peter Placzek <peter.placzek1996@gmail.com>",
+  "homepage": "https://github.com/rollup/plugins/tree/master/packages/swc#readme",
+  "bugs": "https://github.com/rollup/plugins/issues",
+  "main": "dist/cjs/index.js",
+  "module": "dist/es/index.js",
+  "exports": {
+    "types": "./types/index.d.ts",
+    "import": "./dist/es/index.js",
+    "default": "./dist/cjs/index.js"
+  },
+  "engines": {
+    "node": ">=10.0.0"
+  },
+  "scripts": {
+    "build": "rollup -c",
+    "ci:coverage": "nyc pnpm test && nyc report --reporter=text-lcov > coverage.lcov",
+    "ci:lint": "pnpm build && pnpm lint",
+    "ci:lint:commits": "commitlint --from=${CIRCLE_BRANCH} --to=${CIRCLE_SHA1}",
+    "ci:test": "pnpm test -- --verbose && pnpm test:ts",
+    "prebuild": "del-cli dist",
+    "prepare": "if [ ! -d 'dist' ]; then pnpm build; fi",
+    "prerelease": "pnpm build",
+    "pretest": "pnpm build",
+    "release": "pnpm --workspace-root plugin:release --pkg $npm_package_name",
+    "test": "ava",
+    "test:ts": "tsc types/index.d.ts test/types.ts --noEmit"
+  },
+  "files": [
+    "dist",
+    "!dist/**/*.map",
+    "src",
+    "types",
+    "README.md"
+  ],
+  "keywords": [
+    "rollup",
+    "plugin",
+    "swc",
+    "speedy-web-compiler",
+    "npm",
+    "modules"
+  ],
+  "peerDependencies": {
+    "@swc/core": "^1.x",
+    "rollup": "^2.x || ^3.x"
+  },
+  "peerDependenciesMeta": {
+    "rollup": {
+      "optional": true
+    }
+  },
+  "dependencies": {
+    "smob": "^0.0.6"
+  },
+  "devDependencies": {
+    "@swc/core": "^1.3.27",
+    "rollup": "^3.0.0-7",
+    "typescript": "^4.8.3"
+  },
+  "types": "./types/index.d.ts"
+}
--- a/packages/swc/package.json
+++ b/packages/swc/package.json
@@ -21,7 +21,7 @@
     "default": "./dist/cjs/index.js"
   },
   "engines": {
-    "node": ">=10.0.0"
+    "node": ">=14.0.0"
   },
   "scripts": {
     "build": "rollup -c",
--- a/packages/swc/src/module.ts
+++ b/packages/swc/src/module.ts
@@ -1,6 +1,7 @@
-import { Plugin } from 'rollup';
+import type { Plugin } from 'rollup';
 import { transform } from '@swc/core';
 import { merge } from 'smob';
+
 import type { Options } from './type';
 
 export function swc(input: Options = {}): Plugin {
--- a/packages/swc/src/module.ts
+++ b/packages/swc/src/module.ts
@@ -24,7 +24,7 @@ export function swc(input: Options = {}): Plugin {
 
   return {
     name: 'swc',
-    renderChunk(code: string, chunk, outputOptions: OutputOptions) {
+    renderChunk(code, chunk, outputOptions) {
       const chunkOptions = merge({}, options);
 
       if (outputOptions.sourcemap === 'inline') {
--- a/packages/swc/src/module.ts
+++ b/packages/swc/src/module.ts
@@ -1,7 +1,7 @@
 import { Plugin } from 'rollup';
 import { transform } from '@swc/core';
 import { merge } from 'smob';
-import { Options } from './type';
+import type { Options } from './type';
 
 export function swc(input: Options = {}): Plugin {
   const options: Options = merge(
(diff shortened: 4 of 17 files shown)


README.md
```diff
@@ -42,6 +42,7 @@ This repository houses plugins that Rollup considers critical to every day use o
 | [run](packages/run)                                 | Run your bundles in Node once they're built                                               |
 | [strip](packages/strip)                             | Remove debugger statements and functions like assert.equal and console.log from your code |
 | [sucrase](packages/sucrase)                         | Compile TypeScript, Flow, JSX, etc with Sucrase                                           |
+| [terser](packages/swc)                              | Transpile TypeScript/JavaScript with the speedy-web-compiler                              |
```
dasa, line 45:
```suggestion
| [swc](packages/swc)                              | Transpile TypeScript/JavaScript with the speedy-web-compiler                              |
```

tada5hi: good point :see_no_evil:

shellscape:
We'll need to disambiguate this plugin from https://github.com/rollup/plugins/tree/master/packages/sucrase within the README for the plugin.

tada5hi:
> We'll need to disambiguate this plugin from https://github.com/rollup/plugins/tree/master/packages/sucrase within the README for the plugin. Also lets get the rest of the PR template filled in.

what do you mean by disambiguate the plugin?
-> It uses `swc` instead of `sucrase`.

You mean the missing issue number?

tada5hi:
@shellscape can you have a look again? Tests are running through now.

packages/swc/README.md
```diff
@@ -0,0 +1,57 @@
...
+[![npm][npm]][npm-url]
+[![size][size]][size-url]
+[![libera manifesto](https://img.shields.io/badge/libera-manifesto-lightgrey.svg)](https://liberamanifesto.com)
+
+# @rollup/plugin-swc
+
+🍣 A Rollup plugin to transpile TypeScript/JavaScript with the speedy-web-compiler (swc).
+
+The plugin makes it possible to avoid the usage of `@rollup/plugin-babel` and `@rollup/plugin-typescript`.
+It is also blazingly fast 🔥 ( between 20 - 70 times faster than babel depending on how many cpu cores are used).
```
dasa, line 15:
```suggestion
It is also blazingly fast 🔥 (between 20 - 70 times faster than Babel depending on how many cpu cores are used).
```

shellscape:
> what do you mean by disambiguate the plugin?

That means using the readme to explain to people why they might use swc over sucrase, highlight differences, etc. This is a project where developer experience really matters, so when we're able to, it's good to give users info to make good choices.

tada5hi:
> > what do you mean by disambiguate the plugin?
> 
> 
> 
> That means using the readme to explain to people why they might use swc over sucrase, highlight differences, etc. This is a project where developer experience really matters, so when we're able to, it's good to give users info to make good choices.

You mean the root Readme.me of the whole rollup repository or? 
Otherwise I think we should mention alternatives in the Readme.md of both plugins `sucrase ` & `swc` and technically also technically in the `babel` plug-in readme, which is not the case in the moment, if I have seen this correctly. But I think we should address that in a followup pull request ( `docs(babel,sucrase,swc): xxx ) which I can also create with pleasure. For that we should release the swc plug-in to make the appropriate changes to the readme of all plug-ins

shellscape:
because this (swc) is the newest plugin, that info only needs to be in the readme for it. don't overthink 😺

tada5hi:
> because this (swc) is the newest plugin, that info only needs to be in the readme for it. don't overthink 😺 

"Overthinker/Overthinking"is a word I hear not too rarely 🙈, but I like the feature and don't see it as a disadvantage ☺️. 

My motive was to give the different transpilers the same treatment and that the respective audience or viewers get the same overview of possible alternatives 😇.

tada5hi:
@shellscape it would really be great to release this as soon as possible ✌️☺️
The desire for this also results from self-interest, since I just want to replace rollup based projects based on babel, due to the demonstrably much faster build time with swc. Currently I still redundantly copy the same logic, which I personally do not like and would like to avoid

shellscape:
🙏 Please enhance your patience. This PR was put up 20 hours ago. I and the other maintainers will review when we're able to. General rule of thumb is to wait a week before pinging people. If you need to use this immediately you can publish the package to your own scope on NPM or use a symlink and workspace reference locally.

packages/swc/src/type.ts
```diff
@@ -0,0 +1,3 @@
+import type { Options as SWCOptions } from '@swc/core';
+
+export type Options = SWCOptions & {};
```
tjenkinson, line 3: I’d suggest instead of making the top level the swc options object making the root object have a `swc` property, and have that be the thing that is passed to `swc`

This means if plug-in specific options get added in the future they are separate to the swc options and there can’t be any collisions

packages/swc/src/module.ts
```diff
@@ -0,0 +1,35 @@
...
+    },
+    input
+  );
+
+  return {
+    name: 'swc',
+    transform(code) {
+      return transform(code, {
+        ...options,
+        sourceMaps: true
```
tjenkinson, line 31: I’d have to check but doesn’t rollup provide a signal as to whether source maps are requested?

packages/swc/package.json
```diff
@@ -0,0 +1,73 @@
...
+  "peerDependenciesMeta": {
+    "rollup": {
+      "optional": true
+    }
+  },
+  "dependencies": {
+    "smob": "^0.0.6"
+  },
+  "devDependencies": {
+    "@swc/core": "^1.3.27",
```
tjenkinson, line 68: I wonder if this should be a dependency instead of bundling it in? 🤔

tjenkinson (commented):
Looks good. Few questions. Also please can you add a test that uses typescript?

packages/swc/src/type.ts
tada5hi: good point :ok_hand:

packages/swc/src/module.ts
tada5hi: rollup requires, as far as I could verify, that all transformers provide a sourcemap. Only plugins that use e.g. the renderChunk hook are supposed to provide a sourcemap depending on the corresponding value of the OutputOptions parameter.
[https://rollupjs.org/plugin-development/#renderchunk](https://rollupjs.org/plugin-development/#renderchunk)

packages/swc/package.json
tada5hi: I think we should avoid bundling external libraries, unless it is an end application. This prevents that a library has already been bundled several times via different third-party libraries and is therefore redundant in the final bundle.

tada5hi:
> Member
> ###

i'm trying, but i'm also well aware that this is not one of my strengths 🙈 .

packages/swc/README.md
```diff
@@ -0,0 +1,70 @@
...
+[![npm][npm]][npm-url]
+[![size][size]][size-url]
+[![libera manifesto](https://img.shields.io/badge/libera-manifesto-lightgrey.svg)](https://liberamanifesto.com)
+
+# @rollup/plugin-swc
+
+🍣 A Rollup plugin to transpile TypeScript/JavaScript with the speedy-web-compiler (swc).
+
+The plugin makes it possible to avoid the usage of `@rollup/plugin-babel` and `@rollup/plugin-typescript`.
+It is also blazingly fast 🔥 ( between 20 - 70 times faster than Babel depending on how many cpu cores are used).
```
dasa, line 15: Remove extra space:

```suggestion
It is also blazingly fast 🔥 (between 20 - 70 times faster than Babel depending on how many cpu cores are used).
```

packages/swc/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+declare type Options = {
+  swc?: SWCOptions;
+};
+```
+
+## Alternatives
+
+Alternative transpiler/transformer, which are also officially offered, are [Babel](https://www.npmjs.com/package/@rollup/plugin-babel) and [Sucrase](https://www.npmjs.com/package/@rollup/plugin-sucrase).
+Both Babel and Sucrase are written in JavaScript, whereas the core of Swc is written in Rust. In addition, Sucrase transforms code to ES6+, whereas babel and SWC also support ES3 upwards.
```
dasa, line 64:
```suggestion
Both Babel and Sucrase are written in JavaScript, whereas the core of SWC is written in Rust. In addition, Sucrase transforms code to ES6+, whereas babel and SWC also support ES3 upwards.
```

packages/swc/package.json
```diff
@@ -0,0 +1,73 @@
...
+  "bugs": "https://github.com/rollup/plugins/issues",
+  "main": "dist/cjs/index.js",
+  "module": "dist/es/index.js",
+  "exports": {
+    "types": "./types/index.d.ts",
+    "import": "./dist/es/index.js",
+    "default": "./dist/cjs/index.js"
+  },
+  "engines": {
+    "node": ">=10.0.0"
```
dasa, line 24: Should this be ">=14.0.0" to be the same as the other plugins and since that's what's tested?

tada5hi: swc and rollup v2 (https://www.npmjs.com/package/rollup/v/2.77.4-1?activeTab=explore) require a version of `>=10.0.0`. If someone installs rollup v3 instead the new engine requirement will be provided by the rollup dep ;)

dasa: Seems odd to be inconsistent with the other plugins though: ae4e9b3718c6bb9714443dac9cfa1dca6210480b

tada5hi: this min engine version corresponds to the min engine version of [slash](https://www.npmjs.com/package/slash?activeTab=explore) and slash is dependency, so its usage is mandatory. Andjust because other plugins have a different value, this one does not automatically have the same. I have already made considerations or research before I have adapted anything :v: :relaxed:

packages/swc/package.json
tjenkinson: Yep in which case this should be in `dependencies`?

packages/swc/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "@swc/core": "^1.x",
+    "rollup": "^2.x || ^3.x"
+  },
+  "peerDependenciesMeta": {
+    "rollup": {
+      "optional": true
+    }
+  },
+  "dependencies": {
+    "smob": "^0.0.6"
```
tjenkinson, line 65: I searched the repo and noticed `node-resolve` is using `deepmerge` for the same functionality. Is there a reason yo chose `smob` specifically? Otherwise maybe we should use `deepmerge` for consistency?

packages/swc/package.json
tada5hi: It is defined as `devDependency` and `peerDependency` so it is exluded for bundling. This is a trick to avoid linting and compiler issues, when importing different components and you are still able to let the end user have a arbitary `v1.3.x` version of the `@swc/core` package.

packages/swc/package.json
tada5hi: `deepmerge` does not provide type definition by default, which is a disadvantage, because you have to install `@types/deepmerge` additionaly. Besides, deepmerge only accepts **one** source, while smob accepts **multiple** sources to be merged to a target object. Furthermore, with smob a custom merger can be created, which allows to customize the merge behavior in certain situations.

tada5hi: Smob is also a zero dependency library and is used in the terser plugin as well ;)

packages/swc/package.json
tjenkinson: Ah I missed that it was a peer dependency

Personally I’m not a fan of peer dependencies as it puts the burden on the user to install it and make sure it stays within the required version range. When it isn’t a valid version it usually ends up as a warning amongst a load of other warnings that get missed

Also given the plugin requires the dependency in order to function is it really a peer dependency?

As a standard dependency there is still no duplication, unless the user also needs a different version for something else (which would work fine). Also on install it would normally resolve to the latest version and the user can update it when they update their lock file.

Maybe let’s see what others think.

If it does stay as a peer dependency I think we need to document it in the readme and it would be good to throw an error asking the user to add the peer dependency if it is missing

tada5hi: > Ah I missed that it was a peer dependency
> 
> Personally I’m not a fan of peer dependencies as it puts the burden on the user to install it (I think depending on the package manager?) and make sure it stays within the required version range. When it isn’t a valid version it usually ends up as a warning amongst a load of other warnings that get missed
> 
> Also given the plugin requires the dependency in order to function is it really a peer dependency?
> 
> As a standard dependency there is still no duplication, unless the user also needs a different version for something else (which would work fine). Also on install it would normally resolve to the latest version and the user can update it when they update their lock file.
> 
> Maybe let’s see what others think.
> 
> If it does stay as a peer dependency I think we need to document it in the readme and it would be good to throw an error asking the user to add the peer dependency if it is missing

Peer dependencies are automatically installed by the package manager and has many advantages. The problem with dependencies is that from v1+, only patches within the minor range are allowed. So we would always have to manually patch the version when security updates or new features are implemented. 
Only peer dependencies which are declared as **required: 'optional' ** in the `peerDependenciesMeta` section of the package.json file are not installed. 
I think it would be good if you would have another look at the npm and especially package.json features, because there are already cool gadges.

tjenkinson: Cool I will have a read about `peerDependenciesMeta` when I get chance :)

What about npm in node lts and other popular package managers? Do they follow the same spec and auto install?

Not following the version thing. If it’s a peer dependency and the user is using a lock file they have to choose when to update it, and if it’s a dependency, they also choose when to update it by running `npm update`. That would update the nested dependencies right (or maybe I remembered wrong)

tada5hi: I know that yarn supports it too. They also give a neat example https://classic.yarnpkg.com/lang/en/docs/dependency-types/#toc-peerdependencies
You got a good poin there with `npm update` :blush: , but this only applies for "end" libraries. Lock files are not passed when publishing, so everyone in the chain has to do the update process again, but otherwise you are of course right.

tjenkinson: Yeh so if iiuc dependency or peer dependency shouldn’t effect how frequently this plugin needs to be updated. With either option you can still configure a range with semver

Would be good to know what node lts does out of the box and if it has the same behaviour as latest

tada5hi: But I would like to focus more on the actual content, if that's okay with you, and not on whether a particular dependency should be managed as a dependency or peer-dependency, which is basically a general design question and has nothing to do with the plugin request itself.

Kocal (approved):

packages/swc/package.json
tada5hi: > Yeh so if iiuc dependency or peer dependency shouldn’t effect how frequently this plugin needs to be updated. With either option you can still configure a range with semver
> 
> Would be good to know what node lts does out of the box and if it has the same behaviour as latest

Node 16 (npm 7) definitly supports it, and regarding the official node repo, Node 16 is already not longer in LTS mode. https://github.com/nodejs/Release

tjenkinson: > But I would like to focus more on the actual content, if that's okay with you, and not on whether a particular dependency should be managed as a dependency or peer-dependency, which is basically a general design question and has nothing to do with the plugin request itself

I think it’s important to get this right as changing it later is possible but probably a breaking change

I’m happy to go with peer dependency if others also think this makes sense. I may also agree with you but want to do a bit of reading first on how peer dependencies work now first

Given they definitely didn’t use to install automatically and there may be people on older versions I think throwing a helpful error if the package is missing is probably worthwhile

tjenkinson: So essentially I don’t think this thread should be resolved unless another maintainer understands this and agrees it makes sense (which may be me once I’ve had chance to read up on it more) :)

tjenkinson: Ok done a bit more googling

>I know that yarn supports it too. They also give a neat example https://classic.yarnpkg.com/lang/en/docs/dependency-types/#toc-peerdependencies

>Having a peer dependency means that your package needs a dependency that is the same exact dependency as the person installing your package. This is useful for packages like react that need to have a single copy of react-dom that is also used by the person installing it.

I don't think that really applies in this use case. We don't _need_ the user to have the same version of `@swc/core` or even use it themselves. If they had to construct an instance of `@swc/core` and pass it in then I think that could apply.

The [docs for `peerDepdenciesMeta`](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#peerdependenciesmeta) say

>When a user installs your package, npm will emit warnings if packages specified in peerDependencies are not already installed. The peerDependenciesMeta field serves to provide npm more information on how your peer dependencies are to be used. Specifically, it allows peer dependencies to be marked as optional.

>Marking a peer dependency as optional ensures npm will not emit a warning if the soy-milk package is not installed on the host. This allows you to integrate and interact with a variety of host packages without requiring all of them to be installed.

Which reads to me like npm auto installing a peer dependency is not the ideal outcome. If it installs, but still emits a warning for the user, that's not ideal.

And docs for [`peerDependencies`](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#peerdependencies)

>In some cases, you want to express the compatibility of your package with a host tool or library, while not necessarily doing a require of this host. This is usually referred to as a plugin. Notably, your module may be exposing a specific interface, expected and specified by the host documentation.

We _are_ doing an import of `@swc/core` ourselves. I don't think `@swc/core` is a plugin for the rollup swc package.

>Node 16 (npm 7) definitly supports it, and regarding the official node repo, Node 16 is already not longer in LTS mode. https://github.com/nodejs/Release so we are fine ✌️

Thanks for checking. At least supported npm versions would be consistent if we did this.

I still don't completely understand what you meant by

>The problem with dependencies is that from v1+, only patches within the minor range are allowed.

? If we use `^1.3.27` that would mean users can get 1.3.27 and any v1 version later than that, without us having to publish an update to this package.

So my opinion is this should be a `dependency` not `peerDependency`

tada5hi: > Ok done a bit more googling
> 
> > I know that yarn supports it too. They also give a neat example https://classic.yarnpkg.com/lang/en/docs/dependency-types/#toc-peerdependencies
> 
> > Having a peer dependency means that your package needs a dependency that is the same exact dependency as the person installing your package. This is useful for packages like react that need to have a single copy of react-dom that is also used by the person installing it.

I think you have misunderstood. Maybe check out the following articles/sections.
https://nodejs.org/es/blog/npm/peer-dependencies/
https://docs.npmjs.com/cli/v9/configuring-npm/package-json#peerdependencies

> I don't think that really applies in this use case. We don't _need_ the user to have the same version of `@swc/core` or even use it themselves. If they had to construct an instance of `@swc/core` and pass it in then I think that could apply.
> 
> The [docs for `peerDepdenciesMeta`](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#peerdependenciesmeta) say
> 
> > When a user installs your package, npm will emit warnings if packages specified in peerDependencies are not already installed. The peerDependenciesMeta field serves to provide npm more information on how your peer dependencies are to be used. Specifically, it allows peer dependencies to be marked as optional.
> 
> > Marking a peer dependency as optional ensures npm will not emit a warning if the soy-milk package is not installed on the host. This allows you to integrate and interact with a variety of host packages without requiring all of them to be installed.
> 
> Which reads to me like npm auto installing a peer dependency is not the ideal outcome. If it installs, but still emits a warning for the user, that's not ideal.

For npm versions less than v7 a warning is issued and for versions greater than v7 the peer dependency is automatically installed if not already present.
https://github.blog/2021-02-02-npm-7-is-now-generally-available/#peer-dependencies

> And docs for [`peerDependencies`](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#peerdependencies)
> 
> > In some cases, you want to express the compatibility of your package with a host tool or library, while not necessarily doing a require of this host. This is usually referred to as a plugin. Notably, your module may be exposing a specific interface, expected and specified by the host documentation.
> 
> We _are_ doing an import of `@swc/core` ourselves. I don't think `@swc/core` is a plugin for the rollup swc package.

We are also doing an import on `rollup` in nearly every package of this monorepo and rollup is listed in every package.json as peer-dependency. Soa this is not an argument.

> > Node 16 (npm 7) definitly supports it, and regarding the official node repo, Node 16 is already not longer in LTS mode. https://github.com/nodejs/Release so we are fine ✌️
> 
> Thanks for checking. At least supported npm versions would be consistent if we did this.
> 
> I still don't completely understand what you meant by
> 
> > The problem with dependencies is that from v1+, only patches within the minor range are allowed.
> 
> ? If we use `^1.3.27` that would mean users can get 1.3.27 and any v1 version later than that, without us having to publish an update to this package.
> 
> So my opinion is this should be a `dependency` not `peerDependency`

tada5hi: By the way, the same is true for the babel plugin. So I really don't understand why you criticize the plugin but not there. This makes no sense to me. 
https://github.com/rollup/plugins/blob/master/packages/babel/package.json
Since this procedure is also the case with other transpiler plugins such as babel, I will not go into further detail here.

tjenkinson: >We are also doing an import on rollup in nearly every package of this monorepo and rollup is listed in every package.json as peer-dependency. Soa this is not an argument.

In this package the rollup import is just for types, which is just for dev, which is why it's a dev dependency. It also being a peer dependency does makes sense to me, because this plugin is adhering to the rollup plugin api, which may change in a future rollup major version

>By the way, the same is true for the babel plugin. So I really don't understand why you criticize the plugin but not there. This makes no sense to me.
https://github.com/rollup/plugins/blob/master/packages/babel/package.json

Good point. If that plugin was being added now I'd be making the same points, but that ship has sailed :)

My view after reading the docs is that this should be a `dependency`, for the reasons above.

As I've said before If other maintainers think it's fine as a peer dependency, sure lets go with that. @rollup/plugin-maintainers thoughts?

Please stop resolving this thread though. IMO it's not resolved yet.

tada5hi: Nevertheless, I would like to thank you for the time you have invested in the investigation and I am also well aware that you have the best intentions :relaxed:  . Thank you very much for that :v: . I only temporarily had the feeling that we were going around in circles when discussing peer dependency or rather dependency :see_no_evil: . But I also appreciate it very much to exchange ideas about such things. Maybe we get to work together on one or the other project / package independently of the review process :sunglasses: .

tada5hi:
@shellscape can you have a look ?

packages/swc/package.json
shellscape: Stepping in to be the tie-breaker, per request:

> By the way, the same is true for the babel plugin. So I really don't understand why you criticize the plugin but not there. This makes no sense to me.

It is never invalid to point out current issues even if those issues previously exist.

You're both making valid points, but there is needed nuance. Questions that lend to general guidance on this kind of thing:

- is the dependency _HUGE_?
- is the dependency pure JavaScript?
- is the dependency platform-depdendent?

### is the dependency _HUGE_?

This is a no-brainer. If the dependency is _huge_ like Babel or TypeScript, then we set it as a peer and let the user install the correct version within range. We don't want to be in the business of bundling typescript with the typescript plugin only to have the user also have typescript in their repo and bloat it further.

### is the dependency pure JavaScript?

If yes, then it's probably safe to bundle and it's one less thing the user has to worry about, but we also want to make sure we document what version of the target dependency that the plugin currently supports. `sucrase` and `buble` do this.

### is the dependency platform-depdendent?

If any of our dependencies need to be built for the platform of the machine the user is running on, we need to leave them as peers.

And this is where both of your arguments fall short. Within the `@swc/core` package there is a post-install script which runs and compiles bindings _which is platform dependent_. That means that there is no way we can reasonably bundle `swc` in this plugin so that it works on all platforms. We might get lucky and it just works, but with any dependency which has node bindings (or binaries) compiled as part of an install step, we need to leave that as something the user is responsible for.

WRT peers being auto-installed or not, documentation fixes this. If we list what the requirements are in documentation, then the user has recourse if they are not auto-installing peers. PNPM, NPM, and Yarn all do this now and all have settings to prevent this. So, should a user disable auto-installing peers (as I do) they need to look to the docs. We have a `Requirements` section in each README right at the top for exactly this purpose.

So, in summary, for the reasons listed above, `swc` dependencies should be `peerDependencies` for this plugin.

benmccann: Different package managers (npm, pnpm, yarn, etc.) and different versions of those package managers all handle `peerDependencies` differently. Very often the user must add the dependency to their `package.json` and install it though it sounds like the latest version of npm will do it for you.

I'm not quite following the argument that native libraries must be `peerDependencies`. I've not seen the installation of swc fail before and I'm not sure if installing swc fails how that would differ if it's in `dependencies` vs `peerDependencies` and whether they would need to follow the same steps or different steps in each case.

If it's helpful to see what others have done, Vite added it in `dependencies`: https://github.com/vitejs/vite-plugin-react-swc/blob/0d6588e76471bf5a801a0cab14808dafc78e7d09/package.json#L19

And it's done the same with `esbuild` as well: https://github.com/vitejs/vite/blob/356ddfe2e534800c1fd59fa502a2c4c8945f4f92/packages/vite/package.json#L69

I've not heard any complaints about this or instances of these libraries failing though there are probably others who would be more tuned in.

To me you use `peerDependencies` when you need the user to control the exact version being used or want them to provide the library. E.g. plugins declaring a dependency on the main package are a great use case for `peerDependencies`. I'd find making it a peer dependency in this case to be annoying because it's a package I wouldn't already have installed and would double the setup steps.

Personally, I would say to put it in `dependencies` though I don't really want to create a tie again and @shellscape's vote counts more than mine.

shellscape: The kicker is that we're attempting to bundle it. That's the differentiator. If we weren't bundling this plugin to ship it to NPM, `dependencies` would be correct.

benmccann: I agree I would definitely not bundle it. Can we exclude that dependency from bundling and list it in `dependencies`?

tjenkinson: Sorry I haven’t caught up, but to be clear I was proposing `dependencies` over `peerDependencies`.

I don’t think we should bundle it in and just put it in the dev list

shellscape: Excellent question. Probably? https://www.rollupjs.org/configuration-options/#external

tjenkinson: I think if we just put it in `dependencies` the `createConfig` helper would already configure `external` to include it

shellscape: yep. that's correct https://github.com/rollup/plugins/blob/69146cdf1f7353aa72364f69fa5e49c732a1abd8/shared/rollup.config.mjs#L15

That goes back to #363 which I didn't have a hand in. But there we go. It goes into `dependencies` and we're automatically not bundling it.

dasa:
> have your concerns beed addressed?

I think we still disagree on the supported Node version. Discussed at: https://github.com/rollup/plugins/pull/1412#pullrequestreview-1267973004

packages/swc/package.json
shellscape: We don't support Node versions out of LTS. Minimum rollup version should be v3 and Node version 14.

packages/swc/package.json
tada5hi: Since consistency is one of my basic principles, I would like us to take the same approach for `babel`. In general, the reasoning why it is dependency instead of peer-dependency has not convinced me much. But since @shellscape is code owner, I will follow his decision. 
To avoid arbitrariness or making someone additional work, I created a pull request for babel:
https://github.com/rollup/plugins/pull/1423

tada5hi:
Now that as far as I can survey, I think all concerns are resolved and we can complete the pull request @shellscape.

packages/swc/package.json
shellscape: @tada5hi this is still pending

tada5hi: sorry, I had misinterpreted that. :see_no_evil:

tada5hi: but is also done but is also solved by the latest commit

packages/swc/README.md
```diff
@@ -0,0 +1,70 @@
...
+# @rollup/plugin-swc
+
+🍣 A Rollup plugin to transpile TypeScript/JavaScript with the speedy-web-compiler (swc).
+
+The plugin makes it possible to avoid the usage of `@rollup/plugin-babel` and `@rollup/plugin-typescript`.
+It is also blazingly fast 🔥 (between 20 - 70 times faster than Babel depending on how many cpu cores are used).
+
+## Requirements
+
+This plugin requires an [LTS](https://github.com/nodejs/Release) Node version (v10.0.0+) and Rollup v2.0+.
```
dasa, line 19: I think this is meant to be Node 14+ and Rollup 3+

packages/swc/package.json
dasa: The `rollup` version under `peerDependencies` wasn't updated to remove v2.

packages/swc/README.md
tada5hi: Rollup v2 also works, which is also the case for most other plugins in this repository

shellscape: @tada5hi our baseline is rollup v3 for new plugins. I'll also add that to the contributing docs.

dasa: I based that on @shellscape comment at https://github.com/rollup/plugins/pull/1412#pullrequestreview-1267973004

tada5hi: Good Point 👍. I will change that asap.

tada5hi: done

tada5hi:
@shellscape min rollup version set to v3+ and swc is now peer-dep again. I think this should be it :v:

packages/swc/README.md
```diff
@@ -0,0 +1,70 @@
...
+# @rollup/plugin-swc
+
+🍣 A Rollup plugin to transpile TypeScript/JavaScript with the speedy-web-compiler (swc).
+
+The plugin makes it possible to avoid the usage of `@rollup/plugin-babel` and `@rollup/plugin-typescript`.
+It is also blazingly fast 🔥 (between 20 - 70 times faster than Babel depending on how many cpu cores are used).
+
+## Requirements
+
+This plugin requires an [LTS](https://github.com/nodejs/Release) Node version (v10.0.0+) and Rollup v3.0+.
```
dasa, line 19:
```suggestion
This plugin requires an [LTS](https://github.com/nodejs/Release) Node version (v14.0.0+) and Rollup v3.0+.
```

tada5hi: done

tada5hi:
@shellscape let me know if something is missing.

tada5hi:
@shellscape any news ?

alexgwolff:
any news ?

tada5hi:
I think the concerns of @dasa and @tjenkinson  have been resolved, maybe one of them will agree to give the last missing approval. Otherwise we have to wait for @shellscape ✌️

shellscape:
thanks!

Issue #1414, opened 2023-01-23
Plugin: CommonJS Shim


tada5hi:
Request Checklist:

- [x] I have searched https://npmjs.com/ before opening this issue
- [x] I have searched https://github.com/rollup/awesome before opening this issue

### New Plugin Use Case
This allows us to use __filename , __dirname, ... also for esm output bundles. Therefore, it is possible to use cjs syntax without to worry
about compatibility for esm bundles.

### New Plugin Proposal

This plugin allows to use specific cjs (__dirname,__filename, ...) syntax which will be replaced in esm output bundles.

tada5hi:
https://github.com/rollup/plugins/issues/1411

PR #1415, opened 2023-01-24
feat(esm-shim): initial release

tada5hi:
## Rollup Plugin Name: `esm-shim`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

https://github.com/rollup/plugins/issues/1414


tada5hi:
@shellscape here maybe to ✌️😅

packages/commonjs-shim/README.md
```diff
@@ -0,0 +1,49 @@
...
+[size]: https://packagephobia.now.sh/badge?p=@rollup/plugin-commonjs-shim
+[size-url]: https://packagephobia.now.sh/result?p=@rollup/plugin-commonjs-shim
+
+[![npm][npm]][npm-url]
+[![size][size]][size-url]
+[![libera manifesto](https://img.shields.io/badge/libera-manifesto-lightgrey.svg)](https://liberamanifesto.com)
+
+# @rollup/plugin-commonjs-shim
+
+🍣 A Rollup plugin to replace CJS syntax like `__filename`, `__dirname`, ... for esm output bundles.
```
shellscape, line 12:
```suggestion
🍣 A Rollup plugin to replace CJS global variables within esm output bundles.
```

Let's list all of them that are shimmed directly below this the above;

```
The list of global variables that are shimmed are:
- ...
```

packages/commonjs-shim/README.md
```diff
@@ -0,0 +1,49 @@
...
+```console
+npm install @rollup/plugin-commonjs-shim --save-dev
+```
+
+## Usage
+
+Create a `rollup.config.js` [configuration file](https://www.rollupjs.org/guide/en/#configuration-files) and import the plugin:
+
+```typescript
+import commonjsShim from '@rollup/plugin-commonjs-shim';
```
shellscape, line 31: We're technically not shimming CJS; we're shimming ESM so it can use CJS things. A more accurate name for this would `esm-shim`. I know that's kind of a PITA to change at this point but it's important to convey meaning

packages/commonjs-shim/src/constants.ts
```diff
@@ -0,0 +1,14 @@
+export const CJSyntaxRe = /__filename|__dirname|require\(|require\.resolve\(/;
+
+export const CJSShim = `
+// -- CommonJS Shims --
+import cjsUrl from 'url';
+import cjsPath from 'path';
+import cjsMod from 'module';
+const __filename = cjsUrl.fileURLToPath(import.meta.url);
```
shellscape, line 8: While it may seem obvious, we can't account for all the different ways this may be used and what whacky edge cases users may come up with. We should document that `__filename` is intended to contain the filename for the bundle in which the shim is inserted, and that there may be edge cases depending on its use.

packages/commonjs-shim/src/constants.ts
```diff
@@ -0,0 +1,14 @@
+export const CJSyntaxRe = /__filename|__dirname|require\(|require\.resolve\(/;
+
+export const CJSShim = `
+// -- CommonJS Shims --
+import cjsUrl from 'url';
+import cjsPath from 'path';
+import cjsMod from 'module';
```
shellscape, line 7:
```suggestion
import cjsModule from 'module';
```

let's be verbose. it doesn't cost us anything and will make debugging easier.

packages/commonjs-shim/src/index.ts
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2023.
+ * The core of this plugin was conceived by pi0 and is taken from the following repository:
+ * https://github.com/unjs/unbuild/blob/main/src/builder/plugins/cjs.ts
+ */
```
shellscape, line 5: good attempt, but we need to be a tad more proper about this. in the plugin's root directory we should have a copy of https://github.com/unjs/unbuild/blob/main/LICENSE named `unbuild.LICENSE` which contains the original text of that LICENSE. In the plugin's README in the `## Meta` section, we should include a line which states `Portions of this code were used with permission from [link to unbuild]. [unbuild LICENSE](./unbuild.LICENSE)`

If it was MPL or copyleft, we'd have to annotate our source.

```suggestion
```

packages/commonjs-shim/src/module.ts
```diff
@@ -0,0 +1,16 @@
+import type { Plugin } from 'rollup';
+
+import { transformCJSToESMSyntax } from './utils';
```
shellscape, line 3:
```suggestion
import { shimEsm } from './utils';
```

just a suggestion, bc I prefer succinct fn names.

packages/commonjs-shim/src/type.ts
```diff
@@ -0,0 +1,6 @@
+import type { SourceMapInput } from 'rollup';
```
shellscape, line 1: please rename to `types.ts` to match the pattern in the repo

packages/commonjs-shim/src/utils.ts
```diff
@@ -0,0 +1,35 @@
+import MagicString from 'magic-string';
+
+// Shim __dirname, __filename and require
+import { CJSShim, CJSyntaxRe, ESM_STATIC_IMPORT_RE } from './constants';
```
shellscape, line 4: let's pick one format for these const names please. you'll see the format `export const reThingBob` throughout the repo as well.

packages/commonjs-shim/test/test.js
```diff
@@ -0,0 +1,56 @@
...
+
+test.serial('inject cjs shim for esm output', async (t) => {
+  const bundle = await rollup({
+    input: 'test/fixtures/cjs.js',
+    plugins: [terser()]
+  });
+  const result = await bundle.generate({ format: 'es' });
+  t.is(result.output.length, 1);
+  const [output] = result.output;
+  t.is(output.code, `${CJSShim}const child = require('child');\n\nexport { child };\n`);
```
shellscape, line 24: let's use a snapshot here

packages/commonjs-shim/test/test.js
```diff
@@ -0,0 +1,56 @@
...
+
+test.serial('inject cjs shim for esm output with sourcemap', async (t) => {
+  const bundle = await rollup({
+    input: 'test/fixtures/cjs.js',
+    plugins: [terser()]
+  });
+  const result = await bundle.generate({ format: 'es', sourcemap: true });
+  t.is(result.output.length, 2);
+  const [output, map] = result.output;
+  t.is(
```
shellscape, line 36: snapshot please

packages/commonjs-shim/test/test.js
```diff
@@ -0,0 +1,56 @@
...
+
+test.serial('not inject cjs shim for cjs output', async (t) => {
+  const bundle = await rollup({
+    input: 'test/fixtures/cjs.js',
+    plugins: [terser()]
+  });
+  const result = await bundle.generate({ format: 'cjs' });
+  t.is(result.output.length, 1);
+  const [output] = result.output;
+  t.is(output.code, `'use strict';\n\nconst child = require('child');\n\nexports.child = child;\n`);
```
shellscape, line 54: snapshot please

shellscape (changes requested):

packages/commonjs-shim/README.md
tada5hi: done.

packages/commonjs-shim/src/constants.ts
tada5hi: good point. I added a short description for every variable in the README.md

packages/commonjs-shim/src/constants.ts
tada5hi: done.

packages/commonjs-shim/src/module.ts
tada5hi: done.

packages/commonjs-shim/src/type.ts
tada5hi: done.

packages/commonjs-shim/src/utils.ts
tada5hi: Since we are dealing with constants and not variables, the first letter should be capitalized in any case. done.

packages/commonjs-shim/test/test.js
tada5hi: done.

packages/commonjs-shim/test/test.js
tada5hi: done.

packages/commonjs-shim/test/test.js
tada5hi: done.

packages/commonjs-shim/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
```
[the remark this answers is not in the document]
tada5hi: I must honestly admit that I find the official recommendation of typescript, which is a shared configuration within a monorepos, which is then extended by the respective sub packages, the nicer and cleaner variant and would also like to use this preferably.

packages/commonjs-shim/README.md
tada5hi: No problem, since I have many principles by which I live and also program, it happens more often that I have to refactor code due to less than optimal decisions.

tada5hi: Moreover, you are absolutely right with your statement, since it is a function extension for esm and not cjs.

packages/commonjs-shim/src/index.ts
tada5hi: I have to be honest, I only followed the procedure that pi0 also follows for code that he integrates externally into his code, which he does not define as a depenency, but copies into his codebase.
We also have official permission from him to do whatever we want with the code, which is also covered by the MIT license. So I think our reference is more than sufficient and no explicit mention in the README.md or in the LICSENSE file is necessary.

Reference: 
- https://github.com/unjs/h3/blob/main/src/utils/multipart.ts
- https://github.com/unjs/h3/blob/main/LICENSE

tada5hi:
@shellscape let me know if something is missing.

tada5hi:
The requested changes are outdated, but fixed and github still complains about it :/

packages/commonjs-shim/src/index.ts
tada5hi: i like that argument :blush: it is always good to set a good example

packages/commonjs-shim/tsconfig.json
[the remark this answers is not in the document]
tada5hi: ok

shellscape:
thanks!

packages/esm-shim/README.md
```diff
@@ -0,0 +1,57 @@
...
+Create a `rollup.config.js` [configuration file](https://www.rollupjs.org/guide/en/#configuration-files) and import the plugin:
+
+```typescript
+import esmShim from '@rollup/plugin-esm-shim';
+
+export default {
+  input: 'src/index.js',
+  output: {
+    dir: 'output',
+    format: 'cjs'
```
dasa, line 43: Should this have been `format: 'es'` to match the check in packages/esm-shim/src/module.ts?

PR #1423, opened 2023-01-31
build(babel): define babel as dep instead peer-dep

tada5hi:

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### DIFF
--- a/packages/babel/package.json
+++ b/packages/babel/package.json
@@ -52,24 +52,20 @@
     "es6"
   ],
   "peerDependencies": {
-    "@babel/core": "^7.0.0",
-    "@types/babel__core": "^7.1.9",
     "rollup": "^1.20.0||^2.0.0||^3.0.0"
   },
   "peerDependenciesMeta": {
     "rollup": {
       "optional": true
-    },
-    "@types/babel__core": {
-      "optional": true
     }
   },
   "dependencies": {
+    "@babel/core": "^7.19.1",
     "@babel/helper-module-imports": "^7.18.6",
-    "@rollup/pluginutils": "^5.0.1"
+    "@rollup/pluginutils": "^5.0.1",
+    "@types/babel__core": "^7.1.9"
   },
   "devDependencies": {
-    "@babel/core": "^7.19.1",
     "@babel/plugin-external-helpers": "^7.18.6",
     "@babel/plugin-proposal-decorators": "^7.19.1",
     "@babel/plugin-syntax-dynamic-import": "^7.8.3",
@@ -77,7 +73,6 @@
     "@babel/preset-env": "^7.19.1",
     "@rollup/plugin-json": "^5.0.0",
     "@rollup/plugin-node-resolve": "^15.0.0",
-    "@types/babel__core": "^7.1.9",
     "rollup": "^3.2.3",
     "source-map": "^0.7.4"
   },
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -101,12 +101,6 @@ importers:
       rollup: 3.2.3
       typescript: 4.8.4
 
-  packages/auto-install/test/fixtures/pnpm:
-    specifiers:
-      node-noop: ^1.0.0
-    dependencies:
-      node-noop: 1.0.0
-
   packages/babel:
     specifiers:
       '@babel/core': ^7.19.1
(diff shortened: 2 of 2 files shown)


Andarist:
Could you mention **why** you want to make this change? I don't want to be dismissive so I would like to hear out the argument behind this. However, from my PoV, this is likely to be rejected.

tada5hi:
This is for consistency. This is mentioned in the swc pull request, which you are also involved in. There is no reason why Babel should be a peer dependency here and thus be an exception. Both Babel and swc are for transpilation.

shellscape:
@Andarist a few of the maintainers hashed out the same discussion on the new swc plugin PR here: https://github.com/rollup/plugins/pull/1412#discussion_r1085647334

Conclusion was that since `createConfig` is marking those dependencies as external, there's no default reason that a dependency like `@swc/core` (or babel) shouldn't be a direct dependency. Now, that doesn't mean that there isn't a non-obvious reason. This PR was opened to get all of the plugins on the same page, as we have other exceptions like the typescript plugin and the typescript peerDep, but if there's a reason for plugin-babel to keep the peerDep, I'm all ears.

Andarist:
This puts the "host project" at risk of installing different versions of Babel/SWC/etc and at risk of using those different versions by various tools. It's no fun to debug why a thing transpiles OK when running tests with Jest but it doesn't transpile OK when bundling (just an example).

This could also lead to weird issues when a plugin/preset/whatever created in the host project could not be recognized as such by another instance of Babel that got loaded by `@rollup/plugin-babel`. This depends on what kind of checks a tool like Babel do internally but this could easily introduce a breakage of this type. Consider this situation:
```js
// .babelrc.js
module.exports = { plugins: require('./local-plugin.js') }

// local-plugin.js
const babel = require('@babel/core')
module.exports = babel.createPlugin(/* ... */)
```

This doesn't reflect the actual Babel plugin API etc. I'm only using it to illustrate how the project could use a different version of Babel than we'd do internally and how that could lead to problems.

In the case of Babel (and other tools like this) I think that the peer dep is a better choice and that the user has to "provide" that is better.

shellscape:
@benmccann the reply above is a pretty well-reasoned rebuttal of the prior discussion for the swc plugin. thoughts?

benmccann:
babel and swc feel different to me because Babel is used by tons of dependencies like Jest, etc. That means you're more likely to run into conflicts, already have other dependencies specifying it as a peerDependency, etc. I'm not sure that users are likely to have anything else in their projects depending on swc?

Andarist:
A test runner (like Jest) often depends on some kind of a transformer - and ideally, one that uses the same tool under the hood that is used for bundling. For example, I have one project depending on `@swc/jest` - and that defines `@swc/core` as its peer dep.

IMHO the popularity of the project shouldn't impact the outcome of this decision. To me, it all boils down to the nature of a given project. I think that we can all say that Babel and SWC have the same nature here and it makes sense to me to treat both equally.

benmccann:
That's fair. I wasn't aware `@swc/jest` existed. I suppose in that case making it a `peerDependency` is pretty reasonable

tada5hi:
> babel and swc feel different to me because Babel is used by tons of dependencies like Jest, etc. That means you're more likely to run into conflicts, already have other dependencies specifying it as a peerDependency, etc. I'm not sure that users are likely to have anything else in their projects depending on swc?

This is also the case with swc. In addition, swc offers an alternative to ts-jest, with the name ts-jest.

Andarist:
I think the consensus has been reached here that those should stay as peer deps, so im going to close this for housekeeping purposes. However, if somebody is still in doubt - let’s continue discussing this.

tada5hi:
So @shellscape what is your decision ? I agree with @Andarist  that Babel as well as swc should be managed as peer dependency, for which I also argued from the beginning in a discussion with approx. 30 messages under different aspects.

shellscape:
There's no need to quantify your participation in a discussion. That comes across a little salty. We have these kinds of discussions because they can evolve and our maintainers each bring unique perspective. It's the utmost collaborative spirit. @Andarist brought new perspective and info that was compelling and added to the context of the conversation.

I agree we've reached a new consensus; peer dependencies is appropriate. 

I'll canonize this in the contributing docs in the next few days. Really appreciate everyone jumping into this one.

Issue #1211, opened 2022-06-17
[@rollup/plugin-node-resolve]: Should not consider node modules with trailing slash as built-in.


philfontaine:
- Rollup Plugin Name: @rollup/plugin-node-resolve
- Rollup Plugin Version: 13.3.0
- Rollup Version: 2.75.6
- Operating System (or Browser): Windows 11
- Node Version: 16.15.0
- Link to reproduction _(⚠️ read below)_: https://github.com/philfontaine/bug-rollup-string-encoder

### Expected Behavior

node modules with trailing slash should not be resolved as built-in, as explained in this comment https://github.com/nodejs/readable-stream/issues/391#issuecomment-446520961

I can't find the supporting docs to confirm that's how modules with forward slash should be resolved, but it seems to be this way since it breaks when it's required from the rollup bundle.

In my use case, I have a bunch of libs that depend on readable-stream@3 which contains `require('string_decoder/')`.

### Actual Behavior

The plugin thinks it's built-in because the [is-builtin-module](https://www.npmjs.com/package/is-builtin-module) module returns true, so it does not resolve it.

![image](https://user-images.githubusercontent.com/16233697/174311773-d70db45a-351e-4e73-9300-64375f9e95d1.png)

### Additional Information

To reproduce on the repo:

1. yarn build
2. node dist/index.js

To fix:

Seems like it just needs an endsWith('/') check on the module name to confirm that it's not built-in.

This maybe should be fixed upstream on the is-builtin-module package?

drewsynan:
I also ran into this issue trying to bundle `readable-stream` as a transitive dep. I wan't able to get @ericmorand's solution to work, but this worked for me:

```
import { nodeResolve } from '@rollup/plugin-node-resolve;
import isBuiltin from 'is-builtin-module';

...

plugins = [
   nodeResolve({
      resolveOnly: (module) => module === 'string_decoder' || !isBuiltin(module),
      preferBuiltins: false,
      exportConditions: ['node'],
   })
]
...
```

ymmv since this seems super hacky and like it would mess with libs that are expecting the actual builtin, but in my case it worked!

guybedford:
This is a standard CommonJS compat requirement, keeping it open for now.

rsweeney21:
I've submitted a pull request to is-buildin-module to fix this issue.

https://github.com/sindresorhus/is-builtin-module/pull/13

I tested the fix locally and it resolved this issue for me in rollup.

PR #1424, opened 2023-01-31
fix(node-resolve): bump is-builtin-module version, imports with a trailing slash

rsweeney21:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] yes 
- [ ] no

Breaking Changes?

- [x] no

#1211 

### Description

Adding a trailing slash to a module name in an import statement is the convention for telling node to use the local module with that name, rather than the builtin module with the same name. 

For an example, see the docs for [punycode](https://github.com/mathiasbynens/punycode.js) which is available as both a builtin and standalone module for node.

The is-builtin-module package did not respect this convention. A fix was made and released in is-builtin-modules v3.2.1. This change bumps the version of the dependency and adds a test case to verify that import statements with a trailing slash are not treated as builtin modules by the node-resolve package.


shellscape (approved):
Looks good, thanks for the additional test!

tada5hi (approved):

shellscape:
thanks!

PR #1434, opened 2023-02-17
fix(node-resolve): handle star in the middle in exports map

dummdidumm:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

`@rollup/plugin-node-resolve` can't deal with stars in the middle of a key of an exports map. It's allowed though (see [subpath patterns](https://nodejs.org/docs/latest-v18.x/api/packages.html#package-entry-points).

```json
{
	"exports": {
		"./foo/*.js": "./foo/*.js"
	}
}
```

To fix it I suggest to replace most of the current implementation around import and export maps with [resolve.exports](https://www.npmjs.com/package/resolve.exports), a widely-used npm package which aims to consolidate resolving export maps across tooling. It would potentially mean to get rid of some of the validation errors this plugin throws when encountering invalid export maps. I'm not sure if you want that, or to what extend, which is why I marked this as draft for now and only added a failing test.

### DIFF
--- a/packages/node-resolve/test/fixtures/exports-star-inbetween.js
+++ b/packages/node-resolve/test/fixtures/exports-star-inbetween.js
@@ -1,0 +1,5 @@
+import a from 'exports-star-inbetween/foo/a.js';
+import b from 'exports-star-inbetween/foo/b.js';
+import c from 'exports-star-inbetween/foo/bar/c.js';
+
+export default { a, b, c };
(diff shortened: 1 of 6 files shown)


shellscape:
@guybedford would like your input on this

shellscape:
Closing as abandoned.

dummdidumm:
Why was this closed as abandoned? You asked for input from another person, so I waited for a response from them, which never came. I'm happy to look into implementing the proposed PRs if you answer the questions in the PR description.

PR #1447, opened 2023-02-27
fix: commonjs: empty export when setter and defineProperty both used

coderaiser:
## Rollup Plugin Name: `{Commonjs}`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

Fixes:
- https://github.com/rollup/plugins/issues/1446
- https://github.com/rollup/rollup/issues/4882

When `setter` used with `defineProperty` would be better to use `getter` in `return` inside of variable received by `getter`:

before:

```js
var ansiStylesExports = {};
var ansiStyles = {
  get exports(){ return ansiStylesExports; },
  set exports(v){ ansiStylesExports = v; },
};

function requireAnsiStyles () {
	if (hasRequiredAnsiStyles) return ansiStylesExports;
	hasRequiredAnsiStyles = 1;
	(function (module) {
		function assembleStyles() {
			const styles = {
			};

			return styles;
		}

		// Make the export immutable
		Object.defineProperty(module, 'exports', {
			enumerable: true,
			get: assembleStyles
		});
} (ansiStyles));
        // empty object
	return ansiStylesExports;
}
```

after:

```js
var ansiStylesExports = {};
var ansiStyles = {
  get exports(){ return ansiStylesExports; },
  set exports(v){ ansiStylesExports = v; },
};

function requireAnsiStyles () {
	if (hasRequiredAnsiStyles) return ansiStylesExports;
	hasRequiredAnsiStyles = 1;
	(function (module) {
		function assembleStyles() {
			const styles = {
			};

			return styles;
		}

		// Make the export immutable
		Object.defineProperty(module, 'exports', {
			enumerable: true,
			get: assembleStyles
		});
} (ansiStyles));
        // correct result
	return ansiStyles.exports;
}
```


lukastaegert:
While this fixes the mentioned issues, it will also break #1306 again by partially reverting #1363. And in a bad: It keeps the inefficient weird getter/setter syntax while still relying on `syntheticNamedExports`.
I already started work on a proper fix for this some time ago that will go back to using the old object form while avoiding the reliance on `syntheticNamedExports`. If there is a lot of pressure to fix the issues, then please revert #1363 properly, knowing that we need to reopen #1306, otherwise wait until I completed my fix.

coderaiser:
When I write code this way:
- no `syntheticNamedExports`;
- no `weird getter/setter syntax`;
- go back to using the old object form

image

I have an runtime error:

image

shellscape:
Citing @lukastaegert we're going to close this one. #1455 should address this when merged.

Issue #1482, opened 2023-04-20
@rollup/plugin-terser hangs the process forever after completion when rollup is run in a worker thread


dumbmatter:
- Rollup Plugin Name: @rollup/plugin-terser
- Rollup Plugin Version: 0.4.0 and 0.4.1
- Rollup Version: 3.20.6 (or any other recent version)
- Operating System (or Browser): Ubuntu 20.04.06
- Node Version: v18.16.0
- Link to reproduction: https://github.com/dumbmatter/rollup-terser-worker-threads

### Expected Behavior

When the build completes, the process should terminate, even if rollup is being run inside a Node.js worker thread.

### Actual Behavior

When using @rollup/plugin-terser 0.4.x and running rollup inside a Node.js worker thread, the build does complete successfully, but the Node.js process never terminates.

Prior versions of @rollup/plugin-terser such as 0.3.0 did not have this problem, and running the same rollup build in the main process (rather than using a worker thread) also works fine even with @rollup/plugin-terser 0.4.x.

### Additional Information

This is very likely related to https://github.com/rollup/plugins/pull/1409 since that was the only PR merged between versions 0.3.0 and 0.4.0, and it's also related to worker threads.

dasa:
This change in the repro code allows the main thread to exit:

```diff
diff --git a/tools/build-worker.js b/tools/build-worker.js
index 106f78b..40e654d 100644
--- a/tools/build-worker.js
+++ b/tools/build-worker.js
@@ -6,6 +6,7 @@ const promise = new Promise(resolve => {
     );
 
     worker.on("message", () => {
+        worker.unref(); // or worker.terminate();
         resolve();
     });
 });
```

dumbmatter:
Interesting, thanks! Do you know why that's required with @rollup/plugin-terser v0.4, but not with v0.3? Like is it a bug in v0.4, or was I somehow only getting lucky that it worked in v0.3 in the first place? I've used worker threads a few times and hadn't previously come across this issue, so it feels a little strange if sometimes I need to unref to get the process to end, but other times I don't.

dasa:
> Do you know why that's required with @rollup/plugin-terser v0.4, but not with v0.3?

Yes, it's not returning early here:
https://github.com/rollup/plugins/blob/83197203e8fb506851abec0706666502b7ae91a8/packages/terser/src/worker.ts#L25-L27

I'll create a PR.

dasa:
PR created at #1483.

PR #1483, opened 2023-04-20
fix(terser): don't assume code is running in worker created by the worker pool

dasa:
Fixes issue shown at https://github.com/dumbmatter/rollup-terser-worker-threads by making sure the worker code is only executed when the worker was created by the worker pool.

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no (I don't see a good way to test for this change, but the current tests are testing for regressions.)

Breaking Changes?

- [x] no

List any relevant issue numbers: #1482

### Description

Currently the worker only checks if it's not running on the main thread and then assumes that if it it isn't, it was created by the worker pool. This isn't true when Rollup itself is being run in a worker.

### DIFF
--- a/packages/terser/src/constants.ts
+++ b/packages/terser/src/constants.ts
@@ -1,2 +1,3 @@
 export const taskInfo = Symbol('taskInfo');
 export const freeWorker = Symbol('freeWorker');
+export const workerPoolWorkerFlag = 'WorkerPoolWorker';
--- a/packages/terser/src/worker-pool.ts
+++ b/packages/terser/src/worker-pool.ts
@@ -5,7 +5,7 @@ import { EventEmitter } from 'events';
 
 import serializeJavascript from 'serialize-javascript';
 
-import { freeWorker, workerPoolWorkerFlag, taskInfo } from './constants';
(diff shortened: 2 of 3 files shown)


packages/terser/src/constants.ts
```diff
@@ -1,2 +1,3 @@
 export const taskInfo = Symbol('taskInfo');
 export const freeWorker = Symbol('freeWorker');
+export const workerPoolWorkerFlag = 'WorkerPoolWorker';
```
shellscape, line 3: is there a reason we cannot also use a symbol here?

dasa: This is sent as `workerData` in the `Worker` constructor, so it must be clonable.

References:
- https://nodejs.org/dist/latest-v18.x/docs/api/worker_threads.html#workerworkerdata
- https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm

tada5hi (approved):

shellscape:
thanks!

Issue #1492, opened 2023-05-12
Types for processRow option of RollupDSV


jamesscottbrown:
- Rollup Plugin Name: DSV
- Rollup Plugin Version: 3.0.2

### Expected Behavior / Situation

The [README for the DSV plugin](https://github.com/rollup/plugins/tree/master/packages/dsv) sats that the `processRow` function can "either manipulate the passed `row`, or return an entirely new row object."

I would expect that:

* manipulating the passed `row` and returning nothing should not be a type error
* returning a new row object that includes numbers should not be a type error (one of the main purposes of using the function is to convert `string` values to `Number` or `Date` values)

### Actual Behavior / Situation

The [type definition](https://github.com/rollup/plugins/blob/master/packages/dsv/types/index.d.ts) for `processRow` is:

```typescript
  processRow?: null | ((row: DSVRowString, id: string) => DSVRowString | undefined);
```

Problem 1: there is an example in the README, but it gives a type error because the function doesn't return anything, so has a return type of `void` rather than `undefined`:

```js
dsv({
  processRow: (row, id) => {
    Object.keys(row).forEach((key) => {
      var value = row[key];
      row[key] = isNaN(+value) ? value : +value;
    });
  }
});
```

Problem 2: constructing a new `row` object gives a type error, because `processRow` is expected to return a value of type `DSVRowString`, which has the type definition

```typescript
export interface DSVRowString {
    [key: string]: string;
}
```

### Modification Proposal

Change the `processRow` type definition to something like:

```typescript
  processRow?: null | ((row: DSVRowString, id: string) => DSVRowAny | undefined | void);
```

PR #1493, opened 2023-05-12
fix(dsv): update type of processRow option

jamesscottbrown:
## Rollup Plugin Name: `dsv`

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

List any relevant issue numbers: #1492 

### Description

This updates the type definition for the `processRow` option to the DVS plugin, so that:

* a function that doesn't return anything can be used
* a function that returns a row object that includes non-string values (such as number or dates) can be used

### DIFF
--- a/packages/dsv/types/index.d.ts
+++ b/packages/dsv/types/index.d.ts
@@ -1,4 +1,4 @@
-import type { DSVRowString } from 'd3-dsv';
+import type { DSVRowAny, DSVRowString } from 'd3-dsv';
 import type { FilterPattern } from '@rollup/pluginutils';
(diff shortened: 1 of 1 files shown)


shellscape:
Thanks for the PR. https://github.com/rollup/plugins/blob/master/packages/dsv/test/types.ts will need to be updated to consider the changes you've made here.

shellscape:
@jamesscottbrown ping. if this reaches 60 days without a response, it'll be closed as abandoned.

Issue #1503, opened 2023-05-20
[@rollup/plugin-dynamic-import-vars]: Import Assertions are stripped away


jorenbroekema:
- Rollup Plugin Name: dynamic-import-vars
- Rollup Plugin Version: 2.0.3
- Rollup Version: 3.22.0
- Operating System (or Browser): Linux Ubuntu 20.04 (WSL on Windows Home 11)
- Node Version: 18.15.0
- Link to reproduction _(⚠️ read below)_: https://replit.com/@JorenBroekema/rollup-plugin-dynamic-import-vars-repl#output/input.js

### Expected Behavior

Expected is that dynamic imports with variables in them, those that also contain a second parameter such as `{ assert: { type: 'css' } }`, keep this second parameter in the output:

```js
function __variableDynamicImportRuntime0__(path) {
  switch (path) {
    case './foo/foo.js': return import('./foo-1c6cfb55.js', { assert: { type: 'css' } });
    default: return new Promise(function(resolve, reject) {
      (typeof queueMicrotask === 'function' ? queueMicrotask : setTimeout)(
        reject.bind(null, new Error("Unknown variable dynamic import: " + path))
      );
    })
  }
}
```

### Actual Behavior

See line 3 in `output/input.js` in the REPL, the second argument is stripped away by the plugin when the dynamic import has a variable in it.

```js
function __variableDynamicImportRuntime0__(path) {
  switch (path) {
    case './foo/foo.js': return import('./foo-1c6cfb55.js');
    default: return new Promise(function(resolve, reject) {
      (typeof queueMicrotask === 'function' ? queueMicrotask : setTimeout)(
        reject.bind(null, new Error("Unknown variable dynamic import: " + path))
      );
    })
  }
}
```

### Additional Information

The REPL might be slightly confusing, because I'm importing a JS file with `{ assert: { type: "css" } }`, however, in my real world scenario, I'm actually importing a CSS file and using https://www.npmjs.com/package/rollup-plugin-import-css to turn it into a CSSStyleSheet inside a JS file. This plugin correctly create a CSSStyleSheet if CSS import assertion is used, but since this particular plugin strips the import assertion away, I get a CSS string instead, which means my code now fails, since it's different pre-bundling vs post-bundling (CSSStyleSheet vs CSS string)

PR #1504, opened 2023-05-20
fix(dynamic-import-vars): keep dynamic import assert arg intact

jorenbroekema:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: https://github.com/rollup/plugins/issues/1503

### Description

When transforming dynamic imports with variables in them, the second argument (import assertions) is stripped away and this information is lost. This Pull Request addresses that problem (see issue) by grabbing the argument Node and stringifying it and injecting it into the Magic String, so that this argument is kept as is.

This is needed for other plugins to correctly deal with imports that have import assertions, most notably https://github.com/jleeson/rollup-plugin-import-css which for CSS imports returns either a CSS string or CSSStyleSheet depending on whether or not an import assertion for CSS is present. If such an assertion is present, but dynamic-import-vars strips it away, the CSS plugin will bundle the CSS import as a CSS string rather than CSSStyleSheet, which would be different from local development, before bundling, since browsers natively do support import assertions now and return a CSSStyleSheet.

I might need some help with the testing, as it currently doesn't verify that the assertion is kept intact, I'm not sure how to test that, it seems in the final output[0].code the assertions are gone again, but when I test my changes in my local project, it does fix my issue.. and when I console.log the transform hook in this plugin with my changes, it does seem like it returns the correct code.


packages/dynamic-import-vars/package-lock.json
```diff
@@ -0,0 +1,354 @@
+{
```
shellscape, line 1: this file shouldn't exist

lukastaegert (commented):
Looks mostly fine to me, but the comment about adding the acorn plugin makes me curious as this plugin is already added by Rollup.

shellscape:
@jorenbroekema please give the replies here a look 🙏  if we reach 60 days without a reply, we'll consider the PR to be abandoned.

packages/dynamic-import-vars/README.md
```diff
@@ -145,6 +145,29 @@ function importLocale(locale) {
...
+
+Will ensure the following import assertions remain intact:
+
+```js
+// Refer to rollup-plugin-import-css https://github.com/jleeson/rollup-plugin-import-css
+function importLocale(sheet) {
+  return import(`./styles/${sheet}.css`, { assert: { type: 'css' } });
+}
+```
+
```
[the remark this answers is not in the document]
jorenbroekema: You're right, it works without needing to add it manually. I guess I based that off of some other plugin that was based on an older version of Rollup where this was still required.

packages/dynamic-import-vars/package-lock.json
jorenbroekema: removed, I probably accidentally installed with npm instead of pnpm or something

jorenbroekema:
Thanks for the reminder @shellscape , I addressed your comments, I might still need a hand on the test, the snapshot output seems to show the assert arg not appearing in the output, I'm not sure that's intended and I'm missing something, or my test is incorrect?

shellscape (approved):
I think that test is sufficient for now. Thanks for adding it. @lukastaegert please have another look.

lukastaegert (approved):
Looks good to me 👍

shellscape:
Sorry about the delay getting this merged. Thanks again!

PR #1445, opened 2023-02-26
feat(yaml): Support importing `.cff` files

janosh:
## Rollup Plugin Name: `yaml`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] yes (_bugfixes and features will not be merged without tests_)
- [x] no (not yet but happy to add)

Breaking Changes?

- [ ] yes (_breaking changes will not be merged unless absolutely necessary_)
- [x] no

If yes, then include "BREAKING CHANGES:" in the first commit message body, followed by a description of what is breaking. 

List any relevant issue numbers:

#868

### Description

[GH supports citable repos](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files) by adding a `citation.cff` file in YAML format at the root. In scientific contexts, this makes needing to load yaml files with `.cff` extension a common use case. Would be great if this was supported!

PR #1506, opened 2023-05-24
fix(yaml): Add extensions?: string[] to RollupYamlOptions

janosh:
Follow up to #1445.

## Rollup Plugin Name: `yaml`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

Follow up to #1445.

Fixes:

Screenshot 2023-05-24 at 11 45 25

### DIFF
--- a/packages/yaml/test/types.ts
+++ b/packages/yaml/test/types.ts
@@ -19,7 +19,8 @@ const config: RollupOptions = {
         }
 
         return data;
-      }
+      },
+      extensions: ['.yaml', '.yml', '.cff']
     })
   ]
 };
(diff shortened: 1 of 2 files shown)


shellscape:
Our PR and Issue templates are not optional. please replace abs update the PR template.

janosh:
abs?

shellscape (changes requested):
Thanks for the PR. We require new or updated tests for every PR opened. In this case, we need to assert that the type is correct by updating the `test/types.ts` file. Please add the `extensions` option to that code and we should be good.

janosh:
@shellscape Thanks for pointer. Added.

shellscape (approved):

PR #1512, opened 2023-06-05
feat: raw svg support in image plugin

fooku18:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

These changes are adding a flag for exporting raw content for svg extensions. This tries to replicate Webpacks `raw-loader`, but only for svg extensions. It comes handy, when a raw include strategy is used within an application.

Usage:
```js
// rollup.config.js
import image from '@rollup/plugin-image';

export default (async () => ({
	input: 'main.js',
	plugins: [
		image({ sourceSvg: true }),
	],
	output: {
		file: 'bundle.js',
		format: 'cjs'
	}
}))();
```

which will output svg imports as:
```js
var s='<svg role="img" class="svg-icon__icon" focusable="false" viewBox="0 0 22 22" xmlns="http://www.w3.org/2000/svg">    <path d="M12 0h-2v22h2V0Z"/>    <path d="M22 10H0v2h22v-2Z"/></svg>';export{s as default};
```


shellscape:
Thanks for the PR. typically we like folks to open a modification or feature issue to talk about new features and such. I don't see this being a good addition to the `image` plugin. That plugin is more generic and has a very narrow function. That's a good thing.

While I think your changes are cool, they're definitely geared specifically to SVG, as that functionality wouldn't be compatible with the other image formats the plugin supports. Because of that I'm not able to give my thumbs up to these changes and I think they'd be better off in a third party plugin.

My response isn't the final answer, as other maintainers here might have a different view.

shellscape:
As no one else has responded to the request to review in 60 days, we're going to close this one. Again, we appreciate the PR and your effort. I'm sorry we couldn't accept it at this time.

PR #1517, opened 2023-06-12
Update README.md adding an note on sourcemaps

dgrcode:

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

List any relevant issue numbers: https://github.com/rollup/plugins/issues/216 (this is the issue where I found the solution)

### Description
I added a note about sourcemaps in the plugin-typescript documentation. It might not be very detailed, but I felt lost when doing a quick search about sourcemaps didn't bring any result.

Hopefully this note can help people trying to produce sourcemaps.


tada5hi:
This information should be available in the documentation of rollup, as it is not plugin specific.

Issue #1524, opened 2023-06-29
[replace] saner default for delimiters


Hebilicious:
- Rollup Plugin Name: replace
- Rollup Plugin Version: 5.0.2

### Expected Behavior / Situation

Delimiters defaults to  `['(?<!\\.)\\b', '\\b(?!\\.)']`

### Actual Behavior / Situation

Delimiters currently defaults to `['\\b', '\\b(?!\\.)']`

### Modification Proposal

The delimiters defaults are a little too permissive, which means that it's easy to have an error where `globalThis.process.env.NODE_ENV` gets transformed into `globalThis."production"` which causes error in dependencies.

See conversation and and example on how it happened with `graphql` and `nitro` https://github.com/nuxt/nuxt/issues/21768#issuecomment-1613621429

I believe this fix should prevent other users from having the same issues in the future, and shouldn't break existing behaviours. However I am nowhere near a rollup expert and I could be wrong, so feel free to close this if this will break more things than it will fix.

If you think this is a good idea, I am happy to submit a PR with the implementation.

shellscape:
Thanks for the issue. The best route for you would be to make the change and see what tests fail. That would give you (and us) a decent idea of the impact. It'd definitely be considered a breaking change, but that wouldn't stop us from releasing it.

PR #1526, opened 2023-06-30
feat(replace): add safer default delimiters

Hebilicious:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] no

### Description

resolves https://github.com/rollup/plugins/issues/1524

This might be a BREAKING CHANGE for some existing users, but I'm assuming that in 99.9% of cases this is the desired behaviour, as the b delimiter is a little too greedy.

This doesn't break any existing tests, so I'm not sure if additional tests are needed.

### DIFF
--- a/packages/replace/src/index.js
+++ b/packages/replace/src/index.js
@@ -69,7 +69,7 @@ function expandTypeofReplacements(replacements) {
 
 export default function replace(options = {}) {
   const filter = createFilter(options.include, options.exclude);
-  const { delimiters = ['\\b', '\\b(?!\\.)'], preventAssignment, objectGuards } = options;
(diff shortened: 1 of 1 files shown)


shellscape (changes requested):
Generally good with this change. It will be a new major because it is technically breaking. We'll need a test that specifically tests the condition for the change (the overly permissive scenario you mentioned in the issue). That test _should not_ pass when this change is not present, and should pass when this change is present. Otherwise we won't be protected against regression there.

shellscape:
@Hebilicious please see the previous review. If this has no action after 60 days we'll consider it abandoned.

Hebilicious:
I'm sorry I'll close this as I don't have the capacity to work on this at the moment.

PR #1555, opened 2023-08-05
fix(typescript): emit declaration files for type-only source files that are not explicitly included

### LINKED ISSUE #1554: [@rollup/plugin-typescript] Declaration files for type-only source files that are not explicitly included are not emitted
- Rollup Plugin Name: `@rollup/plugin-typescript`
- Rollup Plugin Version: 11.1.2
- Rollup Version: 3.27.2
- Operating System (or Browser): Not relevant? I have reproduced the issue in both Ubuntu 22.04.2 & macOS 13.4.1 if it matters.
- Node Version: 18.17.0
- Link to reproduction _(⚠️ read below)_: https://github.com/ianyong/rollup-plugin-typescript-missing-declaration-files

### Expected Behavior

This issue does not occur with the TypeScript compiler because files that meet the two conditions below but are imported/exported from another source file that is part of the filenames or patterns in the `include` compiler option are still emitted.

### Actual Behavior

When the [`include` compiler option](https://www.typescriptlang.org/tsconfig#include) is specified in `tsconfig.json`, `@rollup/plugin-typescript` does not emit declaration files (`.d.ts`) for source files which:

1. Only contain types (i.e., does not contain any code which is transpiled to JavaScript); and
1. Are not part of the filenames or patterns in the `include` compiler option.

When the above conditions are met, the type definitions in the affected files are not part of the Rollup build output even if they are imported/exported from another source file that is part of the filenames or patterns in the `include` compiler option.

### Additional Information

Given the same `tsconfig.json`, I would expect the same set of type definitions to be outputted by `@rollup/plugin-typescript` & the TypeScript compiler.

ianyong:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

* Fixes #1554.

### Description

*Please read the linked issue first for context*.

The core issue is that TypeScript source files which contain only type definitions transpile into nothing. This means that Rollup has no equivalent JavaScript files to operate on (i.e., we cannot handle these files if they are not already included in `tsconfig.json` because within the Rollup build hooks and output generation hooks, these files do not exist). Specifically in the test case added, the `export` statement in `main.ts` is removed when transpiling into JavaScript because `MyNumber` is a type. This means that by the time Rollup operates on `main.js`, there is no longer any logical link between `main.ts` & `should-be-emitted-types.ts`.

However, the observation here is that we don't have to manually keep track of which files to emit because the TypeScript compiler already does it for us. Specifically, the TypeScript compiler only emits the files which are necessary (this is verified in the test case I added). This means that we can simply emit all declaration & source map files in `emittedFiles` to solve the linked issue.


packages/typescript/src/outputFile.ts
```diff
@@ -16,15 +16,22 @@ export interface TypescriptSourceDescription extends Partial<SourceDescription>
...
 }
 
 /**
  * Checks if the given OutputFile represents some source map
  */
-function isMapOutputFile(name: string): boolean {
-  return name.endsWith('.map');
+export function isMapOutputFile(name: string): boolean {
+  return name.endsWith('ts.map');
+}
```
ianyong, line 28: The reason for this change is that we only need to worry about emitting `.d.ts.map` (and equivalent `.d.cts.map` & `.d.mts.map`) files. The `.js.map` files are already being emitted by the TypeScript compiler (unless the `noEmit` option is set to `true`).

shellscape (approved):
Looks good. Will this alter any existing behavior, cause a difference in output (aside from additional files)

ianyong:
> Will this alter any existing behavior, cause a difference in output (aside from additional files)

No, the only change in behaviour is that in the `generateBundle` output generation hook, we output the set of files emitted by the TypeScript compiler instead of computing this set of files ourselves. Every other step in the build process is unchanged.

tada5hi (approved):

shellscape:
thanks!

GeekyEggo:
Understood, my apologies.

PR #1557, opened 2023-08-10
feat(node-resolve): pass original importee to secondary resolve

jchip:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

pass on original `importee` to other plugins.


shellscape:
Thanks for the PR. Could you please provide a much more detailed explanation on what this PR is adding, solving, or fixing for you, with examples?

jchip:
@shellscape We want to build plugins as part of our dev tools to offer module analytics and stats.  One use case is exploring a different approach to implementing module federation on rollup.  For that, at our plugin, we need two pieces of information:
1. The original import specifier from the code that does the `import`, like `import "react"`.
2. The actual full path to the module that `"react"` resolved to, which node-resolve provided.

Having the information in 1 and 2 above at the same time in a plugin is crucial to implement some of the tooling aspect we are exploring.  So the changes in this PR updates node-resolve to pass them both to further plugins as it does its second resolve.

lukastaegert (commented):
Looks good to me and might indeed be useful. The only thing I hope is that we can use this chance to document this feature (and the "resolved" feature) in the Readme so that people know about it, can you do that?

lukastaegert (approved):
Thanks a lot, looks good!

PR #1558, opened 2023-08-11
feat(wasm): include/exclude options

sky0014:
## Rollup Plugin Name: `{@rollup/plugin-wasm}`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

The mainly include/exclude handle code is from `@rollup/pluginutils` like the other plugin does, It's already been tested yet.

Breaking Changes?

- [x] no

### Description

`@rollup/plugin-wasm` missed the common include/exclude options, this PR added this part.

### DIFF
--- a/packages/wasm/README.md
+++ b/packages/wasm/README.md
@@ -46,6 +46,20 @@ Then call `rollup` either via the [CLI](https://www.rollupjs.org/guide/en/#comma
 
 ## Options
 
+### `exclude`
+
+Type: `String` | `Array[...String]`<br>
+Default: `null`
+
+A [picomatch pattern](https://github.com/micromatch/picomatch), or array of patterns, which specifies the files in the build the plugin should _ignore_. By default no files are ignored.
+
+### `include`
+
+Type: `String` | `Array[...String]`<br>
+Default: `null`
+
+A [picomatch pattern](https://github.com/micromatch/picomatch), or array of patterns, which specifies the files in the build the plugin should operate on. By default all wasm files are targeted.
+
 ### `sync`
 
 Type: `Array[...String]`<br>
--- a/packages/wasm/types/index.d.ts
+++ b/packages/wasm/types/index.d.ts
@@ -1,4 +1,5 @@
 import type { Plugin } from 'rollup';
+import type { FilterPattern } from '@rollup/pluginutils';
 
 /**
  * - `"auto"` will determine the environment at runtime and invoke the correct methods accordingly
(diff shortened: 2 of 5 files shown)


shellscape:
Thanks for the PR. CI is failing because while you added a new dependency, you hadn't run `pnpm install` and so the lockfile was never updated. Please do that and commit the lockfile.

A test for `include`/`exclude` would be welcome, but since most of the plugins don't test that, it's not required.

sky0014:
Hi @shellscape , just updated lock file and README, welcome to review~

sky0014:
Hi @shellscape , no message for several days, is there still some problem?

packages/wasm/README.md
```diff
@@ -92,6 +92,20 @@ Configures what code is emitted to instantiate the Wasm (both inline and separat
 - `"browser"` omits emitting code that requires node.js builtin modules that may play havoc on downstream bundlers
 - `"node"` omits emitting code that requires `fetch`
 
+### `exclude`
```
shellscape, line 95: please place these in the readme in alphabetical order 🙏

pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: 5.3
+lockfileVersion: 5.4
```
shellscape, line 1: we're not yet using pnpm v8 in this repo (see: https://github.com/rollup/plugins/blob/0f73e2ace5774b5b3e78584c6d1ed90d204a319e/package.json#L41) please downgrade temporarily, reset the pnpm-lock.yaml file and run install again for this PR.

(or we can wait on this PR until we update the pnpm usage in the repo)

packages/wasm/types/index.d.ts
```diff
@@ -31,6 +32,18 @@ export interface RollupWasmOptions {
    * Configures what code is emitted to instantiate the Wasm (both inline and separate)
    */
   targetEnv?: TargetEnv;
+  /**
+   * A picomatch pattern, or array of patterns, which specifies the files in the build the plugin
+   * should operate on.
+   * By default all wasm files are targeted.
+   */
+  include?: FilterPattern;
```
shellscape, line 40: please alphabetize here as well 🙏

shellscape (changes requested):

packages/wasm/README.md
sky0014: Of course no problem, but I noticed that the former options is not in alphabetical order either, did you mean that to re-order all the options?

pnpm-lock.yaml
sky0014: Hi, it's strange that I already used pnpm@7.12.2, but after I run `pnpm i`, it still upgrade to lock file 5.4.

sky0014: ok, I updated, it looks good now.

sky0014:
Just updated the PR, take a look when you have time @shellscape

shellscape:
@sky0014 looks good now, thanks. please update your branch from upstream so we can run the latest CI

sky0014:
> @sky0014 looks good now, thanks. please update your branch from upstream so we can run the latest CI

Done.

tada5hi (approved):

PR #1560, opened 2023-08-16
fix(esm-shim): pattern matching

klevente:
## Rollup Plugin Name: `esm-shim`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

When using `esm-shim` to insert the required variables inside ES Modules, the regex matching logic was not working correctly, the `matchAllPolyfill` function did not return the index from where in the file the matches was found. This led to an incorrect calculation on where to put the shim in the bundled file, which could cause invalid JS to be generated.

For example take a look at this Blitz: https://stackblitz.com/edit/node-hz948e?file=index.js.

Since the `String.matchAll` function is not available in this project, as `lib` is set to `es6` in `tsconfig.json`, I decided to modify the polyfill to correctly return the indices of where the matches start. I also refactored the code a bit so it does only what it needs to - find the last ESM import so it know to place the shim after it.

I added 2 test cases to test this feature: 1 with a single ESM import, and 1 with multiple imports.

I also tested a solution where I used `String.matchAll`, and that one worked out of the box - if whenever the `lib` option is updated to a newer standard, it'd be better to just use that instead.


tada5hi (approved):
looks good to me :thumbsup: 
Thank you!

klevente:
@shellscape could I ask for a review please? 🙏

shellscape:
thanks!

PR #1562, opened 2023-08-20
chore(repo): add missing `prepare` script for several packages

jonkoops:
## Rollup Plugin Name: `babel`, `data-uri`, `dynamic-import-vars`, `html`, `strip`

This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

List any relevant issue numbers: None

### Description

This PR adds the missing `prepare` script to several packages, so that they will be built upon running `pnpm install`. This aligns with the behavior of other packages.

Alternatively, this could be solved by having a single `prepare` script in the root package, which can then run this command for all projects in the workspace. This approach would prevent accidental inconsistencies from being introduced, such as those mitigated in this PR. It would also remove duplicated code between package scripts.

Please do let me know what the preferred solution for this would be, and I will adjust the PR accordingly if required.

### DIFF
--- a/packages/babel/package.json
+++ b/packages/babel/package.json
@@ -30,6 +30,7 @@
     "ci:lint:commits": "commitlint --from=${CIRCLE_BRANCH} --to=${CIRCLE_SHA1}",
     "ci:test": "pnpm test -- --verbose",
     "prebuild": "del-cli dist",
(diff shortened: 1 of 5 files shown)


shellscape:
thanks!

PR #1581, opened 2023-09-13
fix(dsv): fix type of processRow option

chika3742:
## Rollup Plugin Name: `dsv`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: #1492

### Description

(successor to #1493)

This fixes the type definition for the processRow option to the DSV plugin, so that:

- a function that doesn't return anything can be used
- a function that returns a row object that includes non-string values (such as `number` or `Date`) can be used

This PR also removes `undefined` from the return value type, since there is no case where `undefined` is explicitly returned.

### DIFF
--- a/packages/dsv/test/types.ts
+++ b/packages/dsv/test/types.ts
@@ -13,7 +13,17 @@ const config: RollupOptions = {
       include: 'node_modules/**',
       exclude: ['node_modules/foo/**', 'node_modules/bar/**'],
       processRow(row) {
-        return row;
+        return {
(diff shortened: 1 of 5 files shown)


shellscape:
@chika3742 please update your branch from upstream so we can run the latest CI

shellscape:
Looks like there are some CI failures that need a look

PR #1609, opened 2023-10-12
fix(replace): prevent variable before colons from being replaced when preventAssigment is true

hhgnsc:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

When `preventAssignment` is `true`, it will replace variable before colons.
For example, my options is:
```javascript
{
  preventAssignment: true,
  values: {
    NAME: JSON.stringify('replace')
  }
}
```
The input is:
```typescript
declare const NAME: string
console.log(NAME)
```
The output will be:
```typescript
declare const 'replaced': string
console.log('replaced')
```
The declare is broken.
I thiink this should be fixed.


shellscape (approved):

shellscape:
thanks!

EAGzzyCSL:
@hhgnsc  Hi, this may cause unexpected skip when ternary operator is used.

```js
console.log(Math.random() > 0.5? FOO: BAR)
```

repl: https://stackblitz.com/edit/rollup-case-unexpected-replace

shellscape:
@EAGzzyCSL could you please open that as a new issue so we can track it?

EAGzzyCSL:
@shellscape OK~ https://github.com/rollup/plugins/issues/1619

PR #1611, opened 2023-10-13
feat(dynamic-import-vars): Error when files found

koddsson:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

Hey hey! I'd like the `dynamic-import-vars` to error if there are no files found. I think this makes sense since the function would also error on runtime since it's just a function with a switch statement that defaults to throwing a runtime error.

Example dynamic import runtime function when there are no files in the directory.
```js
function __variableDynamicImportRuntime0__(path2) {
  switch (path2) {
    default:
      return new Promise(function(resolve, reject) {
        (typeof queueMicrotask === "function" ? queueMicrotask : setTimeout)(
          reject.bind(null, new Error("Unknown variable dynamic import: " + path2))
        );
      });
  }
}
```

I made this a breaking change but I can put this behind a option instead if we'd like. I think having it on by default makes a lot of sense again because this would already result in a runtime error anyways.


shellscape:
Hey thanks for the PR. This makes sense. This only caveat is that this hasn't been an error for _a very long time_. Given that the Vite ecosystem uses this plugin heavily (and will for some time to come yet), I'd like to take this in a slightly different route: hide this behind an option, which is defaulted to `false`. That will give us time to alert the Vite maintainers (and adjacent ecosystem projects) and allow them to turn on the option and prepare for the eventual breaking change. Our next subsequent release would be to deprecate the option, and then make it the default behavior with a new major version. Kind of a PITA but it's the responsible path when there are quite literally a million people relying on a package.

koddsson:
> Hey thanks for the PR. This makes sense. This only caveat is that this hasn't been an error for _a very long time_. Given that the Vite ecosystem uses this plugin heavily (and will for some time to come yet), I'd like to take this in a slightly different route: hide this behind an option, which is defaulted to `false`. That will give us time to alert the Vite maintainers (and adjacent ecosystem projects) and allow them to turn on the option and prepare for the eventual breaking change. Our next subsequent release would be to deprecate the option, and then make it the default behavior with a new major version. Kind of a PITA but it's the responsible path when there are quite literally a million people relying on a package.

Yup that's fair enough. I've moved this to be behind a option. Let me know what you think :)

koddsson:
Hey @shellscape; pinging you in case you don't get notifications for this. No rush though 😄

packages/dynamic-import-vars/src/index.js
```diff
@@ -55,6 +55,12 @@ function dynamicImportVariables({ include, exclude, warnOnError } = {}) {
               r.startsWith('./') || r.startsWith('../') ? r : `./${r}`
             );
 
+            if (errorWhenNoFilesFound && paths.length === 0) {
+              this.error(
+                new Error('No files found when trying to dynamically load concatted string')
```
koddsson, line 60: This error could do with a bit more context. What directory we are expecting there to be files for example.

shellscape:
Added the documentation we'd need for the new option to the README and updated the types. Something to keep in mind that we need for new features for the future. Thanks for the work on this one.

koddsson:
> Added the documentation we'd need for the new option to the README and updated the types. Something to keep in mind that we need for new features for the future. Thanks for the work on this one.

Cheers! I'll keep that in mind for next time :)

PR #1634, opened 2023-11-22
feat(alias): add warning to avoid unintended duplicate modules

sapphi-red:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

resolves https://github.com/rollup/rollup/issues/5250

### Description
When the alias plugin is configured like this
```js
// rollup.config.js
alias({ entries: { $: 'src' } })
```
and there a file like
```js
import { test as test1 } from './sub.js';
import { test as test2 } from '$/sub.js';

console.log(test1);
console.log(test2);
```
https://stackblitz.com/edit/rollup-repro-s5tmli?file=src%2Fmain.js

then, `import './sub.js'` and `import '$/sub.js'` resolves to different module. (The same file is included in the bundle twice.)
I believe this is not what people expect. Also IMO the fact that "`import 'src/sub.js'` doesn't work but `$/sub.js` works" is difficult to understand.

This PR adds a warning to catch these cases.

The warning will be shown for those that is using the alias plugin to duplicate the module, but I guess no one would do that. It'd be easier to create a different plugin in that case.

### DIFF
--- a/packages/alias/src/index.ts
+++ b/packages/alias/src/index.ts
@@ -1,3 +1,5 @@
+import path from 'path';
+
 import type { Plugin } from 'rollup';
 
 import type { ResolvedAlias, ResolverFunction, ResolverObject, RollupAliasOptions } from '../types';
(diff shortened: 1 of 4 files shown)


shellscape (approved):
this is great thank you

Issue #1437, opened 2023-02-19
Add an option to include non-legal identifier named exports to `dataToEsm` function of pluginutils


sapphi-red:
- Rollup Plugin Name: `pluginutils`
- Rollup Plugin Version: 5.0.2

### Feature Use Case
Currently `dataToEsm` function of pluginutils skips named exports for non-legal identifiers.
https://github.com/rollup/plugins/blob/bc5cf91a042b88e14a1da9ba25f501d92591ca8a/packages/pluginutils/src/dataToEsm.ts#L84-L96
For example, `dataToEsm({ foo: 0, 'foo.bar': 0 }, { namedExports: true })` generates the following code.
```js
export var foo = 0;
export default {
        foo: foo,
        "foo.bar": 0
};
```
ES2022 introduced Arbitrary module namespace identifier names(https://github.com/tc39/ecma262/pull/2154) feature that allows exporting non-legal identifiers. By taking advantage of this, we are now able to export variables with any names that is a valid "Well-Formed Code Unit Sequence".
With the previous example, we can generate the following code.
```js
export var foo = 0;
var __arbitary0 = 0;
export { __arbitary0 as "foo.bar" };
export default {
        foo: foo,
        "foo.bar": 0
};
```

Related: https://github.com/vitejs/vite/issues/11359

### Feature Proposal
Add `allowStringAliasedNamedExports: true` option to `dataToEsm` that generates the code above.
[...]

PR #1635, opened 2023-11-22
feat(pluginutils): add `includeArbitraryNames: true` to `dataToEsm`

sapphi-red:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

resolves https://github.com/rollup/plugins/issues/1437

### Description

This PR adds `namedExports: 'include-arbitrary-name'` option to `dataToEsm` function in `pluginutils`.
When this option is set, `dataToEsm` will generate named export for keys that was not a valid identifier by leveraging Arbitrary module namespace identifier names feature (https://github.com/tc39/ecma262/pull/2154).

### DIFF
--- a/packages/pluginutils/types/index.d.ts
+++ b/packages/pluginutils/types/index.d.ts
@@ -10,9 +10,14 @@ export interface AttachedScope {
 
 export interface DataToEsmOptions {
   compact?: boolean;
+  /**
+   * @desc When this option is set, dataToEsm will generate a named export for keys that
+   * are not a valid identifier, by leveraging the "Arbitrary Module Namespace Identifier
+   * Names" feature. See: https://github.com/tc39/ecma262/pull/2154
+   */
+  includeArbitraryNames?: boolean;
   indent?: string;
   namedExports?: boolean;
-  includeArbitraryNames?: boolean;
   objectShorthand?: boolean;
   preferConst?: boolean;
 }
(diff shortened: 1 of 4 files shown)


packages/pluginutils/types/index.d.ts
```diff
@@ -11,7 +11,7 @@ export interface AttachedScope {
 export interface DataToEsmOptions {
   compact?: boolean;
   indent?: string;
-  namedExports?: boolean;
+  namedExports?: boolean | 'include-arbitrary-name';
```
shellscape, line 14:
```suggestion
  namedExports?: boolean | 'include-arbitrary-name';
```

personally not a fan of mixing types on options. I'd rather see a `includeAtrbitraryNames: boolean` that worked in conjunction with `namedExports`. This increases the footprint of `DataToEsmOtions` but clearly defines behavior.

shellscape (changes requested):

packages/pluginutils/types/index.d.ts
sapphi-red: Sure. I changed to that.

shellscape:
We've got a regex issue with our release step. I'll fix that this week and we'll get this published.

PR #1648, opened 2023-12-14
fix(node-resolve): prevent infinite findPackageJson loop

### LINKED ISSUE #1647: Infinite loop in combination with virtual module and missing package.json
- Rollup Plugin Name:  plugin-node-resolve
- Rollup Plugin Version: 15.2.3
- Rollup Version: 4.9.0
- Operating System (or Browser): macOS Ventura 13.1
- Node Version: v20.10.0
- Link to reproduction _(⚠️ read below)_:  https://github.com/dasdeo/rollup-resolve-node-bug

### Expected Behavior

Bundling should work the same regardless from where the bundler was invoked.

### Actual Behavior

If the plugin fails to find a corresponding package.json it will hang forever.

Note that this is not remedied by setting `rootDir`.

### Additional Information

I already tracked down the offending function, it is:

```js
async function findPackageJson(base, moduleDirs) {
    const { root } = path.parse(base);
    let current = base;
    while (current !== root && !isModuleDir(current, moduleDirs)) {
        const pkgJsonPath = path.join(current, 'package.json');
        if (await fileExists(pkgJsonPath)) {
            const pkgJsonString = fs.readFileSync(pkgJsonPath, 'utf-8');
            return { pkgJson: JSON.parse(pkgJsonString), pkgPath: current, pkgJsonPath };
        }
        current = path.resolve(current, '..');
    }
    return null;
}
```

Variable `current` over time:

```
findPackageJson my-virtual-module
findPackageJson /Users/dasdeo/Desktop/rollup-bug-repo
findPackageJson /Users/dasdeo/Desktop
findPackageJson /Users/dasdeo
findPackageJson /Users
findPackageJson /
findPackageJson /
findPackageJson /
...
```

dasdeo:
This is fixed by checking if the root directory (/) was already looped over.

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

resolves #1647

### DIFF
--- a/packages/node-resolve/src/package/utils.ts
+++ b/packages/node-resolve/src/package/utils.ts
@@ -11,7 +11,7 @@ function isModuleDir(current: string, moduleDirs: readonly string[]) {
 export async function findPackageJson(base: string, moduleDirs: readonly string[]) {
   const { root } = path.parse(base);
   let current = base;
-  let prev_current = null;
+  let prevCurrent = null;
 
   while (current !== root && !isModuleDir(current, moduleDirs)) {
     const pkgJsonPath = path.join(current, 'package.json');
@@ -19,11 +19,11 @@ export async function findPackageJson(base: string, moduleDirs: readonly string[
       const pkgJsonString = fs.readFileSync(pkgJsonPath, 'utf-8');
       return { pkgJson: JSON.parse(pkgJsonString), pkgPath: current, pkgJsonPath };
     }
-    prev_current = current;
+    prevCurrent = current;
     current = path.resolve(current, '..');
 
     // issue #1647 prevent infinite loop for virtual modules
-    if (current === "/" && prev_current === "/") break;
+    if (current === '/' && prevCurrent === '/') break;
   }
   return null;
 }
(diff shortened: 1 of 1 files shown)


shellscape:
Thanks for the PR! We'll need a test added for this before we can merge though. The test should probably recreate the situation as noted in the linked issue, and it should fail if the fix isn't present.

dasdeo:
I would write a unit test if I knew how since this is using the file system.

Any resources on how you deal with that in this project?

shellscape:
yeah most of our tests use file based fixtures for executing a build. I'd use those as a reference.

dasdeo:
I just had this thought, I could be using my repository that I used to show the bug as a template for the fixture.

I _might_ be adding the test case for this bug in the next couple of days.

Just to be sure, tests are located at https://github.com/rollup/plugins/tree/master/packages/node-resolve/test/fixtures right?

shellscape:
> Just to be sure, tests are located at https://github.com/rollup/plugins/tree/master/packages/node-resolve/test/fixtures right?

The fixtures are there. The tests are a directory level up

shellscape:
@dasdeo will you have the chance to write a test for this change? I'd like to merge it, but we need to protect against future regressions.

shellscape:
Closing as abandoned.

PR #1659, opened 2023-12-30
Modernize @rollup/plugin-terser to module

vdegenne:
## Rollup Plugin Name: `@rollup/plugin-terser`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [ ] no

### DIFF
--- a/packages/terser/package.json
+++ b/packages/terser/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@rollup/plugin-terser",
   "version": "0.4.4",
+  "type": "module",
   "publishConfig": {
     "access": "public"
   },


tada5hi:
I would like to postpone this until we migrate the whole monorepo to esm.

shellscape:
Let's not close PRs until we've had a chance to review.

shellscape:
@vdegenne could you please explain your reasons for this change? We export both ESM and CJS for each package, which should make usage seamless. Is there a problem that you're trying to solve with this change?

vdegenne:
> @vdegenne could you please explain your reasons for this change? We export both ESM and CJS for each package, which should make usage seamless. Is there a problem that you're trying to solve with this change?

Sorry I think it's a misunderstanding, I was trying to use `@rollup/plugin-terser` in `rollup.config.TS` and I had an error, adding `"type": "module"` manually into the `package.json` solved the issue...  ...until I realized I was using `"module": "NodeNext"` in my ts config. I changed the value to `Node` and the error was gone.
I see this package is still providing CommonJS so I guess you are not going to accept this change any time soon.

shellscape:
OK thanks for explaining. Yeah unfortunately we won't be moving to only ES exports soon.

nikelborm:
Sorry, but why is this comment marked as abuse? Is it by mistake, or is there something to be learned in communication?
image

shellscape:
probably a fat finger

Issue #1665, opened 2024-01-19
export { terser as default } in dist


vdegenne:
https://github.com/rollup/plugins/blob/2a19079892f0bef53b557da965339cdef0a13a93/packages/terser/src/index.ts#L8

I don't know why but this seems to get rewritten as `export { terser as default }`
![image](https://github.com/rollup/plugins/assets/2827383/be96cb00-62ac-45d8-9d86-e8fc032182c9)

I believe this should not be altered for esm. I'm trying to use `@rollup/plugin-terser` in a TypeScript and I'm not even sure how to import it.

PR #1694, opened 2024-03-17
feat(node-resolve): allow preferBuiltins to be a function

### LINKED ISSUE #1655: node-resolve: allow more granular config of preferBuiltins
- Rollup Plugin Name: @rollup/node-resolve
- Rollup Plugin Version: 15.2.3

### Feature Use Case
Some NodeJS builtins (e.g. `punycode`) are deprecated, and [replacement packages with the same name exist in NPM](https://www.npmjs.com/package/punycode). It is desirable to have `preferBuiltins` set to false for these specific cases, while keeping it `true` for all other builtins (e.g. `string_decoder` is also available as both a [builtin](https://nodejs.org/api/string_decoder.html) and a [package](https://www.npmjs.com/package/string_decoder), but in this case the builtin is stable and should be preferred).

### Feature Proposal
Allow `preferBuiltins` to accept a regex, array, or function, matching the behaviour of e.g. [`external`](https://rollupjs.org/configuration-options/#external). The [relevant code](https://github.com/rollup/plugins/blob/7f59e56620d47e6d2179541e6fc9391a92cc451e/packages/node-resolve/src/index.js#L193) is already considering these on an import-by-import basis, so it should be possible without any restructuring needed.

For example, the config for the `punycode` example above could be:

```js
nodeResolve({
  preferBuiltins: (module) => module !== 'punycode',
})
```

(perhaps an even better API would be to reverse the config, preferring builtins by default and offering a `preferModules` config, since that would allow `preferModules: ['punycode']`, but that is not backwards-compatible, or at least introduces ambiguity if both options are used)

younggglcy:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: resolves #1655

### Description

I think #1665 proposes a nice feature that gives `preferBuiltins` a more fine-grained control, so I'm trying to resolve this a little inactive issue.

This PR only makes `node-resolve` accept a function parameter to `preferBuiltins` to introduce as little as possible without changing the existing API.

#1665 also proposes that it should match the behavior of the `external` rollup configuration. But I think in most cases, we only need to mark some modules as non-builtins, e.g., deprecated modules like `punycode`. And this is somehow contrary to the behavior of the `external`.

Is accepting a function enough or should we add some extra configs? I'm willing to contribute if we need to make more changes.

### DIFF
--- a/packages/node-resolve/test/prefer-builtins.js
+++ b/packages/node-resolve/test/prefer-builtins.js
@@ -30,13 +30,12 @@ test('handles importing builtins', async (t) => {
 });
 
 test('warning when preferring a builtin module, no explicit configuration', async (t) => {
-  let warning = null;
+  let warning = '';
   await rollup({
     input: 'prefer-builtin.js',
-    onwarn({ message }) {
-      // eslint-disable-next-line no-bitwise
-      if (~message.indexOf('preferring')) {
-        warning = message;
+    onwarn({ message, pluginCode }) {
+      if (pluginCode === 'PREFER_BUILTINS') {
+        warning += message;
       }
     },
     plugins: [nodeResolve()]
@@ -47,7 +46,8 @@ test('warning when preferring a builtin module, no explicit configuration', asyn
     warning,
     `preferring built-in module 'events' over local alternative ` +
       `at '${localPath}', pass 'preferBuiltins: false' to disable this behavior ` +
-      `or 'preferBuiltins: true' to disable this warning`
+      `or 'preferBuiltins: true' to disable this warning.` +
+      `or passing a function to 'preferBuiltins' to provide more fine-grained control over which built-in modules to prefer.`
   );
 });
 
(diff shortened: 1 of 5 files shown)


packages/node-resolve/test/prefer-builtins.js
```diff
@@ -30,13 +30,12 @@ test('handles importing builtins', async (t) => {
 });
 
 test('warning when preferring a builtin module, no explicit configuration', async (t) => {
-  let warning = null;
```
shellscape, line 33: we always prefer adding another test instead of modifying an existing test. What was the reason that this test was modified?

younggglcy: Because I added new warning message after the original [here](https://github.com/rollup/plugins/pull/1694/files#diff-dbc8c6c4e7383a37fbae0c12d8e6d8901c93631bb7774f4bd1a2a88bbe1980a0R234-R239).
So this test is modified to let the warnings match exactly.

tada5hi (approved):

shellscape:
@younggglcy please take a look at the conflict that popped up. we're good to merge once that's resolved.

younggglcy:
> @younggglcy please take a look at the conflict that popped up. we're good to merge once that's resolved.

Sure, I'll handle it off work

shellscape:
thanks!

PR #1699, opened 2024-03-29
feat(run): add `input` option

mkantor:
## Rollup Plugin Name: `run`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

This changeset adds an `input` option to the `run` plugin. This options allows specifying which entry point to run if you have multiple in your bundle.


packages/run/README.md
```diff
@@ -93,6 +93,13 @@ When this option is enabled, `stdin` will listen for the following input and per
 - `clear` → Clears the screen of all text
   Also allowed: `cls`, `CTRL+L`
 
+### `input`
```
shellscape, line 96: please move this up before `stdin` so it's in alphabetic order

shellscape (changes requested):
thanks for the PR. small README change needed

PR #1702, opened 2024-04-12
chore(repo): fix incorrect import path in virtual plugin

tk-1io:
## Rollup Plugin Name: `virtual`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] yes (covered by existing tests)
- [ ] no

Breaking Changes?

- [x] no

Might fix builds for PR #1578, so it probably could be accepted too then

### Description

The "virtual" plugin package currently [imports its own types from `../`](https://github.com/rollup/plugins/blob/9e7f576f33e26d65e9f2221d248a2e000923e03f/packages/virtual/src/index.ts#L5), even though its types are actually placed inside `../types`. Every other plugin appears to correctly import from `../types` too (see e.g. [alias](https://github.com/rollup/plugins/blob/9e7f576f33e26d65e9f2221d248a2e000923e03f/packages/alias/src/index.ts#L5), [auto-install](https://github.com/rollup/plugins/blob/9e7f576f33e26d65e9f2221d248a2e000923e03f/packages/auto-install/src/index.ts#L9), ...).
I stumbled across this while attempting to build the code of PR #1578 (which i need a fix for too) on my local machine, which errored due to the incorrect import. I noticed the current state is building solely because the virtual plugin lists its type definitions file in its [`package.json` via the `types` property](https://github.com/rollup/plugins/blob/9e7f576f33e26d65e9f2221d248a2e000923e03f/packages/virtual/package.json#L69), which miraculously lets `tsc` accept the faulty import (even though it shouldn't, in my opinion). My guess is it just works somehow because `../` points to the folder where the `package.json` lives in, which then references the types file 🤷‍♂️. Now, because in the PR #1578 the `types` property [is gone](https://github.com/rollup/plugins/commit/1b5ab9316c77c16c02b1b558ef0d7e990a71b272#diff-80feda48e3887b5e7f39974479e9c8220152072cd3ebb2898e6667b24fe09885L71), we get an error there.
I previously [opened a PR](https://github.com/Geylnu/rollup-plugins/pull/1) to fix the branch of the original PR, but @Geylnu did not respond yet. So why did i open this PR then? Because i think this import path is incorrect and should be fixed anyway, no matter if the other PR will at some point be accepted (or maybe never might be).

### DIFF
--- a/packages/virtual/src/index.ts
+++ b/packages/virtual/src/index.ts
@@ -2,7 +2,7 @@ import * as path from 'path';
 
 import type { Plugin } from 'rollup';
 
-import type { RollupVirtualOptions } from '../';
+import type { RollupVirtualOptions } from '../types';
 
 const PREFIX = `\0virtual:`;
 


tk-1io:
@shellscape Is there anything I can do to get this pull request merged?

shellscape:
thanks!

Going to change this to a repo chore, since it doesn't affect the package output itself and is more about local dev.

PR #1722, opened 2024-05-23
feat(virtual): accept a function resolver for the `id` option

### LINKED ISSUE #1721: Allow @rollup/plugin-virtual to be a function that takes a module ID
- Rollup Plugin Name: @rollup/plugin-virtual
- Rollup Plugin Version: v3.0.2

### Feature Use Case
Just like https://github.com/rollup/plugins/blob/master/packages/replace/CHANGELOG.md#120
```js
// src/views/users/index.vue
import { routeName } from 'virtual:routeName';

// the routeName is base on current file, like src/views/users/index.vue ==> users/index
```

### Feature Proposal
```js
import virtual from '@rollup/plugin-virtual';

export default {
  input: 'src/entry.js',
  // ...
  plugins: [
    virtual({
      'virtual:routeName': (importer) => {
        const routeRoot = process.cwd() + '/src/views/';
        const routeName = importer.replace(routeRoot, '').replace(/\.\w+$/, '');

        return `export const routeName = '${routeName}'`;
      },
    }),
  ],
};
```

qzsiniong:
## Rollup Plugin Name: `@rollup/plugin-virtual`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

resolves #1721


packages/virtual/types/index.d.ts
```diff
@@ -1,7 +1,8 @@
 import type { Plugin } from 'rollup';
 
+export type RollupVirtualOption = string | ((importer: string | undefined) => string);
```
shellscape, line 3:
```suggestion
export type VirtualIdResolver = string | ((importer: string | undefined) => string);
```

shellscape (changes requested):
Thanks for the PR. Please change up the name of the new type you've introduced, so it avoids any potential confusion.

qzsiniong:
@shellscape any problem?Please review again

PR #1728, opened 2024-06-03
fix(typescript)!: correctly resolve filenames of declaration files for `output.file`

lameuler:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

Currently, when `output.file` is specified instead of `output.dir`, the filename for emitting declaration files with `this.emitFile` is resolved incorrectly. This means that given a `declarationDir` that is outside of the bundle output directory (which is `output.dir || dirname(output.file!)`), the declarations will still be generated inside the bundle, albeit in a location not actually matching `declarationDir` that does not respect the absolute file path generated by typescript. 

Whereas when using `output.dir`, `declarationDir` is actually validated and filenames for declarations are resolved correctly.

For example, with `output.file = 'some-path/index.js'` and `declarationDir = 'types'`, the declaration files will be generated at `some-path/types/index.d.ts`, instead of `types/index.d.ts` which is the correct location (but invalid as it would be output the bundle directory). When `output.dir = 'some-path'` is used instead, validation fails as expected.

This can be fixed by just using `dirname(output.file)` in place of `output.dir` when `output.file` is specified, and including a validation check that `declarationDir` is valid when using `output.file`.

### DIFF
--- a/packages/typescript/src/index.ts
+++ b/packages/typescript/src/index.ts
@@ -178,20 +178,8 @@ export default function typescript(options: RollupTypescriptOptions = {}): Plugi
         if (outputOptions.dir) {
           baseDir = outputOptions.dir;
         } else if (outputOptions.file) {
-          // find common path of output.file and configured declation output
-          const outputDir = path.dirname(outputOptions.file);
-          const configured = path.resolve(
-            parsedOptions.options.declarationDir ||
-              parsedOptions.options.outDir ||
-              tsconfig ||
-              process.cwd()
-          );
-          const backwards = path
-            .relative(outputDir, configured)
-            .split(path.sep)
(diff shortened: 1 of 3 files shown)


lameuler (commented):
`baseDir`, which is used to resolve the pathname for `this.emitFile` using `path.relative(baseDir, id)`, should just be the bundle output directory since id is the absolute file path we want to emit.

shellscape:
Thanks for the PR. Since this will change some expectations around outputs, we'll have to mark this as a breaking change.

shellscape (approved):

PR #1732, opened 2024-06-06
fix(babel): Update Rollup dependency

ESnark:

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description
`@rollup/plugin-babel@^6.0.4` peer depends on `rollup@^4.0.0` but `@rollup/pluginutils@5.0.1` doesn't.
Because `@rollup/plugin-babel` includes `@rollup/pluginutils`, it interteres with upgrading `rollup`.

This PR resolves a peer-requirements issue for projects using `@rollup/plugin-babel`.

### DIFF
--- a/packages/babel/package.json
+++ b/packages/babel/package.json
@@ -67,7 +67,7 @@
   },
   "dependencies": {
     "@babel/helper-module-imports": "^7.18.6",
-    "@rollup/pluginutils": "^5.0.1"
+    "@rollup/pluginutils": "^5.1.0"
   },
   "devDependencies": {
     "@babel/core": "^7.19.1",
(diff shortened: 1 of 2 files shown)


Andarist:
This should be OK but to be fair - this is not a fixed range so ur package manager should already be able to install that new version of the utils package. You might have to force an update of the transitive dependency somehow though.

ESnark:
> This should be OK but to be fair - this is not a fixed range so ur package manager should already be able to install that new version of the utils package. You might have to force an update of the transitive dependency somehow though.

I found out ur right, my yarn package manager handle this well and there is no problem (still, I don't understand why yarn complains about it)

PR #1735, opened 2024-06-07
chore(node-resolve): remove is-builtin-module

43081j:
Node has shipped `builtinModules` for some time now, so we no longer need a third party package to do this.

Once the `engines` constraint is bumped in the `package.json`, we can also move to using the built-in `isBuiltin` function (available since 16.x).

This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no (existing tests should cover it)

Breaking Changes?

- [x] no

### DIFF
--- a/packages/node-resolve/src/index.js
+++ b/packages/node-resolve/src/index.js
@@ -1,7 +1,8 @@
 /* eslint-disable no-param-reassign, no-shadow, no-undefined */
 import { dirname, normalize, resolve, sep } from 'path';
 
-import isBuiltinModule from 'is-builtin-module';
+import { builtinModules } from 'module';
+
 import deepMerge from 'deepmerge';
 import isModule from 'is-module';
 
@@ -42,6 +43,8 @@ const defaults = {
   // TODO: set to false in next major release or remove
   allowExportsFolderMapping: true
 };
+const nodeImportPrefix = /^node:/;
+
 export const DEFAULTS = deepFreeze(deepMerge({}, defaults));
 
 export function nodeResolve(opts = {}) {
@@ -190,7 +193,7 @@ export function nodeResolve(opts = {}) {
       allowExportsFolderMapping: options.allowExportsFolderMapping
     });
 
-    const importeeIsBuiltin = isBuiltinModule(importee);
+    const importeeIsBuiltin = builtinModules.includes(importee.replace(nodeImportPrefix, ''));
     const resolved =
       importeeIsBuiltin && preferBuiltins
         ? {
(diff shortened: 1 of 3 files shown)


packages/node-resolve/src/index.js
```diff
@@ -190,7 +191,7 @@ export function nodeResolve(opts = {}) {
       allowExportsFolderMapping: options.allowExportsFolderMapping
     });
 
-    const importeeIsBuiltin = isBuiltinModule(importee);
+    const importeeIsBuiltin = builtinModules.includes(importee.replace('node:', ''));
```
tjenkinson, line 194:
```suggestion
    const importeeIsBuiltin = builtinModules.includes(importee.replace(/^node:/, ''));
```
Is a bit more precise

tjenkinson (commented):
Great that this is built in now TIL

Makes sense if we're not supporting <16

43081j:
FYI the engine constraint in here actually specifies 14 or something. Which is why I have used the built in modules list (available in 14), but not `isBuiltin` (available in 16) 

Just to be clear

If we can bump the engine constraint to 16, we could use the built in function

tjenkinson (approved):
Lgtm!

43081j:
any chance we can get this merged? 👀 

would be a good to pull it into a few dependents

MichaelDeBoey (approved):

benmccann (approved):
lgtm

packages/node-resolve/src/index.js
```diff
@@ -1,7 +1,8 @@
 /* eslint-disable no-param-reassign, no-shadow, no-undefined */
 import { dirname, normalize, resolve, sep } from 'path';
 
-import isBuiltinModule from 'is-builtin-module';
+import { builtinModules } from 'module';
```
jakebailey, line 4: Won't this cause a behavior change if you're bundling on one version of node, but a newer node has that module available? Or does it not matter because node will never add any new module without the `node:` prefix?

43081j: It's unlikely they will add any new modules without the prefix as far as I understand

Also the third party package also has a fixed list which won't always be accurate to the runtime node

43081j:
@shellscape any chance you can take a look at this? would be good to close it off

43081j:
any possibility of a review from someone here? would be super nice to merge this so it doesn't get left behind

guybedford (approved):

MichaelDeBoey (approved):

shellscape:
thanks!

PR #1744, opened 2024-06-30
fix(repo): type definitions not being exported correctly for esm modules

### LINKED ISSUE #1329: [typescript]: Default export type not being picked up with `module: "nodenext"`

### LINKED ISSUE #1541: Can't correctly import plugins into a TypeScript module

danielbayley:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

Picking up the baton from @Geylnu in #1578 …

I also ran into this. As mentioned elsewhere, [the types are wrong](https://arethetypeswrong.github.io/?p=%40rollup%2Fplugin-strip%403.0.4) in all Rollup plugins! Meaning they do not play nice with TypeScript.

More detailed explanations can be found [here](https://publint.dev/rules#export_types_invalid_format) and [here](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseCJS.md).

Fixes #1541.
Fixes #1329.

### DIFF
--- a/packages/alias/package.json
+++ b/packages/alias/package.json
@@ -16,7 +16,9 @@
   "main": "./dist/cjs/index.js",
   "module": "./dist/es/index.js",
   "exports": {
-    "types": "./types/index.d.ts",
+    "types": {
+      "import": "./dist/es/index.d.ts"
+    },
     "import": "./dist/es/index.js",
     "default": "./dist/cjs/index.js"
   },
--- a/packages/alias/package.json
+++ b/packages/alias/package.json
@@ -16,9 +16,6 @@
   "main": "./dist/cjs/index.js",
   "module": "./dist/es/index.js",
   "exports": {
-    "types": {
-      "import": "./dist/es/index.d.ts"
-    },
     "import": "./dist/es/index.js",
     "default": "./dist/cjs/index.js"
   },
@@ -70,7 +67,6 @@
     "rollup": "^4.0.0-24",
     "typescript": "^4.8.3"
   },
-  "types": "./types/index.d.ts",
   "ava": {
     "files": [
       "!**/fixtures/**",
--- a/packages/auto-install/package.json
+++ b/packages/auto-install/package.json
@@ -16,7 +16,9 @@
   "main": "./dist/cjs/index.js",
   "module": "./dist/es/index.js",
   "exports": {
-    "types": "./types/index.d.ts",
+    "types": {
+      "import": "./dist/es/index.d.ts"
+    },
     "import": "./dist/es/index.js",
     "default": "./dist/cjs/index.js"
   },
--- a/packages/auto-install/package.json
+++ b/packages/auto-install/package.json
@@ -16,9 +16,6 @@
   "main": "./dist/cjs/index.js",
   "module": "./dist/es/index.js",
   "exports": {
-    "types": {
-      "import": "./dist/es/index.d.ts"
-    },
     "import": "./dist/es/index.js",
     "default": "./dist/cjs/index.js"
   },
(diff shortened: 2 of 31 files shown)


shellscape:
Since this affects all plugins in the repo, this PR also needs to assert that every plugin has a types test.  Moving this away from sibling types is opening an entirely new can of worms, and types now need to be verified for both commonjs and ES tsconfig#module users. 

Had the last PR made it beyond CI, that would have been the ask as well.

danielbayley:
> Since this affects all plugins in the repo, this PR also needs to assert that every plugin has a types test. Moving this away from sibling types is opening an entirely new can of worms, and types now need to be verified for both commonjs and ES tsconfig#module users.
> 
> Had the last PR made it beyond CI, that would have been the ask as well.

@shellscape Sure, will circle back to this and look into it…

shellscape:
@danielbayley last call on this one

Septh:
> @danielbayley last call on this one

This PR makes Node choke when importing certain plugins (typescript and babel, at least) from an ESM package. Generating minimal `package.json` files next to the es bundles shouldn't be disabled.

image

Nevertheless, the types issue is a very, very long-standing one and needs to be fixed. I'm willing to take over if @danielbayley cannot pursue himself.

> Since this affects all plugins in the repo, this PR also needs to assert that every plugin has a types test. Moving this away from sibling types is opening an entirely new can of worms, and types now need to be verified for both commonjs and ES tsconfig#module users.

@shellscape > none of the packages currently have types tests; what would you like to see, exactly?

danielbayley:
> @danielbayley last call on this one

@shellscape I've been on holiday, then had COVID, and a bunch of other life-in-the-way stuff… So only just getting back into things. This is certainly still on my list to fix!

> Nevertheless, the types issue is a very, very long-standing one and needs to be fixed. I'm willing to take over if @danielbayley cannot pursue himself.

Yeah it needs to be resolved! @Septh You're welcome to lend a hand with this PR…

shellscape:
Closing as abandoned for now. Please feel free to re-open should you get back around to it.

PR #1746, opened 2024-07-03
fix(graphql): handle parentheses in fragment import file paths.

grxy:
## Rollup Plugin Name: `graphql`

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

N/A

### Description

This PR fixes an issue in the GraphQL plugin where parentheses in file paths are not properly replaced with a fragment's variable, leading to a parsing error.

The fix works by updating the require statement replacement regex from `/require\(([^)]+)\)/gi` to `/require\(["']([^"']+)["']\)/gi`. The original regex considered a closing parenthesis as the terminator of a require statement, when in fact, we should be looking at a quote character before a parentheses as the terminator. In addition, by excluding the leading and trailing quote from the regex capture group, we can remove the additional `replacePath` variable.

### DIFF
--- a/packages/graphql/test/fixtures/fragments-with-special-characters/query.graphql
+++ b/packages/graphql/test/fixtures/fragments-with-special-characters/query.graphql
@@ -1,0 +1,7 @@
+#import "./(parentheses)/fragment.graphql"
+#import "./[brackets]/fragment.graphql"
+
+query Query {
+  ...ParenthesesFragment
+  ...BracketsFragment
+}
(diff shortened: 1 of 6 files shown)


packages/graphql/test/fixtures/fragments-with-special-characters/query.graphql
```diff
@@ -0,0 +1,7 @@
+#import "./(parentheses)/fragment.graphql"
+#import "./[brackets]/fragment.graphql"
+
+query Query {
+  ...ParenthesesFragment
+  ...BracketsFragment
```
grxy, line 6: Brackets in filenames weren't actually breaking fragment imports, but I added this for extra confidence. Let me know if you'd like this to be removed.

shellscape (approved):
thanks!

PR #1750, opened 2024-07-24
fix(pluginutils): optimize `createFilter` and `normalizePath`

KermanX:

This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

`createFilter` and `normalizePath` are on the hot path. This PR tries to improve their performance.

Simple benchmarks (may not be very accurate, but IMO the direction is right):
- `filename.split(win32.sep).join(posix.sep)` -> `filename.replace(/\\/g, '/')` - [jsbenchmark.com](https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiI1c1M3QTBvN2kzZHpYNFNRNVF3djQiLCJjb2RlIjoiREFUQS5zcGxpdCgnXFxcXCcpLmpvaW4oJy8nKSIsIm5hbWUiOiJzcGxpdCArIGpvaW4gKG9sZCkiLCJkZXBlbmRlbmNpZXMiOltdfSx7ImlkIjoiNW1EUTJrcTl4UjhERFl5YW1fczEtIiwiY29kZSI6IkRBVEEucmVwbGFjZSgvXFxcXC9nLCAnLycpIiwibmFtZSI6InJlcGxhY2UiLCJkZXBlbmRlbmNpZXMiOltdfSx7ImlkIjoiOVI3Zy1Fd2JGdlJXd2dESEQxRk14IiwiY29kZSI6IkRBVEEucmVwbGFjZUFsbCgnXFxcXCcsICcvJykiLCJuYW1lIjoicmVwbGFjZUFsbCIsImRlcGVuZGVuY2llcyI6W119XSwiY29uZmlnIjp7Im5hbWUiOiJub3JtYWxpemVQYXRoIiwicGFyYWxsZWwiOnRydWUsImdsb2JhbFRlc3RDb25maWciOnsiZGVwZW5kZW5jaWVzIjpbXX0sImRhdGFDb2RlIjoicmV0dXJuIFsuLi5BcnJheSgxMDAwKS5rZXlzKCldLm1hcChfID0-IGAke01hdGgucmFuZG9tKCl9YC5zbGljZSgtNSkpLmpvaW4oJ1xcXFwnKSJ9fQ)
  Also, this PR hard-coded `win32.sep` and `posix.sep`. This is because this is faster, and the behavior of `normalizePath` should not change even when one day Windows starts to use `/` as the path separator.
- `/\0/.test(id)` -> `id.includes('\0')` - [jsbenchmark.com](https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiI1c1M3QTBvN2kzZHpYNFNRNVF3djQiLCJjb2RlIjoiL1xcMC8udGVzdChEQVRBKSIsIm5hbWUiOiJyZWdleC50ZXN0IiwiZGVwZW5kZW5jaWVzIjpbXX0seyJpZCI6IjVtRFEya3E5eFI4RERZeWFtX3MxLSIsImNvZGUiOiJEQVRBLmluY2x1ZGVzKCdcXDAnKSIsIm5hbWUiOiJpbmNsdWRlcyhjaGFyKSIsImRlcGVuZGVuY2llcyI6W119LHsiaWQiOiI5UjdnLUV3YkZ2Uld3Z0RIRDFGTXgiLCJjb2RlIjoiREFUQS5tYXRjaCgvXFwwLykiLCJuYW1lIjoibWF0Y2gocmVnZXgpIiwiZGVwZW5kZW5jaWVzIjpbXX1dLCJjb25maWciOnsibmFtZSI6Ik1hdGNoIGBcXDBgIiwicGFyYWxsZWwiOnRydWUsImdsb2JhbFRlc3RDb25maWciOnsiZGVwZW5kZW5jaWVzIjpbXX0sImRhdGFDb2RlIjoicmV0dXJuICdBJy5yZXBlYXQoMjApKydcXDAnKydCJy5yZXBlYXQoMTApIn19)

Also, this PR creates a fast path for empty filters - Usually `createFilters` takes paths from plugins' options, which defaults to empty if not specified.


packages/pluginutils/src/normalizePath.ts
```diff
@@ -1,9 +1,7 @@
-import { win32, posix } from 'path';
-
 import type { NormalizePath } from '../types';
 
 const normalizePath: NormalizePath = function normalizePath(filename: string) {
-  return filename.split(win32.sep).join(posix.sep);
```
shellscape, line 6: going to have to disagree with this change. while that regex is valid, the explicitness of using `win32.sep` and `posix.sep` has semantic value for future maintenance. if you'd like to build your regex using those constants, I would approve the PR.

shellscape (changes requested):

packages/pluginutils/src/normalizePath.ts
KermanX: Thanks for your advice, now it's using `win32.sep` and `posix.sep`.

PR #1753, opened 2024-07-29
fix(pluginutils): improve regex performance

bluwy:

This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

n/a

### Description

Refactors some regex to use non-capturing groups (as captured groups are a little slower), and use `regexp.prototype.exec` instead of `string.prototype.match` for perf.

For the former change, it's hard to use an online tool to measure as the difference is quite small, but here's a local file you can copy using `tinybench` that you can run:

<details>
<summary>File</summary>

```js
import { Bench } from 'tinybench'

const b = new Bench({ iterations: 1000 })

const choices = ['FunctionDeclaration', 'ClassDeclaration', 'VariableDeclaration', 'Identifier']
const strs = Array.from({ length: 10 }).map(_ => choices[Math.floor(Math.random() * choices.length)])

b.add('non-capture', () => {
  strs.forEach(s => { /(?:Function|Class)Declaration/.test(s) })
})

b.add('capture', () => {
  strs.forEach(s => { /(Function|Class)Declaration/.test(s) })
})

await b.warmup();
await b.run();

console.table(b.table());
```

</details>

node18.20.3 results:
```
┌─────────┬───────────────┬─────────────┬────────────────────┬──────────┬─────────┐
│ (index) │   Task Name   │   ops/sec   │ Average Time (ns)  │  Margin  │ Samples │
├─────────┼───────────────┼─────────────┼────────────────────┼──────────┼─────────┤
│    0    │ 'non-capture' │ '5,088,404' │ 196.5252384844669  │ '±0.63%' │ 2544203 │
│    1    │   'capture'   │ '4,831,673' │ 206.96762775341406 │ '±0.67%' │ 2415837 │
└─────────┴───────────────┴─────────────┴────────────────────┴──────────┴─────────┘
```

For the latter change, I've benchmarked this at https://github.com/lukeed/polka/pull/210 which you can run the perf.link test for the results too.

### DIFF
--- a/packages/pluginutils/src/dataToEsm.ts
+++ b/packages/pluginutils/src/dataToEsm.ts
@@ -91,7 +91,7 @@ const dataToEsm: DataToEsm = function dataToEsm(data, options = {}) {
 
   let maxUnderbarPrefixLength = 0;
   for (const key of Object.keys(data)) {
-    const underbarPrefixLength = key.match(/^(_+)/)?.[0].length ?? 0;
+    const underbarPrefixLength = /^(_+)/.exec(key)?.[0].length ?? 0;
     if (underbarPrefixLength > maxUnderbarPrefixLength) {
       maxUnderbarPrefixLength = underbarPrefixLength;
     }
--- a/packages/pluginutils/src/attachScopes.ts
+++ b/packages/pluginutils/src/attachScopes.ts
@@ -66,7 +66,7 @@ const attachScopes: AttachScopes = function attachScopes(ast, propertyName = 'sc
       const node = n as estree.Node;
       // function foo () {...}
       // class Foo {...}
-      if (/(Function|Class)Declaration/.test(node.type)) {
+      if (/(?:Function|Class)Declaration/.test(node.type)) {
         scope.addDeclaration(node, false, false);
       }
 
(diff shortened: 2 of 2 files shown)


packages/pluginutils/src/dataToEsm.ts
```diff
@@ -91,7 +91,7 @@ const dataToEsm: DataToEsm = function dataToEsm(data, options = {}) {
 
   let maxUnderbarPrefixLength = 0;
   for (const key of Object.keys(data)) {
-    const underbarPrefixLength = key.match(/^(_+)/)?.[0].length ?? 0;
+    const underbarPrefixLength = /^(_+)/.exec(key)?.[0].length ?? 0;
```
benmccann, line 94: I would expect here and in the other that perhaps it'd be better to move the regex declaration outside the method declaration so that it doesn't have to be recompiled each method invocation and each iteration of this loop

bluwy: Something I've found over time is that it doesn't necessarily help with perf by hoisting it out. Engines these days are able to optimize it, and in most cases only yield very small gains. This was also discussed at https://github.com/lukeed/polka/pull/210

benmccann: the discussion there suggests the one place it matters is in loops, so it might still be worth changing here, but not in the other locations?

bluwy: Yeah perhaps we can try that here, but I'd prefer to have a benchmark on that first before making the change though

shellscape (approved):
Going to merge this. I think this is a micro optimization to begin with, so not going to ask for additional optimizations on top of that. Please keep in mind that these tools typically run in environments with enough computing power that micro perf improvements barely have an impact.

PR #1761, opened 2024-08-27
feat(swc): sets filename property, so to make additional swc features…

klassm:
## Rollup Plugin Name: swc

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

-

### Description

Background: When using the swc rollup plugin in vitest, debugging will not work. I assume this comes from the filename property not being set, as the documentation states:
```
The filename is optional, but not all of Swc's
functionality is available when the filename is unknown, because a subset of options rely on the filename for their functionality.
```

Setting a filename should not have any side effects, just make debugging and other features work.

### DIFF
--- a/packages/swc/src/module.ts
+++ b/packages/swc/src/module.ts
@@ -37,7 +37,8 @@ export function swc(input: Options = {}): Plugin {
 
       return transform(code, {
         ...swcOptions,
-        sourceMaps: true
+        sourceMaps: true,
+        filename: id
       });
     }
   };


klassm:
If someone has a hint on how to test that, I would also add a test. I've looked at the existing tests, but both `file` and `sourceMap` seem to be exactly the same after the change.

tada5hi:
LGTM

tada5hi (approved):

lukastaegert (approved):

lukastaegert (approved):
I agree with the assessment and think it is safe to merge.

PR #1768, opened 2024-09-20
fix(replace): The `preventAssignment` option is treated as a value to replace

Finesse:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

When the `values` option is not and the `preventAssignment` option is set, the plugin replaces `preventAssignment` in the transformed code with the according option value. This PR fixes this bug.


shellscape (approved):

shellscape:
@Finesse this is approved, but we need to get the snapshots updated first. please run `pnpm --filter @rollup/plugin-replace test --u` to update the snapshots, and merge the result.

Finesse:
@shellscape The command you provided doesn't work. This command works: `pnpm --filter @rollup/plugin-replace test -- -u`.

I've run this command and pushed the changes. Sorry for not checking it on my local machine. The `misc` and `sourcemaps` snapshots also changed after running the command, but they change the same way when I run the command on `master`, so I haven't added them to the PR.

PR #1743, opened 2024-06-30
chore(repo): use shell-emulator for better cross-platform compatibility

danielbayley:
## Rollup Plugin Name: `*`

This PR contains:

- [x] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] yes (_bugfixes and features will not be merged without tests_)
- [x] no

Breaking Changes?

- [ ] yes (_breaking changes will not be merged unless absolutely necessary_)
- [x] no

If yes, then include "BREAKING CHANGES:" in the first commit message body, followed by a description of what is breaking. 

List any relevant issue numbers:

### Description

I started work on a separate PR, but quickly ran into errors with the `prepare` `scripts` apparently not being cross-platform (`if` command not found—I am on macOS). Since this is a `pnpm` monorepo, and the `scripts` are very simple, why not make use of the [`shell-emulator`](https://pnpm.io/cli/run#shell-emulator) `.npmrc` config option (which uses [yarnpkg-shell](https://yarnpkg.com/features/scripting#portable-shell))? This also means the dependency on `del-cli` is no longer necessary.

PR #1782, opened 2024-10-03
Provide typings for both cjs and es builds

### LINKED ISSUE #1541: Can't correctly import plugins into a TypeScript module

Septh:
## Rollup Plugin Name: `all (except Beep)`

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

resolves #1541, #1578 
supersedes #1744 

### Description

As soon as you start using `"moduleResolution": "Node16"` (or `"NodeNext"`) in `tsconfig.json`, TypeScript complains that the default export for the plugins in this repo _"is not callable"_ because it _"has no call signatures"_.

This PR fixes this by providing each package with distinct typings for the ESM build and the CJS build. It is a follow-up of the work @danielbayley started on #1744.

@shellscape > this passes all the current tests. However, given your past comment [here](https://github.com/rollup/plugins/pull/1744#issuecomment-2198411407), I think you'd prefer some specific tests to be added... Could you please just put me on the right track for this?

### DIFF
--- a/packages/alias/package.json
+++ b/packages/alias/package.json
@@ -15,10 +15,16 @@
   "bugs": "https://github.com/rollup/plugins/issues",
   "main": "./dist/cjs/index.js",
   "module": "./dist/es/index.js",
+  "type": "commonjs",
   "exports": {
-    "types": "./types/index.d.ts",
-    "import": "./dist/es/index.js",
-    "default": "./dist/cjs/index.js"
+    "require": {
+      "types": "./dist/cjs/index.d.ts",
+      "default": "./dist/cjs/index.js"
+    },
+    "import": {
+      "types": "./dist/es/index.d.ts",
+      "default": "./dist/es/index.js"
+    }
   },
   "engines": {
     "node": ">=14.0.0"
@@ -65,7 +71,7 @@
     "rollup": "^4.0.0-24",
     "typescript": "^4.8.3"
   },
-  "types": "./types/index.d.ts",
+  "types": "./dist/cjs/index.d.ts",
   "ava": {
     "files": [
       "!**/fixtures/**",
--- a/packages/auto-install/package.json
+++ b/packages/auto-install/package.json
@@ -15,10 +15,16 @@
   "bugs": "https://github.com/rollup/plugins/issues",
   "main": "./dist/cjs/index.js",
   "module": "./dist/es/index.js",
+  "type": "commonjs",
   "exports": {
-    "types": "./types/index.d.ts",
-    "import": "./dist/es/index.js",
-    "default": "./dist/cjs/index.js"
+    "require": {
+      "types": "./dist/cjs/index.d.ts",
+      "default": "./dist/cjs/index.js"
+    },
+    "import": {
+      "types": "./dist/es/index.d.ts",
+      "default": "./dist/es/index.js"
+    }
   },
   "engines": {
     "node": ">=14.0.0"
@@ -68,7 +74,7 @@
     "rollup": "^4.0.0-24",
     "typescript": "^4.8.3"
   },
-  "types": "./types/index.d.ts",
+  "types": "./dist/cjs/index.d.ts",
   "ava": {
     "workerThreads": false,
     "files": [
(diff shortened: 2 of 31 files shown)


shellscape:
Great addition. Thank you for taking this on (and sorry for the delay in getting eyes on this. I'm in Florida and was hit by two hurricanes in the last few weeks)

You're correct. We could merge this as-is, but we'd have no protection against regressions that others might introduce. Ideally, we'd have a `.ts` file for each plugin that would use the types that were generated, and checked by tsc for extra assurance. Alternatively, we could snapshot the types file content in tests as a looser form of assurance. Both are lifts, so I understand if it's something you're hesitant to do. Please let us know what you think.

Septh:
I went with the first option. 😄

shellscape:
Impressive turn-around on that. A few things:

- what's the significance of the Node16 vs Node10 directories - just support for `type: module` in package.json? 
- the `current_package` setup seems to be causing some problems on CI. I'd probably suggest just using `from '../'` and it should pick up the right files

Septh:
> * what's the significance of the Node16 vs Node10 directories - just support for `type: module` in package.json?

The names refer to the `moduleResolution` strategy set in `tsconfig.json`. 

I'm testing the 4 possible cases : 
- `Node16` within a `type: "module"` directory
- `Node16` within a `type: "commonjs"` directory
- `Node10` (aka `Node`) with `module: "CommonJS"` in tsconfig
- and `Node10` with `module: "ESNext"`

> * the `current_package` setup seems to be causing some problems on CI. I'd probably suggest just using `from '../'` and it should pick up the right files

That was my first shot, but it didn't work with `Node16`. I did some builds with `--traceResolution` and it appears that TypeScript resolves `'../'` as a straight relative identifier in that case, rather going through the whole Node resolution algorithm. This made the tests mostly irrelevant.

I borrowed the idea of linking the package to itself to [node-resolve](https://github.com/Septh/rollup-plugins/blob/fix-typings/packages/node-resolve/test/node_modules), so if it's not working in CI maybe I did something wrong? I made symbolic links, was this the right choice?

Septh:
I read on SO that symlinks have to be relative to the root of the repo (makes sens). Let try again.

danielbayley:
> I read on SO that symlinks have to be relative to the root of the repo (makes sens). Let's try again.

@Septh FYI, there's a useful little utility for that: https://github.com/brandt/symlinks.

Septh:
@shellscape > any chance you could review my latest push, please?

shellscape:
@Septh taking a look now

shellscape:
Looks like CI is failing on Windows. Not entirely sure why, but unfortunately I don't have the time to triage this for you, and we can't merge until that's passing.

Septh:
> Looks like CI is failing on Windows. Not entirely sure why, but unfortunately I don't have the time to triage this for you, and we can't merge until that's passing.


Well, I'm not sure either what's going wrong. Maybe someone reading this can provide some help? 🙏

danielbayley:
> Looks like CI is failing on Windows. Not entirely sure why, but unfortunately I don't have the time to triage this for you, and we can't merge until that's passing.

@shellscape @Septh Looks like just some unrelated thing to do with the `git config` email… I had a similar problem in one of my worflows recently… The solutions was something like `${{ github.event.pusher.email }}`… So I will see if I can find the time to take a look at this and see if we can get this PR over the line, and finally fix those types!

Also, see my https://github.com/rollup/plugins/pull/1743#issuecomment-2543948873 @shellscape…

shellscape:
We're going to be moving the repo to ESM-only here in short order. Unfortunately a large portion of this PR won't be usable after the transition. Thanks for all of your work on this @Septh and I'll make sure to give you a shout out in the releases.

PR #1786, opened 2024-10-14
fix(estree-walker): bump estree-walker to v3

bashmish:

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

- estree-walker has been updated to v3 which is an ESM package
- I'm using v3, it works best with my TypeScript setup with ESM everywhere and JSDoc syntax for types
- the goal here is to update it across all tooling and minimize dependecies/duplicates


pnpm-lock.yaml
```diff
@@ -5774,6 +5780,8 @@ snapshots:
 
   '@types/estree@1.0.0': {}
 
+  '@types/estree@1.0.6': {}
```
bashmish, line 5783: I see some of these can be dedupe, but I only know one way which is to run `pnpm dedupe`, but it gives lots of other unrelated changes too, so probably best do it centrally and not in this PR

shellscape:
Thanks for the PR. Looks like this change set causes issues with CI. Please have a look.

bashmish:
This is puzzling me. I tried this locally in different ways, not only with `pnpm install --frozen-lockfile`, but also separately for the package in question, all works fine.

image

I also used the same pnpm as in the Actions `npm i -g pnpm@9.4.0`, no luck.

Looks like `@types/estree` dependency is not deduped optimally and hence I tried to fix manually the lock file, but I can't test this locally, I mean it works before and after. And running the build here requires a manual approv, so my test loops is very long, not sure how to process, just wanna hope it works this time in Actions.

bashmish:
I figured `pnpm --recursive build` gives this error. Can test locally now 🎉

packages/pluginutils/types/index.d.ts
```diff
@@ -66,7 +66,7 @@ export function dataToEsm(data: unknown, options?: DataToEsmOptions): string;
  * Extracts the names of all assignment targets based upon specified patterns.
  * @param param An `acorn` AST Node.
  */
-export function extractAssignedNames(param: BaseNode): string[];
```
bashmish, line 69: changes between `BaseNode` to `Node` seems to have a but of history here
I saw a few issues and PRs addressing them, seemed like a back and forth

according to recent https://github.com/Rich-Harris/estree-walker/issues/28, using `Node` is the way, at least from the walker perspective

hope it's OK also for `pluginutils`

packages/commonjs/package.json
```diff
@@ -90,4 +90,4 @@
       "!**/types.ts"
     ]
   }
-}
+}
```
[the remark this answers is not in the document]
bashmish: I had to run with `--no-verify` to prevent husky script breaking this... :(
but the check in Actions should be happy

shellscape:
closing as abandoned

PR #1787, opened 2024-10-15
refactor(pluginutils,node-resolve): replace `test` with `includes`

btea:
## Rollup Plugin Name: `{pluginutils,node-resolve}`

This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] yes (_bugfixes and features will not be merged without tests_)
- [ ] no

Breaking Changes?

- [ ] yes (_breaking changes will not be merged unless absolutely necessary_)
- [ ] no

If yes, then include "BREAKING CHANGES:" in the first commit message body, followed by a description of what is breaking. 

List any relevant issue numbers:

### Description
It seems that `includes` are much more efficient than `test`.  [jsbenchmark](https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiI1c1M3QTBvN2kzZHpYNFNRNVF3djQiLCJjb2RlIjoiL1xcMC8udGVzdChEQVRBKSIsIm5hbWUiOiJyZWdleC50ZXN0IiwiZGVwZW5kZW5jaWVzIjpbXX0seyJpZCI6IjVtRFEya3E5eFI4RERZeWFtX3MxLSIsImNvZGUiOiJEQVRBLmluY2x1ZGVzKCdcXDAnKSIsIm5hbWUiOiJpbmNsdWRlcyhjaGFyKSIsImRlcGVuZGVuY2llcyI6W119XSwiY29uZmlnIjp7Im5hbWUiOiJNYXRjaCBgXFwwYCIsInBhcmFsbGVsIjp0cnVlLCJnbG9iYWxUZXN0Q29uZmlnIjp7ImRlcGVuZGVuY2llcyI6W119LCJkYXRhQ29kZSI6InJldHVybiAnQScucmVwZWF0KDIwKSsnXFwwJysnQicucmVwZWF0KDEwKSJ9fQ)

PR #1788, opened 2024-10-16
fix(pluginutils): optimize `createFilter` the matching rules when regexp carry flags

btea:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] no


shellscape:
Thanks, could you roll this into #1787 instead of a separate PR? Both are small enough to be combined.

btea:
Okay.

btea:
i have updated it.

PR #1794, opened 2024-10-18
fix(dynamic-import-vars): normalize paths at runtime

### LINKED ISSUE #1791: [plugin-dynamic-import-vars] non-normalized paths no longer work
- Rollup Plugin Name: plugin-dynamic-import-vars
- Rollup Plugin Version: 2.1.4
- Rollup Version: *
- Operating System (or Browser): *
- Node Version: *
- Link to reproduction: https://stackblitz.com/edit/rollup-repro-blsfye?file=src%2Fmain.js

### Expected Behavior
Imports passed non-normalized paths resolve the bundled file

### Actual Behavior
Imports passed non-normalized paths produce a `Unknown variable dynamic import` error

### Additional Information

#1780 breaks any code that doesn't build normalized paths because `tinyglobby` always outputs normalized glob paths but the plugin doesn't touch the original code.

Given this code:
```javascript
import(`./../${myFile}.js`)
```

`tinyglobby` produces `../*.js` -> `../myFile.js`-> `case '../myFile.js'`, which now doesn't match `./../myFile.js`.

I think the only reasonable options are:
1. Revert
2. Ask `tinyglobby` to make normalization optional (will take time and maybe isn't even feasible/desired)
3. Normalize the path inside `__variableDynamicImportRuntime__` before the `switch` with [`path-normalize`](https://www.npmjs.com/package/path-normalize)

bearfriend:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: #1791 

### Description

Fixes #1791 by normalizing paths at runtime to match the glob results. I decided again using a port of node's `posix.normalize` to both avoid forcing dependencies on consumers and also because the existing restrictions of the plugin mean it was (relatively) easy to transform the paths into how we need them.


packages/dynamic-import-vars/test/snapshots/rollup-plugin-dynamic-import-vars.test.js.md
```diff
@@ -234,25 +297,3 @@ Generated by [AVA](https://avajs.dev).
     ␊
     export { importModule };␊
     `
-
-## no files in dir
```
bearfriend, line 238: Do I just add this back manually?

shellscape:
Really appreciate the work on this. However, there are still some failing tests and this ends up injecting more code into a bundle on the resulting end. I thought on this for a while this weekend, and we're going to revert this plugin back to the previous version's state, since the only benefit to the change was fewer dependencies, and the juice is just not worth the squeeze.

bearfriend:
I think the tests got into a weird state where they should have been passing.

Regardless, the changes were weirder than I thought they'd be. I support reverting.

Issue #1583, opened 2023-09-14
[@rollup/plugin-typescript] { compilerOptions: { module: "nodeNext" } } overwritten when coming from "extends" in tsconfig.json


jessevanassen:
- Rollup Plugin Name: @rollup/plugin-typescript
- Rollup Plugin Version: 11.1.3
- Rollup Version: 3.29.1
- TypeScript version: 5.2.2
- Operating System (or Browser): Ubuntu 22.04.3 LTS
- Node Version: 20.4.0
- Link to reproduction: https://stackblitz.com/edit/rollup-repro-su4nqt?file=tsconfig.json,tsconfig.base.json,rollup.config.js

### Expected Behavior
I have two typescript config files: tsconfig.base.json and tsconfig.json.
In the tsconfig.base.json I have `{ compilerOptions: { module: "nodeNext" } }`.
tsconfig.json extends tsconfig.base.json.

When I run the typescript config, I expect that it takes the `module: "nodeNext"` into account that I configured in the tsconfig.base.ts. 

### Actual Behavior
The plugin overwrites `{ compilerOptions: { module: "nodeNext" } }` with `{ compilerOptions: { module: "esNext" } }`. This produces the following error with TypeScript 5.2 and up (see also https://github.com/microsoft/TypeScript/pull/54567):
```
(!) Plugin typescript: @rollup/plugin-typescript TS5110: Option 'module' must be set to 'NodeNext' when option 'moduleResolution' is set to 'NodeNext'.
```

#### Cause of the bug
https://github.com/rollup/plugins/blob/typescript-v11.1.3/packages/typescript/src/options/tsconfig.ts#L155-L196
```ts
export const DEFAULT_COMPILER_OPTIONS: PartialCompilerOptions = {
  module: 'esnext',
  skipLibCheck: true
};

// ...

ts.parseJsonConfigFileContent(
  {
    ...tsConfigFile,
    compilerOptions: {
      ...DEFAULT_COMPILER_OPTIONS,
      ...tsConfigFile.compilerOptions
    }
  },
  // ...
);
```    

This block gets the JSON parsed tsConfigFile that's used to process the project (in the reproduction, thats `./tsconfig.json`), adds the default options to the `compilerOptions`, and passes it on to TypeScript to completely parse the file, fill in defaults, etc.

However, if the tsconfig.json contains an extends which defines the module, the tsconfig.json ends up like this:
```js
{
  "compilerOptions": {
    module: 'esnext',
[...]

PR #1812, opened 2024-11-08
Refactor readTsConfigFile to fully resolve nested "extends" properties

jelovac:

This PR contains:

- [x] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] no

### Description

Refactored readTsConfigFile to fully resolve nested "extends" properties.

This refactor updates readTsConfigFile to correctly handle nested "extends" properties by using TypeScript’s parsing mechanism instead of basic JSON reading. The new behavior returns a unified JSON object that consolidates all properties from extended tsconfig files.

This change addresses issue #1583, ensuring complete config resolution across extended tsconfigs while retaining existing behavior with minimal changes and no breaking modifications.

### DIFF
--- a/packages/typescript/src/options/tsconfig.ts
+++ b/packages/typescript/src/options/tsconfig.ts
@@ -74,10 +74,31 @@ function getTsConfigPath(ts: typeof typescript, relativePath?: string | false) {
  * @param tsConfigPath Absolute path to tsconfig JSON file.
  */
 function readTsConfigFile(ts: typeof typescript, tsConfigPath: string) {
-  const { config, error } = ts.readConfigFile(tsConfigPath, (path) => readFileSync(path, 'utf8'));
-  if (error) {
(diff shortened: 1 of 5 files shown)


jelovac:
The primary issue here is that the `compilerOptions` passed to `parseTypescriptConfig` are incompatible with TypeScript’s native compilerOptions. The type definitions differ, which necessitates the if/else block.

The else block is effectively a workaround, as we can’t destructure compilerOptions into the existingOptions argument of ts.parseJsonConfigFileContent due to the incompatible type definitions. To achieve that, a type guard would be required. Consequently, the compiler options were passed as a JSON object, which accepts any input but only parses recognized properties. While this workaround addresses one problem, it also introduces the issues we are now facing.

I’m closing this PR as I don’t have time to work on a more comprehensive fix. Moreover, it’s uncertain if a solution exists that wouldn’t require breaking changes.

PR #1838, opened 2025-01-17
refactor(pluginutils): use pathe to support browser environments

zhiyuanzmj:

This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

Uses [pathe](https://github.com/unjs/pathe) instead of Node's native path module to support browser environments

### DIFF
--- a/packages/pluginutils/package.json
+++ b/packages/pluginutils/package.json
@@ -63,6 +63,7 @@
   "dependencies": {
     "@types/estree": "^1.0.0",
     "estree-walker": "^2.0.2",
+    "pathe": "^2.0.1",
     "picomatch": "^4.0.2"
   },
   "devDependencies": {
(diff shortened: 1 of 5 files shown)


shellscape:
Without discussion on why this is needed, I'm not inclined to approve this. Why not just alias `path` to `pathe` within bundler configuration?

zhiyuanzmj:
Thanks for your reply.

I created a REPL supports vite plugins, but when plugin are using createFilter that's will be don't support browser.
https://repl.zmjs.dev/vue-macros/vue-jsx-vapor

No matter, this maybe not a popular demand. you can close it at will.

shellscape:
Yes I think a fork here would be more appropriate. Thank you for the PR none the less.

PR #1850, opened 2025-03-01
fix(commonjs): fix error when bundle contains require() of module with falsy `__esModule` export

easrng:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

when a file with a falsy export named `__esModule` is `require`d the bundle will throw `TypeError: Cannot redefine property: __esModule`. this fixes the issue by checking for an `__esModule` own property instead of checking truthyness.


shellscape (approved):
thanks!

PR #1886, opened 2025-07-26
feat(pluginutils): add `suffixRegex` & support multiple string

btea:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

To reduce unnecessary interactions between js and rust when executing plugins, rolldown expands the plugin hook into an object containing filter properties. [doc](https://rolldown.rs/guide/plugin-development#:~:text=Ideally%2C%20we%20want%20to%20be%20able%20to%20determine%20whether%20a%20plugin%20hook%20needs%20to%20be%20invoked%20at%20all%20without%20leaving%20Rust.%20This%20is%20why%20Rolldown%20augments%20Rollup%20plugin%27s%20object%20hook%20format%20with%20the%20additional%20filter%20property)

Rollup also supports this feature in version 4.38.0. https://github.com/rollup/rollup/pull/5882 

In order to facilitate the generation of filter condition parameters, #1865 added several auxiliary functions.

When I was refactoring the filtering conditions of vite's internal plugins, I found that some plugins' filtering often consisted of multiple conditions. In this case, it would be more convenient for functions such as `prefixRegex`/`exactRegex` to support passing in multiple strings. Therefore, I expanded the parameters.

At the same time, I found that some plug-ins are more suitable for file suffix matching, such as `.vue`, so I added a new `suffixRegex` function to facilitate this situation.

### DIFF
--- a/packages/pluginutils/README.md
+++ b/packages/pluginutils/README.md
@@ -291,7 +291,7 @@ prefixRegex('foo(bar)', 'i'); // /^foo\(bar\)/i
 
 ### suffixRegex
 
-Constructs a RegExp that matches a value that has the specified prefix. This is useful for plugin hook filters.
+Constructs a RegExp that matches a value that has the specified suffix. This is useful for plugin hook filters.
 
 Parameters: `(str: String | Array[...String], flags?: String)`<br>
 Returns: `RegExp`


shellscape:
I'm not inclined to review this without a thorough PR description.

btea:
Sorry, I have updated the description. Please review it at your convenience. Thank you!

shellscape (approved):
thanks!

PR #1893, opened 2025-08-18
fix(node-resolve): error thrown with empty entry

jolyndenning:

This PR contains:

- [X] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [X] no

### Description

This improves the error message displayed to the user when the entry is `""` and the `node-resolve` plugin is used.

Before this change, the node-resolve plugin throws an error when the entry is `""` that is not helpful for the user to understand the problem ( `Cannot read properties of undefined (reading \'match\')`). Contrastingly, when not using the node-resolve plugin, rollup's error message is helpful: `Could not resolve entry module "".`

This pull request avoids the node-resolve error so that the clearer error message is displayed to the user.

### DIFF
--- a/packages/node-resolve/src/index.js
+++ b/packages/node-resolve/src/index.js
@@ -152,7 +152,7 @@ export function nodeResolve(opts = {}) {
 
     const importSpecifierList = [importee];
 
-    if (importer === undefined && !importee[0].match(/^\.?\.?\//)) {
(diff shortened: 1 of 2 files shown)


shellscape:
Please run prettier locally. There are issues with the code in the PR.

internettrans:
formatted the code

PR #1894, opened 2025-09-04
Support node16+ `moduleResolution` in ESM directories through named exports

### LINKED ISSUE #1541: Can't correctly import plugins into a TypeScript module

Septh:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no - read on! :)

Breaking Changes?

- [x] no

Fixes #1541, #1578
Supersedes #1744, #1782

### Description
Since #1782 is still failing in CI and since I still have no idea how to fix it, this new PR is a shameful attempt to work around the issue by providing named exports for all plugins, in addition to the default export.

```js
// ❌ fails with `type="module"` and `moduleResolution="node16"`

// ✅ works everywhere
import { typescript } from '@rollup/plugin-typescript'
```

I've also added a note in each README.md, right below the usage example, that reads like this:

> [!NOTE]
> If your editor complains that _"XXX is not a function"_, then use the named export instead:
>
> ```js
> import { XXX } from '@rollup/plugin-XXX';
> ```

Because there is absolutely no code change except for an additional export of the main factory function, I don't think new tests are necessary here. Please let me know if you feel otherwise.

### DIFF
--- a/packages/alias/README.md
+++ b/packages/alias/README.md
@@ -67,6 +67,13 @@ export default {
 };
 ```
 
+> [!NOTE]
+> If your editor complains that _"alias is not a function"_, then use the named export instead:
+>
+> ```js
+> import { alias } from '@rollup/plugin-alias';
+> ```
+
 Then call `rollup` either via the [CLI](https://www.rollupjs.org/guide/en/#command-line-reference) or the [API](https://www.rollupjs.org/guide/en/#javascript-api). If the build produces any errors, the plugin will write a 'alias' character to stderr, which should be audible on most systems.
 
 ## Options
--- a/packages/alias/src/index.ts
+++ b/packages/alias/src/index.ts
@@ -62,7 +62,7 @@ function resolveCustomResolver(
   return null;
 }
 
-export default function alias(options: RollupAliasOptions = {}): Plugin {
+function alias(options: RollupAliasOptions = {}): Plugin {
   const entries = getEntries(options);
 
   if (entries.length === 0) {
@@ -114,3 +114,4 @@ export default function alias(options: RollupAliasOptions = {}): Plugin {
     }
   };
 }
+export { alias, alias as default };
--- a/packages/alias/types/index.d.ts
+++ b/packages/alias/types/index.d.ts
@@ -41,4 +41,5 @@ export interface RollupAliasOptions {
 /**
  * 🍣 A Rollup plugin for defining aliases when bundling packages.
  */
-export default function alias(options?: RollupAliasOptions): Plugin;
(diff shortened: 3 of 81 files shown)


shellscape:
Appreciate the additional effort here, but we need a different direction so we're not dealing with bandaids.

Node 20.19.0 released proper compatibility for importing/requiring EMS from CJS. Now that Node 18 is out of LTS, we can safely enforce Node 20.19.0 as the minimum version. That means there's no reason for this repo to ship support for both CJS and ESM anymore.

What the project needs is for every plugin to be authored in TS and updated to ship only ESM. That should nuke any TS compat issues as well as bring the repo up to date. (For what it's worth, this is also something that needs to happen on a per-package basis, not a single PR that updates the world at once)

Septh:
Well, I tried 😅

Agreed, this is the way to go. Any ETA about this? Do you need any help?

shellscape:
@Septh I'll try and tackle this at some point this week. I'm planning on going one plugin at a time. If you'd like to begin that process, please feel free. I do want to avoid a large PR with many plugins affected at once.

Septh:
Sorry for the late reply, I've been quite busy these last days.

> If you'd like to begin that process, please feel free.

Yes, I am willing to contribute! Maybe we should synchronize so we don't both work on the same plugins. Any suggestions as to which packages I could start with?

oddantfr:
> What the project needs is for every plugin to be authored in TS and updated to ship only ESM. That should nuke any TS compat issues as well as bring the repo up to date. (For what it's worth, this is also something that needs to happen on a per-package basis, not a single PR that updates the world at once)

wrong! `@rollup/plugin-node-resolve` works perfectly fine in esm and cjs. This PR is great, you guys just find reasons not to change anything. I remember having this issue 2 years ago and now here's approaching 2026 and most package are not even modern JS friendly... come on:)

PR #1729, opened 2024-06-03
feat(eslint): supports eslint v9

younggglcy:
## Rollup Plugin Name: `@rollup/plugin-eslint`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] yes (_bugfixes and features will not be merged without tests_)
- [ ] no

Breaking Changes?

- [x] yes (_breaking changes will not be merged unless absolutely necessary_)
- [ ] no

If yes, then include "BREAKING CHANGES:" in the first commit message body, followed by a description of what is breaking. 

List any relevant issue numbers: resolves #1708

PR #1905, opened 2025-10-07
feat(eslint): add support for ESLint v9

emersion:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

Update package.json to make eslint a peerDependency, so that the
user has control over which version to pick. Move the existing
eslint dependency to devDependencies so that tests work as expected.

Introduce a new test to ensure the plugin works with ESLint v9. To
do so without breaking v8 tests, install ESLint v9 under a different
alias, and use an ESM import mocking library to override the "eslint"
import.

The current directory needs to be changed during the ESLint v9 test
so that ESLint finds its config file properly. This requires ava
workerThreads to be disabled:

    TypeError [ERR_WORKER_UNSUPPORTED_OPERATION]: process.chdir() is not supported in workers

Ref https://github.com/rollup/plugins/pull/1729 which was a previous attempt at this.

### DIFF
--- a/packages/eslint/package.json
+++ b/packages/eslint/package.json
@@ -54,6 +54,7 @@
     "lint"
   ],
   "peerDependencies": {
+    "eslint": "^8.57.1||^9.0.0",
     "rollup": "^1.20.0||^2.0.0||^3.0.0||^4.0.0"
   },
   "peerDependenciesMeta": {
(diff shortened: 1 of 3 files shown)


emersion:
@shellscape, if you have some time to review this patch, that would be great :)

emersion:
By the way - I'd be more than happy to help out with reviews for the packages used by my company if that's deemed useful.

shellscape:
@emersion thanks for your patience and poking on this. If you'd like to help review that would be much appreciated. If you end up with more time for maintenance, we can add you to the repo as a maintainer.

emersion:
Cool! I'll be monitoring eslint, node-resolve and typescript then, feel free to ping me on PRs touching these!

Issue #1869, opened 2025-05-09
[@rollup/plugin-yaml] Hook Filter


Namchee:
- Rollup Plugin Name: `@rollup/plugin-yaml`
- Rollup Plugin Version: 4.1.2

### Feature Proposal

Implement [hook filter](https://github.com/rollup/rollup/pull/5882) for better compatibility with rolldown.

PR #1916, opened 2025-10-18
feat(pluginutils): add `createHookFilter()`

lazka:

This PR contains:

- [ ] bugfix
- [x] feature
- [x] refactor
- [x] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

- #1869 tried to port the yaml plugin, which would be easier with this, except the "extensions" filter there would still need to be checked in the handler, so it only solves one part of it.

### Description

When trying to port a plugin that uses `createFilter()` to hook filters it's hard keep the functionality 100% the same, as `createFilter()` does various pre-processing, like normalising paths and, resolving relative paths, exluding paths which contain \0, etc.

To make porting easier, provide a `createHookFilter()` which instead of returning a function, returns a filter object that can be assigned to "filter.id" or "filter.code", and behaves the same as if createFilter() was used and the returned function used in a hook.

To make sure that the implementation of `createHookFilter()` matches `createFilter()`, `createFilter()` now calls `createHookFilter()` internally and uses the same object to create the test function.

----

This was motivated by me trying to port a a plugin which used `createFilter()` to the new API for easy faster rolldown performance, and having a hard time figuring out how to port while making sure that nothing changed for existing users. Not that many new tests added, since all createFilter() calls now also go through the new code, but I can add more if wanted.

Note sure if this is something that would be accepted :) (and my typescript is wonky... keep that in mind)

### DIFF
--- a/packages/pluginutils/README.md
+++ b/packages/pluginutils/README.md
@@ -139,6 +139,54 @@ export default function myPlugin(options = {}) {
 }
 ```
 
+### createHookFilter
+
+Constructs a StringFilter object which can be passed to Rollup for module filtering. Provides the same filtering behavior as `createFilter()` but returns a declarative filter object instead of a function. The content of the
+returned object is not guaranteed to be stable between versions.
+
+Parameters: `(include?: <picomatch>, exclude?: <picomatch>, options?: Object)`<br>
+Returns: `StringFilter`
+
+_Note: Please see the TypeScript definition for complete documentation of the return type_
+
+#### `include` and `exclude`
+
+Type: `String | RegExp | Array[...String|RegExp]`<br>
+
+A valid [`picomatch`](https://github.com/micromatch/picomatch#globbing-features) pattern, or array of patterns. If `options.include` is omitted or has zero length, filter will return `true` by default. Otherwise, an ID must match one or more of the `picomatch` patterns, and must not match any of the `options.exclude` patterns.
+
+Note that `picomatch` patterns are very similar to [`minimatch`](https://github.com/isaacs/minimatch#readme) patterns, and in most use cases, they are interchangeable. If you have more specific pattern matching needs, you can view [this comparison table](https://github.com/micromatch/picomatch#library-comparisons) to learn more about where the libraries differ.
+
+#### `options`
+
+##### `resolve`
+
+Type: `String | Boolean | null`
+
+Optionally resolves the patterns against a directory other than `process.cwd()`. If a `String` is specified, then the value will be used as the base directory. Relative paths will be resolved against `process.cwd()` first. If `false`, then the patterns will not be resolved against any directory. This can be useful if you want to create a filter for virtual module names.
+
+#### Usage
+
+```js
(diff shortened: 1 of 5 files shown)


shellscape:
hey thanks for the PR. is this something specific for Rolldown? it's the first time I've seen a proposal for the function/need, and I'm not aware of plugins that are filtering on hooks as yet. At least not in this repo. I'd like to get some more information from you before we expand the API of the pluginutils package.

lazka:
From what I understand this is available in both rollup and rolldown, see https://rollupjs.org/plugin-development/#build-hooks and scroll down to `filter` for rollup, and https://rolldown.rs/apis/plugin-hook-filters for rolldown. It was added to rollup in https://github.com/rollup/rollup/pull/5882.

The gist of it is that by passing the filter to the bundler declaratively, the bundler doesn't need to call out to plugins that often, which in the case of rust<->js saves a bit of time. I'd assume that in case of rollup there is no performance difference, as the code just moved from pluginutils to rollup core.

edit: I've noticed that I regressed the case when a regex has a sticky flag, I've added two tests + a fix.

edi2: I noticed that rollup does post-process patterns for hook filters too, but doesn't allow control how globs are resolved, so I changed things to convert picomatch to regex first, so the patterns are used in rollup as is.

shellscape:
Yes I'm aware of them being supported in Rollup proper. What I'm missing here is this pattern being used in plugins. I know that we don't have any plugins in this repo which are supporting or passing hook filters. Is that something commonly used in rolldown plugins or community plugins where we see the pattern used frequently in an ad hoc manner?

What I'm trying to determine is whether this belongs in pluginutils (which is meant to be used by plugins, and not build configs etc) or a third party convenience lib.

lazka:
I could see it replacing most `createFilter()` calls in this monorepo for example. For example for the url plugin: https://github.com/lazka/plugins/commit/ae5f4fb76a37154ee4598c7c865a9f1ae2d43b75?w=1

This would give better rolldown performance (8% build time reduction for me) with a few lines changed while not changing behavior for users.

shellscape:
I'm sorry but your reply doesn't help to answer the questions that I'm after. A repo-wide refactor for a project which isn't a direct dependent is just out of the question as we simply do not have enough active maintainers.

I think at this time it would be more beneficial to create a pluginutils for rolldown specifically, rather than adding that method to the pluginutils package here.

lazka:
I only use plugins from this repo, so creating an external tool would not help me. Feel free to close then.

shellscape:
Fair enough. The criteria for adding to the pluginutils package is that it's something that's widely used by plugins. At this time, it doesn't fit that criteria. If Rolldown has specific improvements it wants to make and submit PRs to individual plugins, and we end up with repetition across those improvements, then it would be warranted.

lazka:
It's a bit of a chicken and egg problem imo, plugins will not move to the new API if it's difficult to do so and they have to replicate undocumented pluginutils internals. But I don't maintain any plugins so that's just a guess on my end.

PR #1921, opened 2025-10-21
fix(pluginutils): only normalize/parse patterns once in createFilter

lazka:

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

When passing patterns to createFilter() it would normalize and parse the patterns on every call to the returned filter function.

Just do it once at the start instead.

Something I've noticed while working on #1916.

### DIFF
--- a/packages/pluginutils/src/createFilter.ts
+++ b/packages/pluginutils/src/createFilter.ts
@@ -26,19 +26,16 @@ function getMatcherString(id: string, resolutionBase: string | false | null | un
 const createFilter: CreateFilter = function createFilter(include?, exclude?, options?) {
   const resolutionBase = options && options.resolve;
 
-  const getMatcher = (id: string | RegExp) =>
-    id instanceof RegExp
-      ? id
-      : {
-          test: (what: string) => {
-            // this refactor is a tad overly verbose but makes for easy debugging
-            const pattern = getMatcherString(id, resolutionBase);
-            const fn = pm(pattern, { dot: true });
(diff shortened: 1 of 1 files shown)


shellscape (changes requested):
Thanks for the PR. Could you please add a test that performs multiple calls to getMatcher in succession, and add the PR number in the test description? That will help protect against future possible regressions.

lazka:
Thanks for the review. Sadly I don't quite understand what type of test you have in mind. Do you want me to move getMatcher to the module level and export it, and then test it? Though it still behaves the same, so there is no way to test the behaviour change.

Adding a spy on picomatch would makes sense, but I couldn't find a way to do that.

shellscape:
No worries. Your PR is changing the code so that instead of creating the matcher on every invocation of `test`, it's creating it once in the scope above, and every invocation of `test` uses the same matcher. To prevent against regressions of that change, we need a test in place which asserts that `test` is using the single matcher. Spitballing here - that could be as simple as exporting `getMatcher` individually, and using a symbol attached to the pattern to assert that each invocation of the resulting `test` method is using the same one. (It might be enough to copy and past the above to an agent and see what one comes up with)

Changes like this one are very susceptible to regressing in a project like this. We use tests to prevent that from happening.

shellscape:
Closing as abandoned.

Issue #1924, opened 2025-10-23
[@rollup/plugin-commonjs] `import { createRequire } from 'node:module'` is injected automatically, breaking codes for edge runtimes


sapphi-red:
- Rollup Plugin Name: commonjs
- Rollup Plugin Version: 28.0.7+
- Rollup Version: 4.52.5
- Operating System (or Browser): Windows, Linux
- Node Version: 22.18.0
- Link to reproduction _(⚠️ read below)_: https://stackblitz.com/edit/rollup-repro-ddvmzobf?file=rollup.config.js

### Expected Behavior
`import { createRequire } from 'node:module'` is not injected unless allowed by an option.

### Actual Behavior
`import { createRequire } from 'node:module'` is injected automatically.

### Additional Information
Edge runtimes like workerd and vercel edge runtime does not support `module.createRequire` ([workers docs](https://developers.cloudflare.com/workers/runtime-apis/nodejs/#supported-nodejs-apis), [vercel edge runtime docs](https://vercel.com/docs/functions/runtimes/edge#compatible-node.js-modules)). It isn't possible to polyfill this function as well.
This means the code generated cannot be executed by those runtimes.

related: #1909, https://github.com/vitejs/vite/issues/20983, https://github.com/vitejs/vite-plugin-react/pull/929#discussion_r2446461476, https://github.com/wakujs/waku/pull/1731

shellscape:
Sounds like we need to introduce an option to disable injecting that. 

@CharlieHelps please investigate what steps would be necessary

shellscape:
@sapphi-red please see and comment on the outline above

sapphi-red:
@CharlieHelps 

> injectCreateRequireForNodeBuiltins: boolean
> 
> * Default: true (keeps current behavior; no breaking change).

In a way #1909 was a breaking change and should have been an opt-in. I think the default should be `false`.

Other than that, I think it's good.

shellscape:
@sapphi-red unfortunately Charlie has a limitation right now where only maintainers can trigger him. 

#1900 was a chicken-egg bug. it was previously incorrect, so relying on incorrect behavior feels like a breaking change when it's fixed. Always sticky to know where the line is drawn. 

@CharlieHelps please see the previous comment from @sapphi-red. also consider https://github.com/rollup/plugins/pull/1930 which was recently submitted as a fix, though I don't think it nails what we're after here. Still, please include the changes in that PR and re-evaluate. Please don't be overly agreeable.

[...]

PR #1930, opened 2025-10-24
fix(commonjs): handle `node:` builtins with `strictRequires: auto`

### LINKED ISSUE #1929: [commonjs] "default" is not exported by some node core modules with `strictRequires: 'auto'`
- Rollup Plugin Version: v28.0.8
- Rollup Version: 4.52.4
- Link to reproduction: https://stackblitz.com/edit/rollup-repro-w9uxyy67?file=src%2Fmain.js,rollup.config.js

### Expected Behavior

With this config:

```
commonjs({
  strictRequires: 'auto',
}),
```

we now have an error when importing default export of some node core modules:

```ts
new (require('node:stream').Readable)({});
// throws RollupError: "default" is not exported by "node:stream?commonjs-external"
```

This worked fine with v28.0.6

### Additional Information

This is code from `@vue/server-renderer`, which means that we spotted the issue in a renovate PR upgrade for Nuxt: https://github.com/nuxt/nuxt/pull/33469.

danielroe:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

resolves #1929

### Description

this fixes an issue introduced in #1909, which added special handling for `node:` builtins to avoid eager loading in wrapped modules...


shellscape:
Thanks for the PR. Please take some time to describe what your changes do differently. 

We're also discussing this in https://github.com/rollup/plugins/issues/1924

danielroe:
my apologies!

charlie's summary in the linked issue and the code comment in this PR cover what i was intending, but i'll write in more detail when i'm back at my computer

packages/commonjs/src/resolve-require-sources.js
```diff
@@ -196,16 +196,21 @@ export function getRequireResolver(extensions, detectCyclesAndConditional, curre
           // Special-case external Node built-ins to be handled via a lazy __require
           // helper instead of hoisted ESM imports when strict wrapping is used.
           const isExternalWrapped = isWrappedId(dependencyId, EXTERNAL_SUFFIX);
-          if (
-            parentMeta.initialCommonJSType === IS_WRAPPED_COMMONJS &&
-            !allowProxy &&
-            isExternalWrapped
-          ) {
+          let resolvedDependencyId = dependencyId;
+          if (parentMeta.isCommonJS === IS_WRAPPED_COMMONJS && !allowProxy && isExternalWrapped) {
```
danielroe, line 200: 1. Previously, we checked `parentMeta.initialCommonJSType === IS_WRAPPED_COMMONJS` instead of `parentMeta.isCommonJS === IS_WRAPPED_COMMONJS`. With strictRequires: 'auto', I believe the initial type (before cycle detection) can differ from the final type.

packages/commonjs/src/resolve-require-sources.js
```diff
@@ -196,16 +196,21 @@ export function getRequireResolver(extensions, detectCyclesAndConditional, curre
...
               parentMeta.isRequiredCommonJS[dependencyId] = isCommonJS;
             }
+          } else if (isExternalWrapped && !allowProxy) {
+            // If the parent is not wrapped but the dependency is a node: builtin external,
+            // unwrap the EXTERNAL_SUFFIX so it's treated as a normal external.
+            // This avoids trying to load the lazy __require proxy for non-wrapped contexts.
+            const actualExternalId = unwrapId(dependencyId, EXTERNAL_SUFFIX);
+            if (actualExternalId.startsWith('node:')) {
+              resolvedDependencyId = actualExternalId;
+            }
```
danielroe, line 213: 2. `node:` builtins with the EXTERNAL_SUFFIX were always loaded with `getExternalBuiltinRequireProxy()` which only exports `__require()`. However, when the parent module is not wrapped (no cycles detected with strictRequires: 'auto'), the generated code tried to import a default export that didn't exist.

(so in this case we can remove the suffix)

shellscape:
thanks!

PR #1938, opened 2025-10-29
fix(replace): update delimiters to respect valid js identifier chars

danielroe:

This PR contains:

- [x] bugfix

Are tests included?

Breaking Changes?

- [x] no

### Description

the word boundary delimiter (`\b`) would not correctly handle some valid JS identifiers (like `$`, leading to an incorrect replacement of `someterm$` being replaced by a rule for `someterm`), so I've updated delimiters to be more exhaustive.


shellscape:
@danielroe thanks for the PR! Some failing tests there in CI that will need attention before we can merge.

PR #1949, opened 2025-11-28
fix(typescript): avoid resolving definition files with arbitrary extensions

### LINKED ISSUE #1858: `@rollup/plugin-typescript` seems not recognizing `.d.[ext].ts` style definition files?
- Rollup Plugin Version: 12.1.2
- Rollup Version: 4.28.1
- Node Version: v20.18.3
- Link to reproduction: https://stackblitz.com/edit/rollup-repro-w7wg6yag?file=src%2Fmain.ts

### Expected Behavior

Bundle successfully built with `npm run build`.

### Actual Behavior

```
src/main.ts → dist...
[!] RollupError: src/chromaprint_bg.d.wasm.ts (2:9): Expected ',', got '?' (Note that you need plugins to import files that are not JavaScript)
src/chromaprint_bg.d.wasm.ts (2:9)
1: export default function (
2:   imports?: WebAssembly.Imports
            ^
3: ): Promise<WebAssembly.Instance>;
    at Object.getRollupError (/home/projects/rollup-repro-w7wg6yag/node_modules/rollup/dist/shared/parseAst.js:285:41)
    at ParseError.initialise (/home/projects/rollup-repro-w7wg6yag/node_modules/rollup/dist/shared/rollup.js:15820:40)
    at convertNode (/home/projects/rollup-repro-w7wg6yag/node_modules/rollup/dist/shared/rollup.js:17714:10)
    at convertProgram (/home/projects/rollup-repro-w7wg6yag/node_modules/rollup/dist/shared/rollup.js:16957:12)
    at Module.setSource (/home/projects/rollup-repro-w7wg6yag/node_modules/rollup/dist/shared/rollup.js:18695:24)
    at async ModuleLoader.addModuleSource (/home/projects/rollup-repro-w7wg6yag/node_modules/rollup/dist/shared/rollup.js:22265:13)
…
```

### Additional Information

I was trying to integrate WASM in a TS project, utilizing TypeScript 5.0 `allowArbitraryExtensions` compiler option (where I provide defs with a `.d.wasm.ts` file) and got errors on it. It seems that the def file was not transformed or dropped by TypeScript plugin and went to later stages.
I guess maybe it's `@rollup/plugin-typescript` not recognizing the def file? (though I did not actually dig into the code to verify the idea 🤦)

ychavoya:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: resolves #1858

### Description

This PR updates the regular expression used to detect declaration files (`*.d.ts`) to include declaration files for arbitrary extensions, such as the ones allowed with the [`allowArbitraryExtensions` compiler option](https://www.typescriptlang.org/tsconfig/#allowArbitraryExtensions) (`*.d.*.ts`). By including these in the regex, they are not resolved and therefore prevent the error displayed in #1858.

Also, in order to add a test for this behavior, the typescript version in the dev dependencies was upgraded to v5 and therefore some other pieces of code had to be changed.

### DIFF
--- a/packages/typescript/test/fixtures/arbitrary-dts/module.d.custom.ts
+++ b/packages/typescript/test/fixtures/arbitrary-dts/module.d.custom.ts
@@ -1,4 +1,4 @@
 declare const CustomModule: {
-    x: Number;
+    x: number;
 }
 export default CustomModule;


packages/typescript/src/outputFile.ts
```diff
@@ -38,7 +38,7 @@ export function isTypeScriptMapOutputFile(name: string): boolean {
  * Checks if the given OutputFile represents some declaration
  */
 export function isDeclarationOutputFile(name: string): boolean {
-  return /\.d\.[cm]?ts$/.test(name);
+  return /\.d(\..+)?\.[cm]?ts$/.test(name);
```
emersion, line 41: I was wondering whether an arbitrary string is allowed between `.d.` and `.ts`, or if only a restricted set of characters are allowed. It turns out any arbitrary string is allowed here:

https://github.com/microsoft/TypeScript/blob/01c23d68b1113a4b71acb8484f1bf78a98307933/src/compiler/parser.ts#L10568 

tl;dr this regexp sounds correct to me.

packages/typescript/test/test.js
```diff
@@ -1538,8 +1538,9 @@ test.serial('correctly resolves types with nodenext moduleResolution', async (t)
   const code = await getCode(bundle, outputOptions);
 
   t.true(code.includes('var bar = foo'), code);
-  t.is(warnings.length, 1);
-  t.is(warnings[0].code, 'UNRESOLVED_IMPORT');
+  t.is(warnings.length, 2);
+  t.is(warnings[0].pluginCode, 'TS5110');
```
emersion: What is this warning? Is it about the `module` value not being aligned with the `moduleResolution` value?

Shouldn't we resolve it instead?

packages/typescript/src/options/tsconfig.ts
```diff
@@ -122,7 +122,7 @@ function setModuleResolutionKind(parsedConfig: ParsedCommandLine): ParsedCommand
   };
 }
 
-const configCache = new Map() as typescript.ESMap<string, ExtendedConfigCacheEntry>;
+const configCache = new Map() as Map<string, ExtendedConfigCacheEntry>;
```
emersion: Nit: could we avoid the cast here by using `new Map<string, ExtendedConfigCacheEntry>()` instead?

packages/typescript/test/test.js
ychavoya: Resolved! I was not sure if I should update this to include the warning, or this other statement https://github.com/rollup/plugins/pull/1949/changes#diff-9b47757e1347eae5ab5cefedce0669826e925f9b073a02815b9507b782f2211bL1540 since changing the module changes how the code is output in the test.

packages/typescript/src/options/tsconfig.ts
ychavoya: Thanks, I missed that when changing it

emersion:
These new changes look good! Can you squash them into the commits that introduced what they're fixing?

packages/typescript/test/test.js
emersion: What the test cares about is the unresolved import, so I'd say it's better if we avoid any other unnecessary noise :P

ychavoya:
Done, I squashed the changes into the typescript upgrade, so now there are only the 3 commits I initially added. Let me know if you want them combined as well

emersion (approved):
LGTM!

emersion:
@shellscape, would you mind having a look?

shellscape:
@emersion I would advise not adding an approving review to PRs in repos which you do not maintain. 

emersion:
@shellscape I thought my reviews were welcome according to https://github.com/rollup/plugins/pull/1905#issuecomment-3453994323

If not, no problem at all, just means less work for me :)  

(I do wonder though why I've been added as org member.)

shellscape:
@emersion providing your takes on the changes yes, approvals as a non-maintainer not so much.

emersion:
Would you mind explaining why? An approval is a clear statement that I've reviewed the changes, and GitHub makes it clear that the approval doesn't unlock the merge (the checkmark is greyed out).

I will refrain from approving in the future.

shellscape:
> Would you mind explaining why?

Mostly confusion. Contribs submitting PRs see an approval from userland (non-maintainers) and wonder why it's not moving forward, when those approvals have little/no bearing on whether a PR will get merged. Peer reviews are great, but I'd keep them limited to `Comment` unless you're a maintainer. 

I looked into why you're in the org, and you're on the Plugin Contributors team, which is a small team of people who occasionally contribute with limited permissions on the org. 

Going to hide these as off-topic in the PR just to clean it up.

shellscape:
@ychavoya please have a look at this review https://github.com/rollup/plugins/pull/1949#issuecomment-3906358016 and address the must-fix items.

ychavoya:
@shellscape The timeout error is also present in master, though it doesn't always run out of memory, especially outside of CI. However I managed to fix it :)

The problem was due to different `rollup` versions inside of `packages/typescript` and in the root. This is because version `^4.0.0-24` in the typescript package did not match with `4.x.x` versions (due to the `-24`), so it was stuck on that pre-release version while the root had a release version `4.52.5`. This caused a specific type (`RollupBuild`) to not match type definitions, which apparently got some test files to silently fail but never exit (and ava passed without running them after a while).

Note that this same problem of two different versions of rollup `^4.0.0-24` and `4.x.x` is also present on other packages. I will create a separate PR to fix the rest of the packages (or one per package if you prefer)

Most other issues the AI found should also be fixed, let me know if anything else needs changes!

PR #1951, opened 2025-12-01
fix(util): remove Rollup type dependency from shared test helper

koal44:
## Rollup Plugin Name: `repo`

This PR contains:

- [x] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] yes (_existing plugin test suites already cover this helper; no new tests added_)
- [ ] no

Breaking Changes?

- [x] no

### Description

A shared test helper at `util/test.js` imported Rollup types from the repo root, which meant plugin tests were effectively using whatever Rollup version resolved in the root. With no pinned version there, transitive dependencies (notably `vitest`) dictated the Rollup version seen by plugin tests.

This mismatch (root resolving 4.52.x, TypeScript plugin tests expecting 4.0.0-24) produced type noise in `declarations.ts` and `tslib.ts`. It also caused the TypeScript plugin's AVA run to stall around 77 tests; AVA still reported a pass, but 13 tests were silently skipped.

#### Changes
- Rewrote the helper as `util/test.ts`, using minimal generic interfaces (`BundleLike`, `OutLike`) instead of importing Rollup types.
- Added a small JS shim at `util/test.js` so existing plugin test suites keep working unchanged.
- Matched the behavior of the original JS helper, including ESM support.

#### Testing
- Ran test suites for all plugins that use `util/test.js`: `babel`, `commonjs`, `data-uri`, `graphql`, `html`, `image`, `json`, `legacy`, `multi-entry`, `node-resolve`, `sucrase`, `typescript`, `wasm`, `yaml`.
- `@rollup/plugin-typescript` now runs its full suite (90/90) with no timeouts.
- `wasm` still shows the same intermittent flake seen on `master`, suggesting it's unrelated to this PR.

#### Lint
TS-ESLint didn't play well with the `test.{js,ts}` twin, so the shim is eslint-ignored for now. Once the shim is gone and callers use the TS helper directly, this disappears.

#### Commit Note
There's a small intermediate commit to carry the original helper's history across the `*.js → *.ts` rename.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -12,3 +12,4 @@ packages/node-resolve/test/fixtures/package-json-in-path
 # temporary workaround for TypeScript as it doesn't support "Arbitrary module namespace identifier names"
 # https://github.com/microsoft/TypeScript/issues/40594
 packages/json/test/fixtures/arbitrary/main.js
+util/test.js
--- a/packages/typescript/test/tslib.ts
+++ b/packages/typescript/test/tslib.ts
@@ -23,7 +23,7 @@ test.serial('supports overriding tslib with a custom path', async (t) => {
   });
   const code = await evaluateBundle(bundle);
 
-  t.is((code as any).myParent.baseMethod(), 'base method');
+  t.is(code.myParent.baseMethod(), 'base method');
 });
 
 test.serial('supports overriding tslib with a custom path in a promise', async (t) => {
@@ -38,7 +38,7 @@ test.serial('supports overriding tslib with a custom path in a promise', async (
   });
   const code = await evaluateBundle(bundle);
 
-  t.is((code as any).myParent.baseMethod(), 'base method');
+  t.is(code.myParent.baseMethod(), 'base method');
 });
 
 test.serial('fails on bad tslib path', async (t) => {
--- a/util/test.js
+++ b/util/test.js
@@ -1,0 +1,9 @@
+const tsModule = require('./test.ts');
+
+exports.getCode = tsModule.getCode;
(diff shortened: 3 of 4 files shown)


shellscape:
@koal44 which agent was responsible for the code in this PR?

koal44:
I'll take it you were impressed enough to ask, which is nice, because considerable work went into the PR. As for the agent responsible: that’s me. I wrote the code, set the direction, and ran the tests. I also worked with an LLM during the process, and they deserve some recognition too.

shellscape:
I see. Thanks for the reply and PR.

It's curious that you adopted the pattern of using a js file to import a ts file, that's just not something a human would intuitively do. Combined with the description that was absolutely written, at least in part, by an LLM, and it's hard to believe this wasn't authored by an LLM in the majority. We're open to LLM work.and leverage it, but it must come packaged with a clear understanding of what is being modified and for what purpose. 

Since this PR is really only addressing a dep issue and test run (for which the LLM claims is timing out, but for which we have no evidence of in CI. https://github.com/rollup/plugins/actions/runs/18736385783/job/53443950552?pr=1923 for example the last run of tests against the plugin) this isn't a PR that I'm open to accepting at this time.

I'd encourage you to open an issue with reproduction next time to discuss before opening a PR.

koal44:
We don't yet live in a world where an LLM can produce the work in this PR. Your own experience with Charlie Bot should've make that clear. You've said more than once that the community wasn't contributing enough, so I offered a hand. What you do with that is up to you.

PR #1954, opened 2025-12-16
feat(babel)!: improve filtering capability & performance

davidtaylorhq:
## Rollup Plugin Name: `babel`

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] yes (_bugfixes and features will not be merged without tests_)
- [ ] no

Breaking Changes?

- [x] yes (_breaking changes will not be merged unless absolutely necessary_)
- [ ] no

If yes, then include "BREAKING CHANGES:" in the first commit message body, followed by a description of what is breaking. 

### Description

- drops support for rollup 1.x, which allows simplifying the implementation

- implements 'plugin hook filters' for improved performance under rolldown

- passes a new `code` argument to custom filter functions, so that consumers can add more advanced filtering logic

- allows custom filter functions to be async

PR #1956, opened 2026-01-12
feat(babel): add parallel processing via worker threads

davidtaylorhq:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: n/a

### Description

feat(babel): add parallel processing via worker threads

Add a `parallel` option that processes files concurrently using Node.js worker threads. This reduces build times for large projects where Babel transformation is a bottleneck. This is similar to the existing parallel behavior of `@rollup/plugin-terser`.

This required some fairly significant refactoring, because we can only pass serializable objects between the main thread and the worker threads. It also required changes to the plugin's own build config, so that we can generate a dedicated worker entrypoint.

Validations are added to ensure that unserializable config (e.g. inline babel plugins) cannot be used alongside the new parallel mode. For people using dedicated babel config files, this isn't a problem, because they are loaded directly by babel in the worker thread itself.

The worker threads do have a setup cost, so this only makes sense for large projects. In Discourse, enabling this parallel mode cuts our overall vite (rolldown) build time by about 45% (from ~11s to ~6s) on my machine.

---

This PR is based on #1954, which should ideally be merged first. Happy to rebase this one on `main` if the other PR isn't mergable.

For testing, I've pushed a built version of this branch to https://github.com/davidtaylorhq/plugins/tree/babel-parallel-built. In pnpm, it can be installed like:

```json
"@rollup/plugin-babel": "davidtaylorhq/rollup-plugins#babel-parallel-built&path:/packages/babel",
```


davidtaylorhq:
@Andarist @shellscape  sorry for the ping, but just wanted to check if this is something you'd consider looking at? The performance benefits are pretty massive for large projects using Babel.

NullVoxPopuli:
I'm really looking forward to this landing! I have a local tarball that I've been copying into my projects to have the benefits of this PR - and it's really good. I can no longer imagine using babel in rollup/vite/rolldown without the parallel mode <3

(usage without parallel mode is painfully slow)

packages/babel/src/index.js
```diff
@@ -110,77 +126,113 @@ function createBabelInputPluginFactory(customCallback = returnObject) {
...
-          throw new Error('Could not handle include or exclude with custom filter together');
-        }
-        const userDefinedFilter =
-          typeof customFilter === 'function' ? customFilter : createFilter(include, exclude);
-        filter = (id) => extensionRegExp.test(stripQuery(id).bareId) && userDefinedFilter(id);
+      }
+
+      const parallelWorkerCount = typeof parallel === 'number' ? parallel : 4;
+      workerPool = new WorkerPool(
+        new URL('./worker.js', import.meta.url).pathname,
```
Andarist: is this compatible with repo-wide supported node.js version?

packages/babel/src/workerPool.js
Andarist, line 1: have you considered using some existing worker pool libraries?

packages/babel/src/index.js
NullVoxPopuli: Would ci tell us?

packages/babel/src/workerPool.js
davidtaylorhq: I aimed to match the implementation of the existing Terser plugin, since I wasn't sure if adding new dependencies would be welcome. I'm certainly happy to explore it though! Did you have any particular libraries in mind?

[workerpool](https://www.npmjs.com/package/workerpool) looks like it could be a good bet 👀

packages/babel/src/index.js
davidtaylorhq: Worker Threads and `import.meta.url` are both supported in Node 14 👍

shellscape:

shellscape:
@davidtaylorhq looks like Github's buggy UX bit us on this and I can't see/approve the CI runs. please push a new commit to the branch to trigger that UX again.

NullVoxPopuli:
@shellscape you can close and re-open a PR to trigger that, usually -- no reason to make a contributor do a silly because GH is buggy <3

packages/babel/rollup.config.mjs
```diff
@@ -1,14 +1,37 @@
...
+  output: [
+    {
+      format: 'cjs',
+      dir: 'dist/cjs',
+      exports: 'named',
+      footer(chunkInfo) {
+        if (chunkInfo.name === 'index') {
+          return 'module.exports = Object.assign(exports.default, exports);';
+        }
+        return null;
```
Andarist, line 24: why do we need this now? I don't see any change in the index.js that would affect the package's shape

packages/babel/package.json
```diff
@@ -55,7 +55,7 @@
   "peerDependencies": {
     "@babel/core": "^7.0.0",
     "@types/babel__core": "^7.1.9",
-    "rollup": "^1.20.0||^2.0.0||^3.0.0||^4.0.0"
+    "rollup": "^2.0.0||^3.0.0||^4.0.0"
```
Andarist, line 58: @shellscape this required a major bump of this package - we should ensure that the automation will be able to bump this properly. I think that's achieved using some special tags in commit messages

packages/babel/src/workerPool.js
Andarist: I guess it's fine to keep it as-is if it matches the one in the terser plugin. It would be nice to extract this to a shared package but I don't know what's the proper drill for this. It probably could/should be extracted once this PR lands.

packages/babel/types/index.d.ts
```diff
@@ -23,7 +23,7 @@ export interface RollupBabelInputPluginOptions
    *   const filter = createFilter(include, exclude, {});
    * @default undefined;
    */
-  filter?: ReturnType<CreateFilter>;
+  filter?: (id: string, code: string) => Promise<boolean>;
```
Andarist: let's revert this

packages/babel/src/transformCode.js
```diff
@@ -16,18 +22,23 @@ export default async function transformCode(
     return null;
   }
 
-  let transformOptions = !overrides.config
+  let transformOptions = !overrides?.config
     ? config.options
-    : await overrides.config.call(ctx, config, {
```
Andarist, line 21: this would technically be a breaking change, right? we should call it out

packages/babel/types/index.d.ts
davidtaylorhq: This PR is based on https://github.com/rollup/plugins/pull/1954, which should ideally be merged first. Would you prefer that I drop the other one? (they both touch the same code, so it would require some unpicking to separate them)

packages/babel/src/transformCode.js
```diff
@@ -16,18 +22,23 @@ export default async function transformCode(
...
-    transformOptions = await finalizeOptions(transformOptions);
+  if (runPreflightCheck) {
+    await preflightCheck(error, babelHelpers, transformOptions);
   }
 
-  if (!overrides.result) {
+  transformOptions =
+    babelHelpers === BUNDLED
+      ? addBabelPlugin(transformOptions, bundledHelpersPlugin)
+      : transformOptions;
```
Andarist, line 39: Hm, so this now relies on `babelHelpers` input to skip the `addBabelPlugin` call in the `renderChunk`'s case, right?

packages/babel/package.json
shellscape: yeah including `BREAKING CHANGES` in the commit body or changing the PR title to reflect a major version in the conventional commit spec.

packages/babel/rollup.config.mjs
davidtaylorhq: I had to refactor `rollup.config.mjs` to add the worker entrypoint. So this code is re-implementing this functionality from the shared config which was being used before:

https://github.com/rollup/plugins/blob/7d16103b995bcf61f5af1040218a50399599c37e/shared/rollup.config.mjs#L28

packages/babel/src/index.js
```diff
@@ -110,77 +126,113 @@ function createBabelInputPluginFactory(customCallback = returnObject) {
...
+          return transformCode({
+            inputCode: code,
+            babelOptions: { ...babelOptions, filename },
+            overrides: {
+              config: overrides.config?.bind(this),
+              result: overrides.result?.bind(this)
+            },
+            customOptions,
+            error: this.error.bind(this),
+            runPreflightCheck: !skipPreflightCheck,
```
Andarist: IMHO, this conversion make the code slightly harder to follow. We should standardize on a single variable name for this

packages/babel/src/transformCode.js
Andarist: Ah, sorry - the function already comes prebound now. So there is no change here really. Those overrides can't be handled here in the parallel mode, right?

davidtaylorhq: It's not a breaking change, because the function is now bound to `ctx` earlier:

https://github.com/davidtaylorhq/rollup-plugins/blob/1a508bf83e656e7d2077906d7eb70d8560bbf014/packages/babel/src/index.js#L215-L216

packages/babel/src/index.js
```diff
@@ -110,77 +126,113 @@ function createBabelInputPluginFactory(customCallback = returnObject) {
...
+          if (!(await filter(filename, code))) return null;
+          if (filename === HELPERS) return null;
+
+          if (parallel) {
+            return workerPool.runTask({
+              inputCode: code,
+              babelOptions: { ...babelOptions, filename },
+              runPreflightCheck: !skipPreflightCheck,
+              babelHelpers
+            });
```
Andarist: q: I'm not sure what we actually get from `this.error` but maybe it's worth rechecking if we shouldn't "convert" the errors from this to `this.error` calls?

packages/babel/src/transformCode.js
davidtaylorhq: > Those overrides can't be handled here in the parallel mode, right?

Correct, and an error will be emitted if `parallel: true` is set alongside these options

packages/babel/src/index.js
```diff
@@ -110,77 +126,113 @@ function createBabelInputPluginFactory(customCallback = returnObject) {
...
-          extensions,
-          babelHelpers,
-          include,
-          filter: customFilter,
-          skipPreflightCheck,
-          ...babelOptions
-        } = unpackInputPluginOptions(pluginOptionsWithOverrides, this.meta.rollupVersion));
+    if (parallel) {
+      const parallelAllowed =
+        isSerializable(babelOptions) && !overrides?.config && !overrides?.result;
```
Andarist: perhaps it would be nice to move this validation to `unpackInputPluginOptions`? It would also be quite great if the thrown error message could mention the offending option.

packages/babel/src/index.js
```diff
@@ -259,7 +311,16 @@ function createBabelOutputPluginFactory(customCallback = returnObject) {
           }
         }
 
-        return transformCode(code, babelOptions, overrides, customOptions, this);
+        return transformCode({
```
Andarist, line 365: q: why we don't attempt to parallelize this one?

packages/babel/rollup.config.mjs
Andarist: q: have you tested the plugin works now correctly in both CJS and ESM modes?

packages/babel/package.json
davidtaylorhq: This change was landed in the other PR. Marking resolved

packages/babel/rollup.config.mjs
davidtaylorhq: It works, and I've now added a test which demonstrates that :+1:

packages/babel/src/index.js
davidtaylorhq: Standardized on skipPreflightCheck :+1:

packages/babel/src/index.js
davidtaylorhq: It's tricky to move it entirely to `unpackInputPluginOptions`, because at that point the 'overrides/customCallback' functions haven't been evaluated, and so we don't have the final babelOptions.

I have improved the error message to give a better indication of which option has failed the serializability check 👍

packages/babel/src/transformCode.js
davidtaylorhq: Yes that's right - the output plugin doesn't pass `babelHelpers` :+1:

packages/babel/src/index.js
davidtaylorhq: Yeah I think that makes sense - that'll make things consistent in parallel/non-parallel mode. Done :+1:

packages/babel/src/workerPool.js
davidtaylorhq: I investigated this some more, and I do think that switching to the `workerpool` package is a good idea. It handles a lot more edge cases than the custom implementation, and is much less code.

I've pushed a refactor to use this. But if there are any concerns about introducing a new dependency, let me know and I can revert back to the custom implementation.

packages/babel/src/index.js
davidtaylorhq: Refactored so that the `parallel` option is available for the output plugin as well :+1:

Andarist:
@davidtaylorhq please ping me for a review explicitly when you wrap up the work here and undraft it. I'll try not to miss it

davidtaylorhq:
@Andarist I think this is in a good spot now. I've worked through all of the human & machine comments above, and I've refactored it to use the `workerpool` package instead of a bespoke implementation. Let me know if you have any further questions/suggestions!

I've updated the `babel-parallel-built` branch on my fork in case anyone wants to test the latest version against their apps (cc @NullVoxPopuli @mansona)

```json
"@rollup/plugin-babel": "davidtaylorhq/rollup-plugins#babel-parallel-built&path:/packages/babel",
```

packages/babel/src/index.js
```diff
@@ -100,6 +102,68 @@ const returnObject = () => {
...
+  const nonSerializableKey = findNonSerializableOption(babelOptions);
+  if (nonSerializableKey) {
+    return Promise.reject(
+      new Error(
+        `Cannot use "parallel" mode because the "${nonSerializableKey}" option is not serializable.`
+      )
+    );
+  }
+
+  return workerPool.exec('transform', [taskOpts]).catch((err) => context.error(err.message));
```
Andarist: q: does `context.error` actually throws/fails the execution or does it only log the error? I hope it's the first one

packages/babel/src/index.js
```diff
@@ -100,6 +102,68 @@ const returnObject = () => {
...
+
+  // Default limits to 4 workers. Benefits diminish after this point, because of the setup cost.
+  const workerCount = typeof parallel === 'number' ? parallel : Math.min(cpus().length, 4);
+  return workerpool.pool(WORKER_PATH, {
+    maxWorkers: workerCount,
+    workerType: 'thread'
+  });
+}
+
+function executeWithWorkerPool(workerPool, context, taskOpts, babelOptions) {
```
Andarist: the name of this suggests it's a generic helper for executing different tasks but actually it's only capable of calling transform. I think the name of this helper should reflect that. Please just rename this to `transformWithWorkerPool` and then `taskOpts` should become `transformsOpts` too

packages/babel/src/index.js
```diff
@@ -162,22 +229,46 @@ function createBabelInputPluginFactory(customCallback = returnObject) {
...
+      async closeBundle() {
+        if (workerPool && !this.meta.watchMode) {
+          await workerPool.terminate();
+        }
+      },
+
+      async closeWatcher() {
+        if (workerPool) {
+          await workerPool.terminate();
         }
```
Andarist: nit: this could be `await workerPool?.terminate()`

packages/babel/src/index.js
```diff
@@ -257,7 +351,36 @@ function createBabelOutputPluginFactory(customCallback = returnObject) {
...
+          customOptions,
+          error: this.error.bind(this),
+          skipPreflightCheck: true
+        });
+      },
+
+      async generateBundle() {
+        if (workerPool && !this.meta.watchMode) {
+          await workerPool.terminate();
+        }
```
Andarist: nit: similarly here this could be simplified
```suggestion
        if (!this.meta.watchMode) {
          await workerPool?.terminate();
        }
```

packages/babel/src/index.js
davidtaylorhq: Correct - it fails the build. See https://rollupjs.org/plugin-development/#this-error

> Structurally equivalent to [this.warn](https://rollupjs.org/plugin-development/#this-warn), except that it will also abort the bundling process with an error.

packages/babel/src/index.js
davidtaylorhq: Done :+1:

packages/babel/src/index.js
davidtaylorhq: Done 👍

packages/babel/src/index.js
davidtaylorhq: Done :+1:

davidtaylorhq:
@Andarist just checking in on this. Please let me know if you have any more suggestions 🙏

packages/babel/src/index.js
```diff
@@ -1,11 +1,13 @@
+import { cpus } from 'os';
```
Andarist:
```suggestion
import { cpus } from 'node:os';
```

packages/babel/src/index.js
```diff
@@ -1,11 +1,13 @@
...
+import { fileURLToPath } from 'url';
```
Andarist:
```suggestion
import { fileURLToPath } from 'node:url';
```

Andarist (approved):

davidtaylorhq:
Hmm, it seems like those `node:os` and `node:fs` changes cause a build error

> [!] Error: "node:os" is imported by "src/index.js", but could not be resolved – treating it as an external dependency.

Other plugins in this repository seem to refer import `os` and `fs` without the `node:` prefix, so maybe we should stick to that. What do you think @Andarist 

I restored the non-`node:`-prefixed versions for now

davidtaylorhq:
Hi @shellscape, just checking in on this. Please let me know if you have any questions/concerns. Thanks!

mansona:
Hey @shellscape 👋 me again. I don't know if you have notifications setup for this stuff but I would really like to get this merged if there are no more issues with it. 

This change has shown to significantly improve performance of people relying on Babel and building with Vite, and Ember still heavily relies on babel and has recently switched our default build system to Vite 😂 so we're super keen to get this out to people when you get a chance 🙏

shellscape:
@CharlieHelps can any part of this PR be considered a breaking change

shellscape:
thanks!

Zaitnoori:
Boom

PR #1957, opened 2026-01-26
fix(commonjs): correctly replaces shorthand "global" property in object

TrickyPi:

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

resolves https://github.com/rollup/rollup/issues/6242#issue-3839241465


shellscape:

PR #1958, opened 2026-02-08
fix(commonjs): handle global shorthand property correctly

kaigritun:
## Problem

When `global` is used as a shorthand property in an object literal (e.g., `{global}`), the commonjs plugin replaces `global` with `commonjsHelpers.commonjsGlobal`, resulting in invalid syntax `{commonjsHelpers.commonjsGlobal}`.

This causes a parse error like:
```
Unexpected token \`.\`. Expected ... , *, (, [, :, , ?, = or an identifier
```

## Root Cause

As identified in rollup/rollup#6242 by @TrickyPi, the CommonJS plugin transforms `{global}` into `{commonjsHelpers.commonjsGlobal}`, which is invalid syntax because you can't have a dotted expression as a shorthand property.

## Solution

This fix follows the same pattern already used for the `require` shorthand case (see existing `shorthand-require` test): when detecting a shorthand property, prepend `global: ` to convert it to a full property before replacing the value.

The transformation now correctly produces:
- Before: `{global}` → `{commonjsHelpers.commonjsGlobal}` (invalid syntax)
- After: `{global}` → `{global: commonjsHelpers.commonjsGlobal}` (valid syntax)

## Changes

1. Added shorthand property handling for the `global` identifier in `transform-commonjs.js`
2. Added a test case in `test/fixtures/function/shorthand-global/`

## Related Issues

Fixes rollup/rollup#6242

### DIFF
--- a/packages/commonjs/src/transform-commonjs.js
+++ b/packages/commonjs/src/transform-commonjs.js
@@ -322,6 +322,12 @@ export default async function transformCommonjs(
             case 'global':
               uses.global = true;
               if (!ignoreGlobal) {
+                if (isShorthandProperty(parent)) {
+                  // as key and value are the same object, isReference regards
+                  // both as references, so we need to skip now
+                  skippedNodes.add(parent.value);
+                  magicString.prependRight(node.start, 'global: ');
+                }
                 replacedGlobal.push(node);
               }
               return;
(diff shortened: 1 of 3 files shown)


TrickyPi:
This PR appears to duplicate #1957

shellscape:

shellscape:
@kaigritun please take a look at this review https://github.com/rollup/plugins/pull/1958#issuecomment-3906321469 and resolve the must-fix items.

Issue #1959, opened 2026-02-11
Feature: Bump terser dependencies


tada5hi:

fskpf:
Would love to see an update here. `rollup/plugin-terser` currently comes with a high vulerability from `serialize-javascript`. This was fixed downstream here https://github.com/yahoo/serialize-javascript/releases/tag/v7.0.3

dargmuesli:
Main breaking change in `serialize-javascript` v7 is the Node v20 set as minimum. Given that the current minimum of Node v18 does not match the minimum of the terser plugin of v14, I think the update can be pulled in. cc @shellscape would you agree?

Edit: actually, `serialize-javascript` v7 makes use of native node crypto functionality which is either available starting in Node v15 or v17.4 (see https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues).

dargmuesli:
PR submitted: https://github.com/rollup/plugins/pull/1968

I also propose to drop support for EOL node versions in that PR which is a breaking change of course.

PR #1960, opened 2026-02-11
fix(terser): bump serialize-javascript, smob & terser

tada5hi:
resovles #1959


tada5hi:
@shellscape shouldn't we drop tests for Node v18, thus EOF is long in the past now?

shellscape:
@tada5hi #1968 made some updates. will need to fix up the conflicts to merge this one.

PR #1961, opened 2026-02-13
feat(sucrase): add support for jsxRuntime and other missing Sucrase options

vinny-silveira:
## Rollup Plugin Name: `sucrase`

This PR contains:

- [ ] bugfix
- [x] feature
- [x] refactor
- [x] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

_(No specific issues found, but this PR adds support for options that already exist in Sucrase but were not exposed in the plugin.)_

### Description

This PR adds support for missing Sucrase options that were not being passed through to the `transform` function:

- `jsxRuntime` - Enables the automatic JSX runtime (React 17+);
- `jsxImportSource` - Custom JSX import source (e.g., `preact`);
- `preserveDynamicImport` - Preserve dynamic import expressions;
- `injectCreateRequireForImportRequire` - Inject createRequire for import.meta.

Additionally, this PR includes:

1. **Refactor**: Simplified the `resolveId` function by removing a redundant ternary operator (the `importer` check guarantees it's always truthy at that point);
2. **Tests**: Added a `getBundle` helper function to reduce boilerplate in tests, plus new tests for plugin initialization and the exclude filter;
3. **Documentation**: Updated README.md to list all available Sucrase options.


shellscape:
Sorry about the copilot noise. I have no idea how that got turned on. But please disregard it.

vmsilveira:
No worries at all, @shellscape! Totally understand, thanks for the heads-up. Happy to make any changes if needed.

shellscape:

vmsilveira:
Thanks for the thorough review, i really appreciate the detailed feedback.

 I’ve addressed all the points you raised:

- **Rollup `context` handling**: removed the hard-coded `context: 'this'` from `getBundle()`, restoring Rollup’s default top-level `this` semantics. Snapshots were updated accordingly;
- **`jsxImportSource` coverage**: added a focused runtime assertion using `jsxRuntime: "automatic"` and a custom `jsxImportSource` to ensure the plugin effectively influences the emitted `*/jsx(-dev)-runtime` import path;
- **`preserveDynamicImport` and `injectCreateRequireForImportRequire`**: Added targeted tests for `preserveDynamicImport` and `injectCreateRequireForImportRequire`, ensuring these options behave as expected.

vmsilveira:
**Quick note on the Rollup `context: 'this'`**

Just to explain the original rationale behind hard-coding `context: 'this'` in `getBundle()` (now removed per the review):

I initially added it to silence Rollup’s build-time warnings about top-level `this` being undefined in ES modules. This was purely a bundler-level configuration choice and has no impact on the plugin’s transformation logic — Sucrase operates entirely at transform time and never interacts with the runtime `this` value.

The snapshot change (`__self: this` vs `__self: undefined`) was simply a reflection of Rollup’s context setting rather than a behavioral change in the plugin itself. That said, I completely agree with keeping tests aligned with Rollup’s default semantics, so I’ve reverted the change and updated the snapshots back to `undefined`. The warnings were just noise during local runs, with no functional impact either way.

shellscape:
That's fair, and Charlie doesn't always get reviews correct. If you feel that's worth pushing back on, you can revert your "fix" in reply to his review, and add a note in the source to the same effect as your comment above.

vmsilveira:
Thanks for the clarification — that makes sense.

I’ve reverted the removal and restored context: `this`, with an inline comment explaining the rationale. The intent is purely to keep test output clean by silencing Rollup’s top-level this warnings, especially as the test surface grows over time.

This remains a bundler-level concern only and doesn’t affect the plugin’s transform behavior in any way.

shellscape:
@vinny-silveira are you using AI to reply to this PR?

vmsilveira:
No. I only use Google Translate because my native language is Brazilian Portuguese, sorry if that sounds too formal.

shellscape:
OK just checking because no humans use an emdash (—).

PR #1968, opened 2026-03-01
chore(terser)!: upgrade serialize-javascript to v7 and node to v20

dargmuesli:
## Rollup Plugin Name: `terser

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [ ] no

Resolves https://github.com/rollup/plugins/issues/1959

### Description

Upgrade serialize-javascript to v7 and node to v20 to resolve a security vulnerability notice.
BREAKING CHANGE: drop support for EOL node versions

### DIFF
--- a/.github/workflows/node-windows.yml
+++ b/.github/workflows/node-windows.yml
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        node: ['20', '18']
+        node: ['24', '20']
 
     name: Node v${{ matrix.node }}
     steps:
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        node: ['20', '18']
+        node: ['20', '24']
 
     name: Node v${{ matrix.node }}
 


dargmuesli:
@tada5hi I've pushed a change to let the ci pipelines match this PR's scope.

.github/workflows/node-windows.yml
```diff
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        node: ['20', '18']
+        node: ['24', '20']
```
stof, line 17: shouldn't this also include `22` ?

dargmuesli: If you want an explicit for it, yes. If you just wanna check bounds, this could be sufficient. Comes down to being entirely sure or limiting resource usage and saving some of the planet as a trade off.

shellscape: yeah I'm good with this. honestly we really only need to be testing the latest, since node does really good back compat, but fine with this

shellscape (approved):
thanks!

PR #1983, opened 2026-03-12
chore(pluginutils): remove unused peer dependency

remcohaszing:
## Rollup Plugin Name: `@rollup/pluginutils`

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

refs https://github.com/mdx-js/mdx/issues/2671

### Description

`@rollup/pluginutils` declares a peer dependency on `rollup`. However, it doesn’t actually use this dependency anywhere.


shellscape:

shellscape:
@remcohaszing thanks for the PR. while it might be attractive to use the pluginutils package outside of rollup, the peerDep here matches dep enforcement with other plugins/packages in the repo, and enforces a contract. We're going to decline this PR for the moment.

PR #1988, opened 2026-03-23
perf(yaml)!: use hook filters

sapphi-red:
## Rollup Plugin Name: `yaml`

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no (covered by existing tests)

Breaking Changes?

- [ ] no

### Description

Similar to https://github.com/rollup/plugins/pull/1954, but for yaml plugin. This PR adds plugin hook filters for improved performance for Rolldown.
Object style hooks are added by https://github.com/rollup/rollup/pull/4600 which is released in 2.78.0, so I bumped the minimum required Rollup version.

### DIFF
--- a/packages/yaml/README.md
+++ b/packages/yaml/README.md
@@ -13,7 +13,7 @@
 
 ## Requirements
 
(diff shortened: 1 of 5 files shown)


shellscape:

PR #1999, opened 2026-05-14
ci: declare minimum permissions on PR title workflow

arpitjain099:
Declares `permissions: contents: read, pull-requests: read` at the workflow level. The job's single step runs a PR-title validator (`amannn/action-semantic-pull-request` or equivalent) which reads the PR title via the pulls API and reports the result as the workflow's own check status. That needs read on `pull-requests` and nothing else.

The reason to be explicit even when the inherited default may already be reasonable is CVE-2025-30066, the March 2025 `tj-actions/changed-files` compromise where a tampered third-party action exfiltrated `GITHUB_TOKEN` from workflow logs. The leaked token carried whatever scope was issued at the workflow level, so per-workflow caps bound the runtime blast radius regardless of repo or org default. The block also gives drift protection if that default ever widens and is what OpenSSF Scorecard's Token-Permissions check looks for.

YAML validated locally with `yaml.safe_load`.

### DIFF
--- a/.github/workflows/pr-title.yml
+++ b/.github/workflows/pr-title.yml
@@ -10,6 +10,10 @@ on:
       - edited
       - synchronize
 
+permissions:
+  contents: read
+  pull-requests: read
+
 jobs:
   prTitle:
     name: Check


shellscape:
OK you just opened an identical one on rollup/stream, and for the second time ignored the PR template, which means you're automating this stuff and that isn't cool. Don't do that again on this organization.

arpitjain099:
Hey @shellscape, thanks for the direct feedback and you're right to call it out. I do lean on tooling to draft these (English is not my first language and the workflow-permissions audit script makes it easier to keep the diff scope tight), but I manually review each PR before opening - that's on me for not catching the template ask and for filing the rollup/stream duplicate. Both rollup PRs are now closed and I'm taking the rollup org off my contribution list. Won't open more here. Apologies for the noise.

PR #2002, opened 2026-05-29
feat(pluginutils): named exports for reserved-word keys + fix duplicate default export in dataToEsm

### LINKED ISSUE #2001: dataToEsm emits a duplicate default export for a `default` key with includeArbitraryNames
- Rollup Plugin Name: @rollup/pluginutils
- Rollup Plugin Version: 5.3.0 (and current `master`)
- Rollup Version: N/A — `dataToEsm` is a standalone utility, no bundling involved
- Operating System (or Browser): macOS
- Node Version: 20+
- Link to reproduction _(⚠️ read below)_: self-contained snippet below (pure function, no bundler needed)

### Expected Behavior

When `dataToEsm` is called with `includeArbitraryNames: true` on data that has a
key named `default`, it should produce valid JavaScript. Since `dataToEsm` always
emits a trailing `export default {...}`, the `default` key can only be exposed
through that default export object — it must not also be emitted as a named
export.

### Actual Behavior

`dataToEsm` emits `export { _arbitrary0 as "default" }` **in addition to** the
trailing `export default {...}`. Both forms declare the module's default export,
so the generated code is invalid:

```
SyntaxError: Duplicate export of 'default'
```

Generated output for the reproduction below:

```js
const _arbitrary0 = "a";
export const normal = "b";
export {
	_arbitrary0 as "default"
};
export default {
	"default": "a",
	normal: normal
};
```

### Additional Information

Reproduction (pure function — runs in Node, no bundler required):

```js
import { dataToEsm } from '@rollup/pluginutils';

const code = dataToEsm(
  { default: 'a', normal: 'b' },
  { namedExports: true, preferConst: true, includeArbitraryNames: true }
);

// Throws: SyntaxError: Duplicate export of 'default'
await import('data:text/javascript,' + encodeURIComponent(code));
```

A `default` key falls into the `includeArbitraryNames` branch and is emitted as
`export { _x as "default" }`. But both the string form `export { x as "default" }`
and the identifier form `export { x as default }` create the module's default
export, which conflicts with the `export default {...}` that `dataToEsm` always
appends.

This surfaced while working on Vite's CSS-modules named exports
(https://github.com/vitejs/vite/pull/22393): a `.default {}` class name compiled
for a modern target produces broken output.

A fix (skip a `default` key from named exports, keeping it object-only) with a
test is ready and will be opened as a PR referencing this issue.

semimikoh:
…in dataToEsm

## Rollup Plugin Name: `@rollup/pluginutils`

This PR contains:

- [x] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

resolves #2001

### Description

Two related changes to `dataToEsm`:

**1. Feature — emit ES2015-safe named exports for reserved-word / identifier-name keys**

A key that is a valid `IdentifierName` but not a legal binding identifier (a
reserved word, global, or literal — `switch`, `await`, `null`, `true`, …) is now
re-exported with the unquoted `export { _x as switch }` form, which is valid
since ES2015. Previously such keys were emitted as named exports only under
`includeArbitraryNames`, and then via the quoted ES2022 form
(`export { _x as "switch" }`).

This is what Vite needs for CSS-modules named exports: a class such as `.switch`
or `.default` should produce a named export even when targeting browsers that
predate arbitrary module namespace names (https://github.com/vitejs/vite/pull/22393).
The unquoted form covers that without `includeArbitraryNames`; the quoted form
stays opt-in and is now used only for keys that are *not* valid `IdentifierName`s
(e.g. `foo-bar`).

> **Behavior change:** for callers not passing `includeArbitraryNames`,
> reserved-word / global / literal keys (e.g. `null`, `true`, `else`) now gain an
> ES2015-valid named export where previously they had none. This is strictly
> additive — it only emits syntax already required by the rest of the output, and
> existing imports are unaffected. Two existing test expectations
> (`serializes null`, `supports a compact argument`) are updated accordingly.

**2. Bugfix — avoid a duplicate default export for a `default` key**

`dataToEsm` always appends a trailing `export default {...}`. A key named
`default` was also emitted as `export { _x as "default" }` (under
`includeArbitraryNames`), producing two default exports and a
`SyntaxError: Duplicate export of 'default'`. A `default` key is now exposed only
through the default export object.

### Steps to execute / reproduce

```js
import { dataToEsm } from '@rollup/pluginutils';

// reserved-word named export (ES2015-safe), no includeArbitraryNames needed
dataToEsm({ switch: 'a', normal: 'b' }, { namedExports: true, preferConst: true });
// -> const _arbitrary0 = "a"; export const normal = "b";
//    export { _arbitrary0 as switch }; export default { "switch": "a", normal };

// the duplicate-default bug, before this PR:
const code = dataToEsm({ default: 'x' }, { namedExports: true, includeArbitraryNames: true });
await import('data:text/javascript,' + encodeURIComponent(code));
// previously threw: SyntaxError: Duplicate export of 'default'
```

### DIFF
--- a/packages/pluginutils/src/dataToEsm.ts
+++ b/packages/pluginutils/src/dataToEsm.ts
@@ -115,7 +115,10 @@ const dataToEsm: DataToEsm = function dataToEsm(data, options = {}) {
       defaultExportRows.push(
         `${stringify(key)}:${_}${serialize(value, options.compact ? null : t, '')}`
       );
-      if (options.includeArbitraryNames && isWellFormedString(key)) {
+      // A `default` key is skipped here and exposed only through the default
+      // export object: a `... as default` re-export would clash with the trailing
+      // `export default` and produce a duplicate default export (a SyntaxError).
+      if (key !== 'default' && options.includeArbitraryNames && isWellFormedString(key)) {
         const variableName = `${arbitraryNamePrefix}${arbitraryNameExportRows.length}`;
         namedExportCode += `${declarationType} ${variableName}${_}=${_}${serialize(
           value,
--- a/packages/pluginutils/src/dataToEsm.ts
+++ b/packages/pluginutils/src/dataToEsm.ts
@@ -70,6 +70,11 @@ function isWellFormedString(input: string): boolean {
   return !/\p{Surrogate}/u.test(input);
 }
 
+// Matches the ECMAScript `IdentifierName` grammar, which (unlike a binding
+// identifier) also accepts reserved words such as `switch`/`await`. Such names
+// can be re-exported with the `export { _x as switch }` form, valid since ES2015.
+const identifierNameRE = /^[$_\p{ID_Start}][$\u200c\u200d\p{ID_Continue}]*$/u;
+
 const dataToEsm: DataToEsm = function dataToEsm(data, options = {}) {
   const t = options.compact ? '' : 'indent' in options ? options.indent : '\t';
   const _ = options.compact ? '' : ' ';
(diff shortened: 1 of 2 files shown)


shellscape:

shellscape:
thanks!

PR #2013, opened 2026-06-29
fix(esm-shim): match typings to actual exports

mahirhir:
The hand-written `types/index.d.ts` for `@rollup/plugin-esm-shim` does not match what the package actually exports.

`src/index.ts` exposes the plugin as both a named and a default export, and re-exports the `Output` type:

```js
export * from './types';   // Output
export default esmShim;
export { esmShim };
```

The typings only declare the default, under the stale name `commonjsShim`, and omit the named `esmShim` export. As a result `import { esmShim } from '@rollup/plugin-esm-shim'` does not type-check even though it works at runtime.

This is the same typings/exports drift that #1985 fixed for `@rollup/plugin-swc`. Every other plugin with a named export already declares it in its typings (swc, babel, wasm, node-resolve, html, dynamic-import-vars); esm-shim is the one that was left out.

Changes:
- declare the named `esmShim` export and give the default its real name, matching the shape #1985 landed (`export function swc(...): Plugin; export default swc;`)
- export the `Output` type, which `src/index.ts` already re-exports from `./types`
- assert the named export in the package's existing `test:ts` type test

`pnpm --filter @rollup/plugin-esm-shim test:ts` reports `has no exported member 'esmShim'` on the current typings and passes with this change.


mahirhir:
Closing this myself: the repository has had no push in 99 days, so there is no one to review it right now. No hard feelings and no action needed. If the project picks up again and this is still wanted, I am happy to reopen or re-send it.

PR #2021, opened 2026-08-10
feat(babel)!: support Babel 8

### LINKED ISSUE #2014: Babel 8 support
- Rollup Plugin Name: babel
- Rollup Plugin Version:

### Feature Use Case

Babel 8 has been released: https://babeljs.io/blog/2026/06/16/8.0.0/

It would be great to support it in `@rollup/plugin-babel`

itsjamie:

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] no

resolves #2014

### Description

Updates the plugin to Babel 8, following the [v8 migration guide](https://babeljs.io/docs/v8-migration) and the [API migration guide](https://babeljs.io/docs/v8-migration-api). The peer dependency is now `@babel/core ^8.0.0`; Babel 7 users should stay on plugin v7.

I bumped the dependencies first and ran the suite against Babel 8 before touching anything, to see what actually breaks: 35 of 57 tests failed. The causes were preset-env's new default targets (browserslist `defaults` instead of `>= 0%`, ~22 tests), the removed `loose`, `useESModules` and decorators `legacy: true` options, the removed `@babel/plugin-syntax-dynamic-import` package, and codegen drift in sourcemaps and helper interop. The fixtures now pin explicit `targets` so the tests keep asserting ES5 output.

Source changes are small: drop the sync `loadPartialConfig` fallback (removed in v8), execute the old TODO in `preflightCheck.js` reverting the helper-path regex (babel/babel#14185 is fixed), and remove the `runtimeHelpers`/`externalHelpers` option shims that have warned since v5. The handwritten types follow Babel 8's renames (`TransformOptions` → `InputOptions`, `BabelFileResult` → `FileResult`) and the `@types/babel__core` peer is gone since Babel 8 ships its own types.

Since all `@babel/*` packages are ESM-only now, the plugin is too: the CJS build is gone and `src/` is published directly, with `engines` matching Babel 8 (`^22.18.0 || >=24.11.0`). `require('@rollup/plugin-babel')` still works on supported Node versions via `require(esm)`. The [`'module.exports'` named export](https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require) preserves the callable shape with named properties that the old CJS build had, and the two CJS tests still cover it.

The CI matrix is untouched; the package's `test` script now skips itself with exit 0 on Node versions Babel 8 can't run on (the Node 20 legs), so real coverage comes from the Node 24 leg.

Two behaviour notes from the test diffs rather than this PR's code: transform-runtime CJS output now emits `require("@babel/runtime/helpers/createClass").default` (helpers are ESM-first, resolved via `@babel/runtime`'s `package.json#exports`), and Babel 8 no longer records sourcemap name mappings for identifiers whose text is unchanged (positions still map correctly).


vanbukin:
@Andarist @shellscape any plans?

